PR #3, opened 2015-12-03
Linux port hdd cp1

dgrove-oss:
Some progress towards building libdispatch on Linux.  

With this set of changes, libdispatch.so builds successfully and a test program can be linked against it.  However, execution of the test program fails during initialization of work queue data structures (called from dispatch_init)


os/base.h
```diff
@@ -0,0 +1,10 @@
+// Stub out Apple internal header file to redirect to linux_base.h.
```
das: this is problematic, OSX and iOS already have a public header with this name from a different project and this will conflict 

Either we have to conditionalize the install of this header based on platform (and fully replicate its contents from OSX), or we have to conditionalize its inclusion.

an `AC_CHECK_HEADERS([os/base.h])` test in configure.ac would be the standard way to do the latter, but since most of the `#include <os/base.h>` are in API headers, that is not really an option either (since we don't want to install config.h on OSX)

the best option for now may be to move the various `#include <os/base.h>` into `#ifdef __APPLE__` conditionals and in the `#else` case do the `#include <os/linux_base.h>` (or possibly include an intermediate header that does non-APPLE platform disambiguation, though based on what ?)

os/linux_base.h
```diff
@@ -0,0 +1,147 @@
...
+#define NOTE_NSECONDS 0
+#define NOTE_LEEWAY 0
+#define NOTE_CRITICAL 0
+#define NOTE_BACKGROUND 0
+
+/*
+ * Stub out misc linking and compilation attributes
+ */
+
+#ifdef OS_EXPORT
```
das: it would be better to put the `OS_*` vmacros from the real os/base.h into the replicated version in the libdispatch repo (if we go that route)

private/voucher_private.h
```diff
@@ -400,7 +400,9 @@ dispatch_queue_create_with_accounting_override_voucher(const char *label,
  * voucher ports directly.
  */
 
+#ifdef HAVE_MACH
```
das: we cannot have these types of checks inAPI headers, they will not pass on OSX (nobody defines HAVE_MACH). I would use either  `__APPLE__` or the `TARGET_OS_MAC` macro from TargetConditionals.h (there is also a `TARGET_OS_UNIX` there that might make sense for the 'linux-specific' checks)

os/base.h
das: thinking about it more, something like this may make the most sense (maybe in a new os/target.h header ?)

    #ifdef __APPLE__
    #else
    #define TARGET_OS_UNIX 1
    #endif

and then conditionalize with `#if TARGET_OS_MAC` resp. `#if TARGET_OS_UNIX` ?

src/internal.h
```diff
@@ -82,6 +82,7 @@
 
 
 #include <os/object.h>
+#include "stubs_internal.h"
```
das: this seems like the wrong spot for this inclusion, plus we have the shims/ directory for this type of thing already, please use shims.h and add new shims/stubs in there

src/io.c
```diff
@@ -387,9 +387,14 @@ dispatch_io_create_with_path(dispatch_io_type_t type, const char *path,
...
 		_dispatch_io_syscall_switch_noerr(err,
-			(path_data->oflag & O_NOFOLLOW) == O_NOFOLLOW ||
-					(path_data->oflag & O_SYMLINK) == O_SYMLINK ?
+			(path_data->oflag & O_NOFOLLOW) == O_NOFOLLOW
+#ifdef __LINUX_PORT_HDD__						  
+					?
 					lstat(path_data->path, &st) : stat(path_data->path, &st),
+#else
+					|| (path_data->oflag & O_SYMLINK) == O_SYMLINK ?
+					lstat(path_data->path, &st) : stat(path_data->path, &st),
```
das: this line doesn't seem like it needs to be duplicated ?

src/io.c
```diff
@@ -1777,17 +1782,37 @@ _dispatch_stream_cleanup_operations(dispatch_stream_t stream,
 	dispatch_operation_t op, tmp;
 	typeof(*stream->operations) *operations;
 	operations = &stream->operations[DISPATCH_IO_RANDOM];
+#ifdef __LINUX_PORT_HDD__
+	LINUX_PORT_ERROR();
+        // NOT CORRECT (doesn't support removal during iteration)
+	TAILQ_FOREACH(op, operations, operation_list) {
```
das: maybe we should add a copy of the bsd [sys/queue.h](https://github.com/freebsd/freebsd/blob/master/sys/sys/queue.h])with the safe macros to shims/?  (BSD license should be compatible)

das: or at least a header that adds the missing macros, I really don't like the duplication of all of the `FOREACH_SAFE` blocks, that will be a maintenance and merging headache

src/io.c
```diff
@@ -2064,6 +2099,9 @@ _dispatch_disk_perform(void *ctxt)
 static void
 _dispatch_operation_advise(dispatch_operation_t op, size_t chunk_size)
 {
+#ifdef __LINUX_PORT_HDD__
```
das: this should be a configure check for the `F_RDADVISE` fcntl (or short-term just an `#ifdef  F_RDADVISE`)

src/stubs.c
```diff
@@ -0,0 +1,102 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @APPLE_APACHE_LICENSE_HEADER_END@
+ */
+
+
+/*
```
das: move this to shims/linux_stubs.c  ?

src/stubs_internal.h
```diff
@@ -0,0 +1,37 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @APPLE_APACHE_LICENSE_HEADER_END@
+ */
+
+// forward declarations for functions we are stubbing out
+// in the intial linux port.
+
+#ifndef __DISPATCH__STUBS__INTERNAL
```
das: move this to shims/linux_stubs.h ?

src/transform.c
```diff
@@ -20,7 +20,9 @@
 
 #include "internal.h"
 
+#ifdef __APPLE__
 #include <libkern/OSByteOrder.h>
```
das: this header is [opensource](http://www.opensource.apple.com/source/xnu/xnu-792.13.8/libkern/libkern/OSByteOrder.h) and all of the implementations of OSSwap* are contained in the header, we should figure out if we can import it into the repo (or re-implement it, it's pretty trivial)

src/io.c
dgrove-oss: Agreed.  This was intended as a really short term stop-gap to get things to build.  Not something that should be merged back.  We'll at least pull in the missing safe macros to shims.

dgrove-oss:
I've reworked the changes and submitted a new pull request: https://github.com/apple/swift-corelibs-libdispatch/pull/5.   Therefore closing this one.

PR #4, opened 2015-12-04
Fix Typos

dbworku:
*Fix misspellings*
There were minor typos on a couple comments.

*Remove ns abbreviation for nanoseconds*
The time parameters are clearly described as being in nanoseconds.

*Unify over-release spelling*
We were using "overrelease" and "over-release". I changed everything to the later.


das:
looking good, thanks!

PR #6, opened 2015-12-05
Add & use DISPATCH_NOESCAPE attribute macro

jtbandes:
This will allow the block parameters to dispatch_sync, dispatch_barrier_sync, and dispatch_apply to be exposed to Swift code as `@noescape`.

### DIFF
--- a/dispatch/base.h
+++ b/dispatch/base.h
@@ -124,6 +124,12 @@
 #endif
 #endif
 
+#if defined(__has_attribute) && __has_attribute(noescape)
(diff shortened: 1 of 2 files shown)


parkera:
Hi @jtbandes - this pull request contains an API change, so let’s follow the [Swift evolution process](https://swift.org/contributing/#participating-in-the-swift-evolution-process), starting with discussing the idea on the `swift-evolution` and `swift-corelibs-dev` lists. We are considering a similar proposal for `swift-corelibs-foundation` as well, so perhaps we can come up with the same answer for both projects.

PR #7, opened 2015-12-06
Add the download command of the library that libdispatch depends

zchee:
More kind installation command for developers.

but,

- libplatform
- xnu build 3247.1.106
- objc4 build 680

these libraries are not open source.
Where can I download? Is still closed source?


parkera:
Let's discuss the question on the mailing list, but for now I will close this because I don't think we should put instructions in that we know aren't going to work.

PR #8, opened 2015-12-09
Linux port hdd timers2

frankeh:
Additional changes to handle Timers using kevent properly.


src/source.c
```diff
@@ -1858,13 +1858,27 @@ _dispatch_timers_get_delay(uint64_t nows[], struct dispatch_timer_s timer[],
 	return ridx;
 }
 
+
+#if HAVE_KEVENT64
```
MadCoder: this block is wrong, and in any case should live in shims/time.h

I think what you mean here is "do we have support for NOTE_NSECONDS"
and not "do we have KEVENT64", you would break FreeBSD doing this.

Please do not mingle OS support with code, use shims/ and clean wrappers.

It'd be best to push NOTE_NSECOND support in libkqueue in the first place, which is probably easiest, so that you can use the thing directly anyway.

src/source.c
```diff
@@ -1881,13 +1895,15 @@ _dispatch_timers_program2(uint64_t nows[], _dispatch_kevent_qos_s *ke,
 		_dispatch_trace_next_timer_set(
 				TAILQ_FIRST(&_dispatch_kevent_timer[tidx].dk_sources), qos);
 		_dispatch_trace_next_timer_program(delay, qos);
-		delay += _dispatch_source_timer_now(nows, DISPATCH_TIMER_KIND_WALL);
+	        nowtime =_dispatch_source_timer_now(nows, DISPATCH_TIMER_KIND_WALL);
+	        adjust_delay(delay,nowtime);
```
MadCoder: should be removed, and libkqueue updated to do conversions from ns to ms internally instead.

src/source.c
```diff
@@ -1881,13 +1895,15 @@ _dispatch_timers_program2(uint64_t nows[], _dispatch_kevent_qos_s *ke,
...
+
 		if (slowpath(_dispatch_timers_force_max_leeway)) {
 			ke->data = (int64_t)(delay + leeway);
-			ke->ext[1] = 0;
+	                kevent_set_ext1(ke,0);
```
MadCoder: ditto

src/source.c
frankeh: Thanks. The intention initially was to get something running.
I agree that better place to get this moved is libkqueue which internally uses nsec time management anyway. We can squash this patch set, clean it up w/o kevent64(@libkqueue)
for now according to your suggestions (shims/time.h) and then do a full update on the
libkqueue to bring kevent64 semantics as known in BSD to that implementation. 
Frankly I prefer that, though not sure yet how to do with the EVFILTER_TIMER.leeway semantics. Where will we host the libkqueue, seems like Haileys has not been active for a while.

MadCoder: The problem is that your patch most likely breaks Mac OS too, so we can't take that even for the sake of the linux port making progress.

the leeway argument is easy: it's always correct to ignore it. it's about "can I delay firing that timer a bit so that I can coalesce several timers firing approximatively at the same time and save power by waking up the CPU once" If you don't do it, the code is still correct.

frankeh:
Took your generic advice. Cleaned up the sloppy code. However moving things to shims/*.h
seems to create typedef dependencies. Created a __linux__ conditionals outside
the _dispatch_timers_program2 .. this is reasonable minimal.

src/source.c
```diff
@@ -1858,6 +1858,43 @@ _dispatch_timers_get_delay(uint64_t nows[], struct dispatch_timer_s timer[],
...
+// to either extend libkqueue with a proper kevent64 API or removing kevent all together 
+// and move to a lower API (e.g. epoll or kernel_module. Also leeway is ignored.
+
+static void 
+kevent_set_delay(_dispatch_kevent_qos_s *ke, uint64_t delay,uint64_t leeway, uint64_t nows[]) 
+{
+	_dispatch_source_timer_now(nows, DISPATCH_TIMER_KIND_WALL); // called to return nows[]
+	
+	delay /= 1000000L;
+        if ((int64_t)(delay) <= 0)
```
MadCoder: tab/space issue.
we use 4-width tabs.

Our coding rules are (in vim speak):

```vimrc
set cinoptions=
set cinoptions+=Ls
set cinoptions+=:0,=1s
set cinoptions+=t0
set cinoptions+=+2s,(2s,u2s,W2s
set cinoptions+=j1,J1,m1
au FileType d,c,cpp,objc setlocal tw=80 ts=4 sw=4 noet
```

please fix the mixed indent here

src/source.c
```diff
@@ -1858,6 +1858,43 @@ _dispatch_timers_get_delay(uint64_t nows[], struct dispatch_timer_s timer[],
...
+
+#ifdef __linux__ 
+// in linux we map the _dispatch_kevent_qos_s  to struct kevent instead of struct kevent64.
+// we loose the kevent.ext[] members and the time out is based on relavite msec based time
+// vs. absolute nsec based time. For now we make the adjustments right here until the solution
+// to either extend libkqueue with a proper kevent64 API or removing kevent all together 
+// and move to a lower API (e.g. epoll or kernel_module. Also leeway is ignored.
+
+static void 
+kevent_set_delay(_dispatch_kevent_qos_s *ke, uint64_t delay,uint64_t leeway, uint64_t nows[])
```
MadCoder: This is about timer so should have timer in the name.
also to match the naming style of the file this function should be

`_dispatch_kevent_timer_set_delay`.
please have a space after the commas, and wrap the prototype if it doesn't fit on 80 colums

src/source.c
```diff
@@ -1881,17 +1918,12 @@ _dispatch_timers_program2(uint64_t nows[], _dispatch_kevent_qos_s *ke,
...
-			ke->data = (int64_t)(delay + leeway);
-			ke->ext[1] = 0;
-		} else {
-			ke->data = (int64_t)delay;
-			ke->ext[1] = leeway;
-		}
-		ke->flags |= EV_ADD|EV_ENABLE;
-		ke->flags &= ~EV_DELETE;
+
+                kevent_set_delay(ke,delay,leeway,nows);
```
MadCoder: tab/space

src/source.c
```diff
@@ -1881,17 +1918,12 @@ _dispatch_timers_program2(uint64_t nows[], _dispatch_kevent_qos_s *ke,
...
 	}
+ 
+	ke->flags |= EV_ADD|EV_ENABLE;
```
MadCoder: bogus, should stay in the previous scope. I would even argue it should be done by the new static inline you just added

src/source.c
```diff
@@ -1858,6 +1858,43 @@ _dispatch_timers_get_delay(uint64_t nows[], struct dispatch_timer_s timer[],
...
+		delay = 1; /* if he value is negative or 0 then the dispatch will stop */
+	ke->data = (int64_t)delay;
+}
+
+#else
+
+static void 
+kevent_set_delay(_dispatch_kevent_qos_s *ke, uint64_t delay,uint64_t leeway, uint64_t nows[]) 
+{
+
```
MadCoder: useless extra line

MadCoder:
I like this approach much better, the remarks I made are now purely cosmetical, please do it, I tend to leave linux_stubs* alone until it converges, but please respect coding rules outside of this

we'll update a document to explain them more clearly for future contributions shortly.

thanks!

PR #9, opened 2015-12-10
enable subdir-objects for automake

dgrove-oss:
Enable subdir-objects to silence warning caused by the addition
of linux_stubs.c to src/shims.  This seems like the most contained
fix to allow a .c file in the shims subdir without warnings.


das:
makes sense, thanks!

PR #12, opened 2015-12-14
Add AC_SEARCH_LIBS stanzas for kqueue and pthread_workqueue

dgrove-oss:
If libdispatch is using functions found in libkqueue or
libpthread_workqueue, then we want libdispatch.so to
depend on those libraries. Putting in AC_SEARCH_LIBS
stanzas results in the proper linker flags being added when
libdispatch.so is built. Therefore the application compilation
command can simply say -ldispatch and not need to explicitly
mention -lkqueue and -lpthread_workqueue.


dgrove-oss:
Redone change as suggested and submitted a new pull request.  Closing this one.

PR #16, opened 2015-12-18
sem_timedwait on Linux needs epoch-based timeout

dgrove-oss:
sem_timedwait expects a timeout argument based on epoch time;
the code was converting from dispatch_time to a relative timeout,
but failing to further convert that relative time to an
epoch-based absolute time before calling sem_timedwait.

This fix localizes the change, but is not very efficient.
Arguably it would be better to compute a conversion from
dispatch_time to epoch time during initialization and simply
apply that instead of recomputing it every time we need it.

### DIFF
--- a/src/semaphore.c
+++ b/src/semaphore.c
@@ -278,6 +278,22 @@ dispatch_semaphore_signal(dispatch_semaphore_t dsema)
 	return _dispatch_semaphore_signal_slow(dsema);
 }
 
+#ifdef USE_POSIX_SEM
+// Convert a dispatch_time_t to POSIX epoch time
+static void
+_dispatch_time_to_epoch_time(dispatch_time_t dt, struct timespec *res_ts) {
+	struct timeval now_tv, dt_tv, res_tv;
+
+	uint64_t nsec = _dispatch_timeout(dt);
+	gettimeofday(&now_tv, NULL);
+	dt_tv.tv_sec = (typeof(dt_tv.tv_sec))(nsec / NSEC_PER_SEC);
+	dt_tv.tv_usec = (typeof(dt_tv.tv_usec))((nsec % NSEC_PER_SEC) / 1000) ;
+	timeradd(&now_tv, &dt_tv, &res_tv);
+	res_ts->tv_sec = res_tv.tv_sec;
+	res_ts->tv_nsec = res_tv.tv_usec * 1000;
+}
+#endif
+
 DISPATCH_NOINLINE
 static long
 _dispatch_semaphore_wait_slow(dispatch_semaphore_t dsema,
@@ -342,9 +358,7 @@ again:
 		}
 #elif USE_POSIX_SEM
 		do {
-			uint64_t nsec = _dispatch_timeout(timeout);
-			_timeout.tv_sec = (typeof(_timeout.tv_sec))(nsec / NSEC_PER_SEC);
-			_timeout.tv_nsec = (typeof(_timeout.tv_nsec))(nsec % NSEC_PER_SEC);
+			_dispatch_time_to_epoch_time(timeout, &_timeout);
 			ret = slowpath(sem_timedwait(&dsema->dsema_sem, &_timeout));
 		} while (ret == -1 && errno == EINTR);
 
@@ -582,9 +596,7 @@ again:
 		}
 #elif USE_POSIX_SEM
 		do {
-			uint64_t nsec = _dispatch_timeout(timeout);
-			_timeout.tv_sec = (typeof(_timeout.tv_sec))(nsec / NSEC_PER_SEC);
-			_timeout.tv_nsec = (typeof(_timeout.tv_nsec))(nsec % NSEC_PER_SEC);
+			_dispatch_time_to_epoch_time(timeout, &_timeout);
 			ret = slowpath(sem_timedwait(&dsema->dsema_sem, &_timeout));
 		} while (ret == -1 && errno == EINTR);
 
--- a/src/semaphore.c
+++ b/src/semaphore.c
@@ -278,31 +278,6 @@ dispatch_semaphore_signal(dispatch_semaphore_t dsema)
 	return _dispatch_semaphore_signal_slow(dsema);
 }
 
-#ifdef USE_POSIX_SEM
-static dispatch_once_t sem_epoch_pred;
-static dispatch_time_t sem_epoch_diff;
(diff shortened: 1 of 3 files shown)


dgrove-oss:
Sure. Will do that first thing in the morning.

MadCoder:
Also, when we integrate opensource contribution into our internal process for Darwin, it's much harder when there are several patches (we have to rebase the patches and the amount of work is linear in the amount of patches), so we'll probably have the policy to ask for squashed rebased commits once a review is done.

MadCoder:
The patch looks good, please resubmit as a rebased + squashed commit

dgrove-oss:
will do.  thanks.

src/semaphore.c
```diff
@@ -278,6 +278,31 @@ dispatch_semaphore_signal(dispatch_semaphore_t dsema)
 	return _dispatch_semaphore_signal_slow(dsema);
 }
 
+#ifdef USE_POSIX_SEM
+static dispatch_once_t sem_epoch_pred;
+static dispatch_time_t sem_epoch_diff;
```
MadCoder: that should be `uint64_t`, `dispatch_time_t` is something else and cannot be used in arithmetics

src/semaphore.c
```diff
@@ -278,6 +278,31 @@ dispatch_semaphore_signal(dispatch_semaphore_t dsema)
...
+
+static void
+sem_epoch_base_init(void *ctxt DISPATCH_UNUSED)
+{
+	struct timeval epoch_now_tv;
+	dispatch_time_t dispatch_now = _dispatch_absolute_time();
```
MadCoder: ditto

src/semaphore.c
```diff
@@ -278,6 +278,31 @@ dispatch_semaphore_signal(dispatch_semaphore_t dsema)
...
+	uint64_t epoch_now = (epoch_now_tv.tv_sec * NSEC_PER_SEC) + (epoch_now_tv.tv_usec * 1000);
+	sem_epoch_diff = (dispatch_time_t)epoch_now - dispatch_now;
+}
+
+// Convert a dispatch_time_t to POSIX epoch time
+// using a computed-once delta between their time bases.
+static void
+_dispatch_time_to_epoch_time(dispatch_time_t dt, struct timespec *res_ts) {
+	dispatch_once_f(&sem_epoch_pred, NULL, sem_epoch_base_init);
+	dispatch_time_t et = dt + sem_epoch_diff;
```
MadCoder: ditto, you need to do _dispatch_timeout(dt) + sem_epoch_diff I think, exactly for that reason

MadCoder:
Actually scratch my remark, I noted a bug with the `dispatch_time_t` use. This is a dirty type that encodes different times, you have to convert it to an absolute `ns` uint64_t timestamp. it is not ok to do arithmetics on `dispatch_time_t`, it doesn't work

dgrove-oss:
If we can't do math on a dispatch_time_t (have to call _dispatch_timeout(...)) which gets us to back to a relative time from 'now', then computing the difference between the epoch time base and dispatch_time_t time base doesn't make things more efficient.  Something like the original commit is what has to happen.  I'll go back to a variant of that on this branch for you to look at, then if that looks ok do the rebase / squash.

MadCoder:
Good point, I think that the fix you're after is to add a new `_dispatch_time_to_epoch_time()` in time.h that does this:

It doesn't even need to be under #if's at all, it's useful it its own right.
IOW move all your code from semaphore.c to time.c and use it in semaphore.

```C
uint64_t
_dispatch_time_to_epoch_time(dispatch_time_t when)
{
	if (when == DISPATCH_TIME_FOREVER) {
		return DISPATCH_TIME_FOREVER;
	}
	if (when == 0) {
		return 0;
	}
	if ((int64_t)when < 0) {
		// time in nanoseconds since the POSIX epoch already
		return -(int64_t)when;
	}
	dispatch_once_f(&sem_epoch_pred, NULL, sem_epoch_base_init);
	return dt + sem_epoch_diff;	
}
```

MadCoder:
Actually the dispatch_once may be broken in the first place. I think you have to recompute anyway because the base between the absolute time (monotonic on linux I think) and gettimeofday fluctuates.

so you probably want a:

`return _dispatch_get_nanoseconds() + _dispatch_timeout(when);`

but at least when the `dispatch_time_t` was a `dispatch_walltime()` your code is efficient and correct.

MadCoder:
Which brings the implementation to something akin to this, no dispatch_once:

```C
uint64_t
_dispatch_time_to_epoch_time(dispatch_time_t when)
{
    if (when == DISPATCH_TIME_FOREVER) {
        return DISPATCH_TIME_FOREVER;
    }
    if (when == 0) {
        return 0;
    }
    if ((int64_t)when < 0) {
        // time in nanoseconds since the POSIX epoch already
        return -(int64_t)when;
    }
    return _dispatch_get_nanoseconds() + _dispatch_timeout(when);
}
```

dgrove-oss:
thanks.  I redid as suggested.  Verified the modified fix still fixes the test cases we were looking at (dispatch_group).   Let me know if this looks good and I will rebase/squash and do a new pull request for the final fix.

src/time.c
```diff
@@ -145,3 +145,19 @@ _dispatch_timeout(dispatch_time_t when)
 	now = _dispatch_absolute_time();
 	return now >= when ? 0 : _dispatch_time_mach2nano(when - now);
 }
+
+uint64_t
+_dispatch_time_to_epoch_time(dispatch_time_t when)
```
MadCoder: I fixed my initial comment, that should be _dispatch_time_to_nanoseconds()

src/time.c
```diff
@@ -145,3 +145,19 @@ _dispatch_timeout(dispatch_time_t when)
...
+{
+	if (when == DISPATCH_TIME_FOREVER) {
+		return DISPATCH_TIME_FOREVER;
+	}
+	if (when == 0) {
```
MadCoder: this is wrong too, I removed it afterwards: you want to return _dispatch_get_nanoseconsd() for that one (or fallback through the last return that will do the right thing.

MadCoder:
I made two comments, rebase + squash once integrated at once, no point in an increment for that last nitpick I think.

thanks, and sorry for the sloppy initial review.

dgrove-oss:
rebased final code into #20

PR #17, opened 2015-12-18
Fixqueue

frankeh:


frankeh:
So in the above, dgq_ctxt was not intialized as part of the global initialization
which predicated on DISPATCH_ENABLE_THREAD_POOL 
where the code faulting was predicated on DISPATCH_USE_PTHREAD_POOL
both are not mutually inclusive.
Any thoughts how to proceed, the fix above works but might not be semantically what you 
want. Predicate the code to DISPATCH_ENABLE_THREAD_POOL then ?
Or are we missing some initialization call ?

MadCoder:
Linux with libpthread workqueues is supposed to have `HAVE_PTHREAD_WORKQUEUES` and then fall in the code below:

```C
#if HAVE_PTHREAD_WORKQUEUES
#if DISPATCH_USE_PTHREAD_POOL
	if (qc->dgq_kworkqueue != (void*)(~0ul))
#endif
	{
```

That's what I was saying above, for you `qc->dgq_kworkqueue` seems to be `~0ul` and that's your real problem.

MadCoder:
If you have pthread workqueues, the only reason to go in the pthread based pool wakeup codepath is for uses of `dispatch_pthread_root_queue_create()` which is a private interface that Linux will probably not use (but maybe some tests exercise that codepath).

root queues initialized that way get a `dgq_ctxt` of their own anyway and should work (or fail to but for other porting reasons than this)

frankeh:
The code path you state is used, however the code asserts ( qc->dgq_kworkqueue == (void*)(~0ul) )
(aka -1).
Will dig through why the queues dgq_kworkqueue are -1 (maybe due to restriction of libpthread_workqueue internally only supporting 4 queue pairs instead of required six.

Regardless, if that is the case, why even do an if statement there.
seems like somewhere we should have an assert(qc->dgq_kworkqueue != (void*)(~0ul))

MadCoder:
because of `dispatch_pthread_root_queue_create()` like I already mentionned.
It is likely because of the 4 pairs instead of 6, libpthread work queue should be fixed to support the two new pairs.

the two new pairs are MAINTENANCE and USER_INTERACTIVE.
The first one should already be mapped to background if WORKQ_FEATURE_MAINTENANCE Is not supported (which libpthread_workqueue should not) so the problem probably comes from the other one.

Short term, USER_INTERACTIVE should be mapped to USER_INITIATED for linux, and do it in a way similar to `WORKQ_FEATURE_MAINTENANCE`.

I think that's your root problem.

frankeh:
OK, but fixing libpthread_workqueue fixes was what we earlier discussed with @das as we 
had fixes for that, but seems we need to pull libpthread_workqueue under libdispatch and change in place then. Seems like both kevent and libpthread_workqueue are not further maintained.
Are you OK with pulling that in once we verify that it fixes that problem (unrolling some changes we made in the initialization).

MadCoder:
libkqueue is another question.
We discussed with @das and we believe that you guys should have a fork of libpthread_workqueue that you maintain and that is incorporated with our repository on master as a git submodule, which let you handle revlocked changes in a simple way. And the code could be linked statically into dispatch when the submodule is present.

Daniel and I feel like this is the best approach fork pthread workqueue. You probably want to ping @ mheily to discuss these altenatives with him.

I'm not keen on pulling the USE/ENABLE PTHREAD POOL change, because we don't want darwin and github to diverge, and this would add more global data for us which affects darwin in a negative way.

MadCoder:
Note that @mheily is active and working on https://github.com/mheily/relaunchd actively and that uses libkqueue so I would expect he has interest in continuing supporting it. libpwq is another story and feels less and less being a good idea to not be revlocked and directly integrated with libdispatch IMO

frankeh:
OK , your wish is our command. We go down that path ... cya

frankeh:
OK .. make a few changes in libpthread_workqueue fixes this and the tests pass without the above pullrequest #17. Will reach out to @mheily to see how to proceed.

MadCoder:
sounds good, closing then.

PR #19, opened 2015-12-18
Enable libbsd for test compilation and cleanup misc. warnings

dgrove-oss:
Make changes in Makefile.am to pass BSD_OVERLAY_CFLAGS
through to test case compilations (excluding dispatch_c99
because header files are not c99 compliant).

Remove stubbed out functions/defines that are now provided
by libbsd.

Include string.h to in bsdtests.h for strrchr.


dgrove-oss:
Thanks for the comment on #18; it helped us realize what the right approach to fixing it was.  (Hubertus and I sitting next to each other and hacking away today...).

MadCoder:
lgtm

PR #21, opened 2015-12-21
Fix warnings.  Add includes to avoid implicit declarations.

ianpartridge:


ianpartridge:
Hi @MadCoder - I've removed the _GNU_SOURCE change for now, and left the include of wait.h (needed by waitpid()) as that is a simple fix.  I may submit a future PR for the _GNU_SOURCE problem.

Please consider merging.  Thanks.

MadCoder:
please resubmit as a squashed single commit as per usual policy and I'll merge, thanks.

_GNU_SOURCE should be added globally through the CFLAGS with configure.ac and looking on google, what you want is to add `AC_GNU_SOURCE` for that effect.

ianpartridge:
Thanks, I will.

PR #27, opened 2016-01-02
Make autogen.sh executable

drewcrawford:
autogen.sh should be executable


das:
FWIW it was my understanding that this was intentionally not executable, the instructions in INSTALL explicitly say to run it via 'sh autogen.sh' for this reason

MadCoder:
why wouldn't it be? it sounds like a weird choice, and not uncommon to have exeutable autogen.sh?

das:
so that checking out the source doesn't add an executable file with arbitrary shell code into . (which some people have in their PATH...). I didn't make the original choice, but I know it was intentional.

MadCoder:
I would argue that people that have `.` in their shell deserve what they get and that most open source projects out there have executable autogen.sh.

I can revert it if you care, but I think it's better that way.

PR #29, opened 2016-01-06
Initialise dispatch_queue_attr_concurrent on Linux

seabaylea:
Three of the tests are failing because the `DISPATCH_QUEUE_CONCURRENT` structure hasn't been initialised (the value for `DISPATCH_QUEUE_CONCURRENT->do_vtable` is `NULL`.

This is occurring because `DISPATCH_QUEUE_CONCURRENT` is aliased to `_dispatch_queue_attrs` in `xcodeconf/libdispatch.aliases`, but aliasing (the use of the `-alias_list` linker option) is only supported on Darwin.

Up until libdispatch-4xx, `DISPATCH_QUEUE_CONCURRENT` was initialised explicitly in init.c. Taking that approach for non-Apple platforms certainly solves the issue, but there may be a better approach to doing the equivalent of aliasing on Linux.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -260,6 +260,13 @@ const struct dispatch_queue_attr_s _dispatch_queue_attrs[]
 	DISPATCH_QUEUE_ATTR_QOS_INITIALIZER(USER_INTERACTIVE),
 };
 
+// _dispatch_queue_attr_concurrent is aliased using libdispatch.aliases
+// and the -alias_list linker option on Darwin but needs to be done manually
+// for other platforms.
+#ifndef __APPLE__
+extern struct dispatch_queue_attr_s _dispatch_queue_attr_concurrent
+	__attribute__((__alias__("_dispatch_queue_attrs")));
+#endif
 
 #pragma mark -
 #pragma mark dispatch_vtables


MadCoder:
ELF supports aliases from gcc/clang.

you should use:

```C
#ifndef __APPLE__
struct dispatch_queue_attr_s _dispatch_queue_attr_concurrent
        __attribute__((__alias__("_dispatch_queue_attrs")));
```

that should work just fine.

MadCoder:
you may need extern on top, like this:

```C
#ifndef __APPLE__
extern struct dispatch_queue_attr_s _dispatch_queue_attr_concurrent
        __attribute__((__alias__("_dispatch_queue_attrs")));
#endif
```

I'll let you play with it until you find the right combination that works, glibc does this everywhere in its source

seabaylea:
Thanks - I'll do some experimenting and then update the PR.

MadCoder:
I'll ask you to submit a new one so that we get a single patch once you get the right one though, but let leaev that one open until you submit the new one so that nobody tries to fix that concurrently ;)

seabaylea:
I can just squash the commits under this PR if that helps?

MadCoder:
I don't think github lets you do that :) but if it allows for it, sure go ahead!

seabaylea:
Looks like your suggestion was all that was needed. I've built/tested and updated the commit in the PR (by squashing the commits and doing a force push to the branch).

PR #33, opened 2016-01-11
Fixes to enable DISPATCH_DEBUG on Linux

dgrove-oss:
Small fixes to code guarded under DISPATCH_DEBUG to
fix Linux compilation issues.


src/source.c
```diff
@@ -4816,7 +4820,12 @@ _dispatch_kevent_debug(const _dispatch_kevent_qos_s* kev, const char* str)
 			"ext[0] = 0x%llx, ext[1] = 0x%llx }: %s", kev, kev->ident,
 			_evfiltstr(kev->filter), _evflagstr(kev->flags, flagstr,
 			sizeof(flagstr)), kev->flags, kev->fflags, kev->data, kev->udata,
-			kev->ext[0], kev->ext[1], str);
+#ifdef __linux__
```
MadCoder: should be `#if DISPATCH_USE_KEVENT_QOS` I think.

dgrove-oss:
I force pushed an amended commit to address comments by Pierre and Ian.

MadCoder:
woops sorry I let that one slide, for some reason I have had missing notifications and just realized there were several pending reviews

PR #36, opened 2016-01-27
fix _dispatch_operation_advise as functionality is not supported in l…

frankeh:
…inux


MadCoder:
Why not use `readahead(2)` that would be the right way to port?

frankeh:
OK, let me look at that tonight .. so hold this request for now.

frankeh:
You are right, having read through the documentation, that fits the bill and lets us reuse the core of the function. I submit a different request.

PR #37, opened 2016-01-27
libpwq as git submodule

dgrove-oss:
On Linux we need recent updates to libpthread_workqueue
to support 6 priority levels instead of 4.  To ensure
we get a compatible libpwq version, use git submodules
to get the right source version, change the autotools
to build it as a substep of building libdispatch, and
statically link the libpwq object files into libdispatch.so
to avoid picking up a system libpthread_workqueue.so via
dynamic linking.

The pieces of this commit are:
 (1) update INSTALL instrictions.
 (2) update configure.ac, Makefile.am to support build
 (3) bind submodule to dgrove-oss/libpwq until pending
     pull requests are merged upsteam into mheily/libpwq
     and we are satisfied that we will not need to make
     further adjustments to libpwq.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "libpwq"]
+	path = libpwq
+	url = https://github.com/dgrove-oss/libpwq.git
```
MadCoder: wasn't this supposed to be upstream now, or am I confused?

Makefile.am
```diff
@@ -4,13 +4,24 @@
 
 ACLOCAL_AMFLAGS = -I m4
 
+if BUILD_OWN_PTHREAD_WORKQUEUES
```
MadCoder: I would rather write this:

```
if BUILD_OWN_PTHREAD_WORKQUEUES
PTHREAD_WORKQUES_SUBDIR = libpwq
else
PTHREAD_WORKQUES_SUBDIR =
endif
```

and add that variable in SUBDIRS, else maintaining the list below will be a pain. or is it not working?

MadCoder:
Beside my two questions, patch looks reasonnable

Makefile.am
dgrove-oss: I had tried variations on this (and I tried this variation just now), but annoyingly it doesn't quite work.  The autoreconf and configure steps work correctly.  But in the generated top-level Makefile we end up with the SUBDIRS definition with the PTHREAD_WORKQUEUE_SUBDIR variable defined textually before PTHREAD_WORKQUEUE_SUBDIR.  This causes 'make' to not recurse into libpwq and the library doesn't get built.

I'm certainly not an autotools guru, but I wasn't able to get anything to work where SUBDIRS included a conditionally defined variable.

MadCoder: ok, ISTR that SUBDIRS is magical and a pain so fine. we'll live with that.

.gitmodules
dgrove-oss: When I did the integration with swift/utils/build-script last week I found some autotools fixes (similar issues as #34 in libdispatch) that need to get upstreamed to libpwq.  I have a unmerged pull request open (https://github.com/mheily/libpwq/pull/8) but in the short run I think we should go with my fork until everything is working & integrated into the swift build (there may be more fixes to libpwq that we don't know about yet).

MadCoder: fair enough

PR #39, opened 2016-01-28
Support for file readahead under linux using readahead

frankeh:
Reusing the base structure of the function to maintain the offset logic
and replace fcntl( F_RDAVICE ) with readahead under linux.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -2066,9 +2066,14 @@ _dispatch_disk_perform(void *ctxt)
 static void
 _dispatch_operation_advise(dispatch_operation_t op, size_t chunk_size)
 {
-#ifndef F_RDADVISE
-  LINUX_PORT_ERROR();
-#else
+#ifdef __linux__
+	// linux does not support fcntl (F_RDAVISE)
+	// define necessary datastructure and use readahead
+	struct radvisory {
+		off_t ra_offset;
+		int   ra_count;
+	};
+#endif
 	int err;
 	struct radvisory advise;
 	// No point in issuing a read advise for the next chunk if we are already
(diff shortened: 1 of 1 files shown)


src/io.c
```diff
@@ -2088,6 +2093,14 @@ _dispatch_operation_advise(dispatch_operation_t op, size_t chunk_size)
 	}
 	advise.ra_offset = op->advise_offset;
 	op->advise_offset += advise.ra_count;
+#ifdef __linux__
+	_dispatch_io_syscall_switch(err,
+		readahead(op->fd_entry->fd, advise.ra_offset, advise.ra_count),
+		case EBADF: break; // bad file desciptor
```
MadCoder: Ignoring `EBADF` is wrong it should be caught by the `dispatch_assume_zero()` below to log as this is a programming error.

src/io.c
```diff
@@ -2066,9 +2066,14 @@ _dispatch_disk_perform(void *ctxt)
 static void
 _dispatch_operation_advise(dispatch_operation_t op, size_t chunk_size)
 {
-#ifndef F_RDADVISE
-  LINUX_PORT_ERROR();
-#else
+#ifdef __linux__
+    // radvisory and F_RDAVICE is not support, to reuse code
```
MadCoder: is not support**ed**

MadCoder:
Two nits, please directly squash the fixes and force-push your pull request, there's no need to create a new one.

src/io.c
frankeh: In that case EBADF and EINVAL both should be ignored, since EINVAL 
recognizes filedescriptor on which this operation can not be applied ( e.g. socket ).

frankeh: I mean both should NOT be ignored and caught by dispatch_assume_zero()

PR #43, opened 2016-02-03
Initial integration of Swift overlay into libdispatch build

dgrove-oss:
This commit adds a src/swift directory that contains
Dispatch.swift and Dispatch.c (based on existing files in
swift/stdlib/public/SDK/Dispatch) and a module.map.in template.
We made some initial changes to Dispatch.swift and Dispatch.c to
adapt them to Linux by dropping MACH API functions/adapters.

The libdispatch build process is extended to include Dispatch.o
in libdispatch.so and to generate Dispatch.swiftmodule,
Dispatch.swiftdoc and a module.map file when configure is
given a --with-swift-sdk argument.

Swift programs that import Dispatch can be compiled, linked,
and run against the generated files, but must include some
extra command line arguments due to the use of @convention(block)
to bridge from Swift blocks to the -fblocks expected by dispatch.
In particular, the following flags are currently needed to compile
a Swift program: -Xcc -fblocks  -Xlinker -lBlocksRuntime

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -101,6 +101,20 @@ AS_IF([test "x$enable_apple_tsd_optimizations" = "xyes"],
     [Define to use non-portable pthread TSD optimizations for Mac OS X)])]
 )
 
+#
+# Enable building Swift overlay support into libdispatch
+#
+AC_ARG_WITH([swift-toolchain],
+  [AS_HELP_STRING([--with-swift-toolchain], [Specify path to Swift toolchain])],
+  [swift_toolchain_path=${withval}
+   AC_DEFINE(HAVE_SWIFT, 1, [Define if building for Swift])
+   SWIFTC="$swift_toolchain_path/bin/swiftc"
+   have_swift=true],
+  [have_swift=false]
+)
+AM_CONDITIONAL(HAVE_SWIFT, $have_swift)
+AC_SUBST([SWIFTC])
+
 AC_USE_SYSTEM_EXTENSIONS
 AM_INIT_AUTOMAKE([foreign no-dependencies subdir-objects])
 LT_INIT([disable-static])
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -1,5 +1,9 @@
 provider.h
+module.map
+module.build.map
 .libs
 *.lo
 *.la
-
+*.o
+*.swiftmodule
+*.swiftdoc
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -46,6 +46,9 @@ libdispatch_la_SOURCES=		\
 	shims/tsd.h				\
 	shims/yield.h
 
+EXTRA_libdispatch_la_SOURCES=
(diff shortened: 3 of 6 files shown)


billabt:
In Dispatch.swift, `typealias` is deprecated, the new term is `associatedtype`. So, line 22:
```
public typealias dispatch_block_t = @convention(block) () -> Void
```
should be this instead...
```
public associatedtype dispatch_block_t = @convention(block) () -> Void
```

ddunbar:
Do we have a plan yet for how to get the -fblocks flag added when appropriate for Swift?

For the -lBlocksRuntime argument, you may be able to deal with this by adding a link declaration to your module map:
  http://clang.llvm.org/docs/Modules.html#link-declaration

mwwa:
It's unfortunate that all the C-defines need computed variables to be shimmed over but it looks like that's also a direct mirror of the stdlib version of this code.

mwwa:
I mirror ddunbar's suggestion about using the module link line to try including BlocksRuntime. I'm not in a position to check if that works on Linux myself though.

dgrove-oss:
@billabt associatedtype doesn't work in this context (not sure if this is by design, or a bug).  Using a swiftc I built from git as of this morning, I get:
`/home/vagrant/test-build/../swift-corelibs-libdispatch/src/swift/Dispatch.swift:22:8: error: associated types can only be defined in a protocol; define a type or introduce a 'typealias' to satisfy an associated type requirement
public associatedtype dispatch_block_t = @convention(block) () -> Void
       ^
/home/vagrant/test-build/../swift-corelibs-libdispatch/src/swift/Dispatch.swift:22:43: error: attribute can only be applied to types, not declarations
public associatedtype dispatch_block_t = @convention(block) () -> Void
                                          ^
/home/vagrant/test-build/../swift-corelibs-libdispatch/src/swift/Dispatch.swift:22:53: error: expected declaration
public associatedtype dispatch_block_t = @convention(block) () -> Void
`

dgrove-oss:
Thanks @mwwa and @ddunbar for the linker suggestion.  I had missed that I could have multiple linker directives in the module.map file.   I had tried putting both dispatch and BlocksRuntime into a single linker directive, which didn't work.   Updated module.map.in to have two linker directives, which does work.

dgrove-oss:
Squashed changes into a single commit and updated commit comment to reflect now only needing -Xcc -fblocks.

dgrove-oss:
Updated & squashed with a change to the command line argument of configure to --with-swift-toolchain as requested in the review of https://github.com/apple/swift/pull/1212

dgrove-oss:
@MadCoder @mwwa.   Any additional changes needed?  Would like to get this merged so we can continue with next steps.  Thanks.

daagaak:
No I'm happy, I'll merge momentarily.

mwwa:
No I'm happy, I'll merge momentarily.

PR #44, opened 2016-02-04
introduce usage of thread local storage structure

frankeh:
Reworked the gettid patch as discussed #42.

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -956,7 +956,7 @@ _libdispatch_tsd_cleanup(void *ctx)
 	tsd->initialized = false;
 }
 
-DISPATCH_EXPORT
+DISPATCH_EXPORT DISPATCH_NOINLINE
 void
 libdispatch_tsd_init(void)
 {
--- a/src/queue.c
+++ b/src/queue.c
@@ -913,6 +913,52 @@ _dispatch_root_queues_init(void *context DISPATCH_UNUSED)
 
 #define countof(x) (sizeof(x) / sizeof(x[0]))
 
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+#include <unistd.h>
+#include <sys/syscall.h>
+
+#ifdef SYS_gettid
+static inline pid_t gettid() { return (pid_t) syscall(SYS_gettid); }
+#else
+#error "SYS_gettid unavailable on this system"
+#endif
+
+#define _tsd_call_cleanup(k,f)  \
+	do { if (f && __dispatch_tsd.k) ((void(*)(void*))(f))(__dispatch_tsd.k); \
+    } while (0)
+
+static
+void
+_libdispatch_thread_cleanup(void)
+{
+	_tsd_call_cleanup(dispatch_queue_key, _dispatch_queue_cleanup);
+	_tsd_call_cleanup(dispatch_voucher_key, _voucher_thread_cleanup);
+	_tsd_call_cleanup(dispatch_cache_key, _dispatch_cache_cleanup);
+	_tsd_call_cleanup(dispatch_io_key, NULL);
+	_tsd_call_cleanup(dispatch_apply_key, NULL);
+	_tsd_call_cleanup(dispatch_defaultpriority_key, NULL);
+	_tsd_call_cleanup(dispatch_pthread_root_queue_observer_hooks_key,
+			NULL);
+#if DISPATCH_PERF_MON && !DISPATCH_INTROSPECTION
+	_tsd_call_cleanup(dispatch_bcounter_key, NULL);
+#endif
+#if !DISPATCH_USE_OS_SEMAPHORE_CACHE
+	_tsd_call_cleanup(dispatch_sema4_key,
+			(void (*)(void *))_dispatch_thread_semaphore_dispose);
+#endif
+}
+
+DISPATCH_EXPORT
+void
+libdispatch_thread_init(void)
+{
+	__dispatch_tsd.tid = gettid();
+   pthread_setspecific(__dispatch_tsd_key, &__dispatch_tsd);
+   __dispatch_tsd.initialized = true;
+}
+
+#endif
+
 DISPATCH_EXPORT DISPATCH_NOTHROW
 void
 libdispatch_init(void)
@@ -951,6 +997,9 @@ libdispatch_init(void)
 	dispatch_assert(sizeof(struct dispatch_root_queue_context_s) %
 			DISPATCH_CACHELINE_SIZE == 0);
 
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+	_dispatch_thread_key_create(&__dispatch_tsd_key, (void(*)(void*))_libdispatch_thread_cleanup);
+#else
 	_dispatch_thread_key_create(&dispatch_queue_key, _dispatch_queue_cleanup);
 	_dispatch_thread_key_create(&dispatch_voucher_key, _voucher_thread_cleanup);
 	_dispatch_thread_key_create(&dispatch_cache_key, _dispatch_cache_cleanup);
@@ -965,6 +1014,7 @@ libdispatch_init(void)
 #if !DISPATCH_USE_OS_SEMAPHORE_CACHE
 	_dispatch_thread_key_create(&dispatch_sema4_key,
 			(void (*)(void *))_dispatch_thread_semaphore_dispose);
+#endif
 #endif
 
 #if DISPATCH_USE_RESOLVERS // rdar://problem/8541707
--- a/src/queue.c
+++ b/src/queue.c
@@ -918,19 +918,25 @@ _dispatch_root_queues_init(void *context DISPATCH_UNUSED)
 #include <sys/syscall.h>
 
 #ifdef SYS_gettid
-static inline pid_t gettid() { return (pid_t) syscall(SYS_gettid); }
+DISPATCH_ALWAYS_INLINE
+static inline pid_t
+gettid(void)
+{
+	return (pid_t) syscall(SYS_gettid);
+}
 #else
 #error "SYS_gettid unavailable on this system"
 #endif
 
-#define _tsd_call_cleanup(k,f)  \
-	do { if (f && __dispatch_tsd.k) ((void(*)(void*))(f))(__dispatch_tsd.k); \
+#define _tsd_call_cleanup(k, f)  do { \
+		if (f && tsd->k) ((void(*)(void*))(f))(tsd->k); \
     } while (0)
 
-static
-void
-_libdispatch_thread_cleanup(void)
+static void
+_libdispatch_tsd_cleanup(void *ctx)
 {
+	struct dispatch_tsd *tsd = (struct dispatch_tsd*) ctx;
+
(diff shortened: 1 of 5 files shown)


src/queue.c
```diff
@@ -913,6 +913,52 @@ _dispatch_root_queues_init(void *context DISPATCH_UNUSED)
...
+#else
+#error "SYS_gettid unavailable on this system"
+#endif
+
+#define _tsd_call_cleanup(k,f)  \
+	do { if (f && __dispatch_tsd.k) ((void(*)(void*))(f))(__dispatch_tsd.k); \
+    } while (0)
+
+static
+void
```
MadCoder: should be on the same line

src/queue.c
```diff
@@ -913,6 +913,52 @@ _dispatch_root_queues_init(void *context DISPATCH_UNUSED)
 
 #define countof(x) (sizeof(x) / sizeof(x[0]))
 
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+#include <unistd.h>
+#include <sys/syscall.h>
+
+#ifdef SYS_gettid
+static inline pid_t gettid() { return (pid_t) syscall(SYS_gettid); }
```
MadCoder: please conform to coding style, and don't forget the `void`

```C
DISPATCH_ALWAYS_INLINE
static inline pid_t
gettid(void)
{
```
...

src/queue.c
```diff
@@ -913,6 +913,52 @@ _dispatch_root_queues_init(void *context DISPATCH_UNUSED)
...
+#else
+#error "SYS_gettid unavailable on this system"
+#endif
+
+#define _tsd_call_cleanup(k,f)  \
+	do { if (f && __dispatch_tsd.k) ((void(*)(void*))(f))(__dispatch_tsd.k); \
+    } while (0)
```
MadCoder: Style would rather be:

```C
#define _tsd_call_cleanup(k, f)  do { \
	if (f && __dispatch_tsd.k) ((void(*)(void*))(f))(__dispatch_tsd.k); \
    } while (0)
```

why the cast? it looks gross

src/queue.c
```diff
@@ -913,6 +913,52 @@ _dispatch_root_queues_init(void *context DISPATCH_UNUSED)
...
+#endif
+#if !DISPATCH_USE_OS_SEMAPHORE_CACHE
+	_tsd_call_cleanup(dispatch_sema4_key,
+			(void (*)(void *))_dispatch_thread_semaphore_dispose);
+#endif
+}
+
+DISPATCH_EXPORT
+void
+libdispatch_thread_init(void)
```
MadCoder: `_libdispatch_tsd_init` look like a better name

src/queue.c
```diff
@@ -913,6 +913,52 @@ _dispatch_root_queues_init(void *context DISPATCH_UNUSED)
...
+#error "SYS_gettid unavailable on this system"
+#endif
+
+#define _tsd_call_cleanup(k,f)  \
+	do { if (f && __dispatch_tsd.k) ((void(*)(void*))(f))(__dispatch_tsd.k); \
+    } while (0)
+
+static
+void
+_libdispatch_thread_cleanup(void)
```
MadCoder: `_libdispatch_tsd_cleanup` looks better

src/queue.c
```diff
@@ -913,6 +913,52 @@ _dispatch_root_queues_init(void *context DISPATCH_UNUSED)
...
+	_tsd_call_cleanup(dispatch_apply_key, NULL);
+	_tsd_call_cleanup(dispatch_defaultpriority_key, NULL);
+	_tsd_call_cleanup(dispatch_pthread_root_queue_observer_hooks_key,
+			NULL);
+#if DISPATCH_PERF_MON && !DISPATCH_INTROSPECTION
+	_tsd_call_cleanup(dispatch_bcounter_key, NULL);
+#endif
+#if !DISPATCH_USE_OS_SEMAPHORE_CACHE
+	_tsd_call_cleanup(dispatch_sema4_key,
+			(void (*)(void *))_dispatch_thread_semaphore_dispose);
```
MadCoder: why the cast?

src/queue.c
```diff
@@ -913,6 +913,52 @@ _dispatch_root_queues_init(void *context DISPATCH_UNUSED)
...
+#endif
+}
+
+DISPATCH_EXPORT
+void
+libdispatch_thread_init(void)
+{
+	__dispatch_tsd.tid = gettid();
+   pthread_setspecific(__dispatch_tsd_key, &__dispatch_tsd);
+   __dispatch_tsd.initialized = true;
```
MadCoder: space versus tabs

src/queue.c
```diff
@@ -951,6 +997,9 @@ libdispatch_init(void)
 	dispatch_assert(sizeof(struct dispatch_root_queue_context_s) %
 			DISPATCH_CACHELINE_SIZE == 0);
 
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+	_dispatch_thread_key_create(&__dispatch_tsd_key, (void(*)(void*))_libdispatch_thread_cleanup);
```
MadCoder: your function takes `(void)` which is why you need the cast and that is not the right way to do it.

`_libdispatch_thread_cleanup` should be:

```C
static void
_libdispatch_tsd_cleanup(void *ctxt DISPATCH_UNUSED)
{
...
```

And actually I think the context shouldn't be "unused", even if you "know" its address, it's cleaner to cast the `ctxt` into your struct type and clean this instead of reloading the __thread thing.

src/shims/linux_stubs.c
```diff
@@ -52,9 +52,15 @@ char* mach_error_string(mach_msg_return_t x) {
 }
 void mach_vm_deallocate() { LINUX_PORT_ERROR();  }
 
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+mach_port_t pthread_mach_thread_np(void) {
```
MadCoder: not the right patch, you should instead re-define _dispatch_thread_port() to be the tid that you computed.

also __dispatch_tsd.tid may not be initialized, seems wrong.

src/shims/linux_stubs.c
```diff
@@ -52,9 +52,15 @@ char* mach_error_string(mach_msg_return_t x) {
...
+  return __dispatch_tsd.tid;
+}
+#else
 mach_port_t pthread_mach_thread_np(void) {
```
MadCoder: and you can probably get rid of that guy

src/shims/tsd.h
```diff
@@ -73,6 +73,56 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+#elif DISPATCH_USE_THREAD_LOCAL_STORAGE
+
+DISPATCH_TSD_INLINE
+static inline void
+_dispatch_thread_key_create(pthread_key_t *k, void (*d)(void *))
+{
+	dispatch_assert_zero(pthread_key_create(k, d));
+}
+
+struct dispatch_tsd {
```
MadCoder: style nits for this struct:
- spaces/tabs
- `void* dispatch_queue_key` -> `void *dispatch_queue_key`

src/shims/tsd.h
```diff
@@ -73,6 +73,56 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+extern __thread struct dispatch_tsd __dispatch_tsd;
+extern pthread_key_t __dispatch_tsd_key;
+extern void libdispatch_thread_init(void);
+
+DISPATCH_ALWAYS_INLINE
+static inline
+struct dispatch_tsd *
+_dispatch_get_tsd_base(void)
+{
+    if (slowpath(!__dispatch_tsd.initialized))
```
MadCoder: please add braces.

`if (a) some_short_stmt;` is sometimes ok if `a` is short, but if you use multiline, always use braces
and here, multiline is the right way

src/shims/tsd.h
```diff
@@ -73,6 +73,56 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+    void* dispatch_pthread_root_queue_observer_hooks_key;
+};
+
+extern __thread struct dispatch_tsd __dispatch_tsd;
+extern pthread_key_t __dispatch_tsd_key;
+extern void libdispatch_thread_init(void);
+
+DISPATCH_ALWAYS_INLINE
+static inline
+struct dispatch_tsd *
```
MadCoder: should be on the same line, just to help the reader/compiler you may want:

```C
DISPATCH_ALWAYS_INLINE DISPATCH_CONST
static inline struct dispatch_tsd *
```

MadCoder:
Additional comments:
I think you're missing patches, at least in init.c. Look for `DISPATCH_USE_DIRECT_TSD` in the source, it's likely that you will want to also have `DISPATCH_USE_THREAD_LOCAL_STORAGE` clauses around them too. e.g. in init.c you have (may be slightly different 

```C
#if !DISPATCH_USE_DIRECT_TSD
pthread_key_t dispatch_queue_key;
pthread_key_t dispatch_sema4_key;
...
```

that should definitely become:
```diff
-#if !DISPATCH_USE_DIRECT_TSD
+#if !DISPATCH_USE_DIRECT_TSD && ! DISPATCH_USE_THREAD_LOCAL_STORAGE
pthread_key_t dispatch_queue_key;
pthread_key_t dispatch_sema4_key;
...
```

Or this:

```C
#if DISPATCH_USE_DIRECT_TSD
const struct dispatch_tsd_indexes_s dispatch_tsd_indexes = {
	.dti_version = 2,
	.dti_queue_index = dispatch_queue_key,
	.dti_voucher_index = dispatch_voucher_key,
	.dti_qos_class_index = dispatch_priority_key,
};
#else // DISPATCH_USE_DIRECT_TSD
#ifndef __LINUX_PORT_HDD__
#error Not implemented on this platform
#endif
#endif // DISPATCH_USE_DIRECT_TSD
```

looks wrong, I think it's fine for a platform to not have some bits of the introspection that make no sense for the platform. Here I would just remove the `#error` for everyone (though @das may disagree?).

Overall this is going in the right direction, and I think it will be a tremendous performance improvement on linux.

Also, @das will need to review that one too, he'll probably have comments on top of mine

src/queue.c
frankeh: to handle calls like this:
    _tsd_call_cleanup(dispatch_io_key, NULL);
in the _libdispatch_thread_cleanup.
One can argue I don't need that anyway, since it will be optimized out (see macro), but
I wanted it there to get some symmetry in the code for TSD and non-TSD code.
If you prefer I don't call the NULL instances of these I can simplify the macro.
Also look at the "dispatch_sema4_key  K,V" .. same ugliness there in the code by default"

src/queue.c
frankeh: code take from other case to maintain code symmetry:
e.g.
#if !DISPATCH_USE_OS_SEMAPHORE_CACHE
    _dispatch_thread_key_create(&dispatch_sema4_key,
            (void (*)(void *))_dispatch_thread_semaphore_dispose);
#endif

frankeh:
@das Is there anything else you would like addressed or you have concerns with
or can we merge

das:
sorry have not yet had time to look

MadCoder:
replaced with #82

PR #46, opened 2016-02-08
Add support for CFRunLoop integration APIs

seabaylea:
Work has already been done by @kovpas and merged by @parkera to add a CFRunLoop implementation for Linux into CoreFoundation by PR #133 (https://github.com/apple/swift-corelibs-foundation/pull/133).

This followed the model in CFLite of using `eventfd` rather than mach ports, and calling the following two (new) APIs:
```
dispatch_get_main_queue_eventfd_np
dispatch_main_queue_drain_np
```
instead of the existing "4CF" calls:
```
_dispatch_get_main_queue_port_4CF
_dispatch_main_queue_callback_4CF
```
Those APIs don't exist in libdispatch, so this adds them. The APIs are defined in `queue.h` which effectively makes then public - that's been done because it matched the approach used by the CoreFoundation PR. If it makes sense to make them "4CF" functions I can modify the PR here and open an equivalent against CoreFoundation.

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -75,12 +75,18 @@ static void *_dispatch_worker_thread(void *context);
 static int _dispatch_pthread_sigmask(int how, sigset_t *set, sigset_t *oset);
 #endif
 
+#if DISPATCH_ENABLE_RUNLOOP_SUPPORT
+static dispatch_queue_t _dispatch_main_queue_wakeup(void);
 #if DISPATCH_COCOA_COMPAT
 static dispatch_once_t _dispatch_main_q_port_pred;
-static dispatch_queue_t _dispatch_main_queue_wakeup(void);
 unsigned long _dispatch_runloop_queue_wakeup(dispatch_queue_t dq);
 static void _dispatch_runloop_queue_port_init(void *ctxt);
 static void _dispatch_runloop_queue_port_dispose(dispatch_queue_t dq);
+#elif DISPATCH_LINUX_COMPAT
+static dispatch_once_t _dispatch_main_q_eventfd_pred;
+static void _dispatch_main_q_eventfd_init(void *ctxt);
+static int main_q_eventfd = -1;
+#endif
 #endif
 
 static void _dispatch_root_queues_init(void *context);
@@ -3466,6 +3472,7 @@ _dispatch_wakeup(dispatch_object_t dou)
 				// probe does
 }
 
+#if DISPATCH_ENABLE_RUNLOOP_SUPPORT
 #if DISPATCH_COCOA_COMPAT
 static inline void
 _dispatch_runloop_queue_wakeup_thread(dispatch_queue_t dq)
@@ -3493,6 +3500,7 @@ _dispatch_runloop_queue_wakeup(dispatch_queue_t dq)
 	_dispatch_runloop_queue_wakeup_thread(dq);
 	return false;
 }
+#endif
 
 DISPATCH_NOINLINE
 static dispatch_queue_t
@@ -3502,9 +3510,21 @@ _dispatch_main_queue_wakeup(void)
 	if (!dq->dq_is_thread_bound) {
 		return NULL;
 	}
+#if DISPATCH_COCOA_COMPAT
 	dispatch_once_f(&_dispatch_main_q_port_pred, dq,
 			_dispatch_runloop_queue_port_init);
 	_dispatch_runloop_queue_wakeup_thread(dq);
+#elif DISPATCH_LINUX_COMPAT
+	dispatch_once_f(&_dispatch_main_q_eventfd_pred, dq,
+			_dispatch_main_q_eventfd_init);
+	if (main_q_eventfd != -1) {
+		int result;
+		do {
+			result = eventfd_write(main_q_eventfd, 1);
+		} while (result == -1 && errno == EINTR);
+		(void)dispatch_assume_zero(result);
+	}
+#endif
 	return NULL;
 }
 #endif
@@ -3752,7 +3772,7 @@ out:
 	return sema;
 }
 
-#if DISPATCH_COCOA_COMPAT
+#if DISPATCH_ENABLE_RUNLOOP_SUPPORT
 static void
 _dispatch_main_queue_drain(void)
 {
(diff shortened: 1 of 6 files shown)


src/queue.c
```diff
@@ -4498,20 +4550,36 @@ _dispatch_get_main_queue_port_4CF(void)
...
+{
+	if (main_q_is_draining) {
+		return;
+	}
+	_dispatch_queue_set_mainq_drain_state(true);
+	_dispatch_main_queue_drain();
+	_dispatch_queue_set_mainq_drain_state(false);
+}
 
-// 6618342 Contact the team that owns the Instrument DTrace probe before
```
MadCoder: not ok to remove a comment that says something quite explicit about it not being removed ever.

src/shims/pthread_main_np.h
```diff
@@ -0,0 +1,40 @@
...
+#if __linux__
+#include <unistd.h>
+#include <sys/syscall.h>
+#include <sys/types.h>
+#endif
+
+static inline int
+pthread_main_np()
+{
+#if __linux__
```
MadCoder: this is very slow, you should instead base this patch on top of the `__thread` infrastructure that is built built.

dispatch/queue.h
```diff
@@ -990,6 +990,54 @@ DISPATCH_EXPORT DISPATCH_PURE DISPATCH_WARN_RESULT DISPATCH_NOTHROW
 void *
 dispatch_get_specific(const void *key);
 
+#if __linux__
```
MadCoder: this should NOT be API but private interface in queue_private, and I don't understand why using the _4CF hooks isn't good enough here.

MadCoder:
I disagree with that patch, I think it should not add new interfaces for the sake of linux when the `_4CF` things are just fine. If you diverge too much from the Mac OS X model you will have pain keeping in sync with our changes that keep iterating on how these interfaces work.

src/queue.c
```diff
@@ -23,6 +23,11 @@
 #include "protocol.h"
 #endif
 
+#if __linux__
```
MadCoder: should be in "internal.h"

src/queue.c
```diff
@@ -77,12 +82,21 @@ static int _dispatch_pthread_sigmask(int how, sigset_t *set, sigset_t *oset);
 
 #if DISPATCH_COCOA_COMPAT
 static dispatch_once_t _dispatch_main_q_port_pred;
-static dispatch_queue_t _dispatch_main_queue_wakeup(void);
 unsigned long _dispatch_runloop_queue_wakeup(dispatch_queue_t dq);
 static void _dispatch_runloop_queue_port_init(void *ctxt);
 static void _dispatch_runloop_queue_port_dispose(dispatch_queue_t dq);
 #endif
 
+#if DISPATCH_COCOA_COMPAT || DISPATCH_LINUX_COMPAT
```
MadCoder: `DISPATCH_LINUX_COMPAT` is not the right name for this, it is about providing runloop backends, you should introduce a `DISPATCH_ENABLE_RUNLOOP_SUPPORT` and make sure that `DISPATCH_COCOA_COMPAT` implies it.

then fix all places that need it to use that new guard.

dispatch/queue.h
parkera: Although I agree that this should not be in the public header, I do think it makes sense to make dedicated entry points for this like we also have on Windows. That means that we can change them independently from each other without having to worry about binary compatibility.

src/queue.c
parkera: This just moved up a few lines.

MadCoder:
For the sake of porting I think the right way is like I said, have:

1. proper guards (suggested `DISPATCH_ENABLE_RUNLOOP_SUPPORT`)
2. same function names `_4CF`, which is not a lie, it's really about giving support to CFRunloop
3. abstract internally the "port" and its use away but do not touch _dispatch_main_queue_wakeup, only what it calls. The structure should completely be the same.

seabaylea:
I'm happy to rename the guards and move to implementing `_4CF` functions. 

As per my initial comment, I took the approach of implementing the interface functions as  `dispatch_get_main_queue_eventfd_np` and `dispatch_main_queue_drain_np` because `CoreFoundation/RunLoop.subproj/CFRunLoop.c` had already been updated to use them.

Presumably we'll new some new `_4CF` functions for Linux as the consuming CFRunLoop code (currently) needs to know its receiving back a eventfd file descriptor rather than a "port".

seabaylea:
@MadCoder I've made an interaction of changes that moves towards what your looking for - they're not complete but I'd like to get some feedback before doing much more.

For the CFRunLoop entry points, I've retained retained Linux specific APIs, but moved the declaration to `queue_internal.h` and renamed them as follows:
```
int dispatch_get_main_queue_eventfd_4CF()
void _dispatch_main_queue_drain_4CF()
```
The equivalents for Darwin and Windows return a "port"/Handle, and the "drain" function for Darwin has the port as a parameter. Using shared `_4CF` functions could be achieved but would require a sizeable refactoring of the interface between CFRunLoop and libdispatch. CFRunLoop currently handles the fact that it it working with a port/Handle/eventfd for notifications whereas that would all have to be pushed down to libdispatch.
Does this seem reasonable for now? Should the definitions be moved from queue_internal.h to private.h?

I've introduced a new `DISPATCH_ENABLE_RUNLOOP_SUPPORT` guard, but had to retain the use of `DISPATCH_COCOA_COMPAT` and `DISPATCH_LINUX_COMPAT` for the Darwin/Windows and Linux specific sections. I'm not sure if this is the best approach - and feedback is appreciated.

I've also left the implementation of `pthread_main_np` as-is for the moment, as that's dependent on PR #44 which isn't yet merged.

seabaylea:
Hi @MadCoder . Can you take a second look at this and give some recommendations on how we can move forward?

seabaylea:
Hi @MadCoder . Have you been able to do another review of this yet?

MadCoder:
Replaced by #101

src/shims/pthread_main_np.h
colemancda: It is indeed slow. Right now I am building a CFRunLoop based GUI in Swift, and with no input added to the runloop I am getting 0ms per loop on Mac and 17ms  on Linux. With actual input im getting up to 3000ms only on Linux, crazy.

colemancda: Please check https://github.com/PureSwift/Cacao/commit/168b18d583deec82987e9ea0e84d108eac554581

PR #50, opened 2016-02-15
add futex semaphore support derived from hutchinson port

frankeh:


src/internal.h
```diff
@@ -242,6 +246,34 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
 #include <unistd.h>
 #endif
 
+#if USE_POSIX_SEM
+#include <semaphore.h>
+#elif USE_FUTEX_SEM
+#include <sys/syscall.h>
+#include <linux/futex.h>
+#endif
```
MadCoder: this block is duplicated (249-254)

src/internal.h
```diff
@@ -242,6 +246,34 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
...
+
+#if defined(__APPLE__)
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
+#define DISPATCH_LITTLE_ENDIAN
+#elif __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+#define DISPATCH_BIG_ENDIAN
+#endif
+
```
MadCoder: remove extra line

src/internal.h
```diff
@@ -242,6 +246,34 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
...
+#elif defined(__linux__)
+#include <endian.h>
+#if __BYTE_ORDER == __LITTLE_ENDIAN
+#define DISPATCH_LITTLE_ENDIAN
+#elif __BYTE_ORDER == __BIG_ENDIAN
+#define DISPATCH_BIG_ENDIAN
+#endif
+
```
MadCoder: extra line

src/internal.h
```diff
@@ -242,6 +246,34 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
...
+#endif  /* defined(__APPLE__) */
```
MadCoder: add
```C
#else
#error please implement DISPATCH_LITTLE/BIG_ENDIAN for your platform
#endif
```

src/semaphore.c
```diff
@@ -352,6 +368,18 @@ _dispatch_semaphore_wait_slow(dispatch_semaphore_t dsema,
 			DISPATCH_SEMAPHORE_VERIFY_RET(ret);
 			break;
 		}
+#elif USE_FUTEX_SEM
+		do {
+			uint64_t nsec = _dispatch_time_to_nanoseconds(timeout);
+			_timeout.tv_sec = (typeof(_timeout.tv_sec))(nsec / NSEC_PER_SEC);
+			_timeout.tv_nsec = (typeof(_timeout.tv_nsec))(nsec % NSEC_PER_SEC);
+            ret = slowpath(_dispatch_futex_wait(&dsema->dsema_futex, &_timeout));
```
MadCoder: spaces/tabs

configure.ac
```diff
@@ -284,6 +286,9 @@ AC_MSG_CHECKING([what semaphore type to use]);
 AS_IF([test "x$have_mach" = "xtrue"],
   [AC_DEFINE(USE_MACH_SEM, 1, [Define to use Mach semaphores])
     AC_MSG_RESULT([Mach semaphores])],
+  [test "x$have_futex" = "xtrue"],
+  [AC_DEFINE(USE_FUTEX_SEM, 1, [Define to use Futex semaphores])
```
MadCoder: `USE_FUTEX` sounds better, Futexes are not semaphores, it's more like "[Define to use futex(7) when appropriate]"

src/semaphore.c
```diff
@@ -474,6 +507,11 @@ _dispatch_group_wake(dispatch_semaphore_t dsema)
 			int ret = sem_post(&dsema->dsema_sem);
 			DISPATCH_SEMAPHORE_VERIFY_RET(ret);
 		} while (--rval);
+#elif USE_FUTEX_SEM
+		do {
+			int ret = _dispatch_futex_signal(&dsema->dsema_futex);
```
MadCoder: this is FUTEX_WAIT(rval), do not loop, futexes support the proper interface already

src/semaphore.c
```diff
@@ -630,6 +680,11 @@ _dispatch_group_wait_slow(dispatch_semaphore_t dsema, dispatch_time_t timeout)
 			ret = sem_wait(&dsema->dsema_sem);
 		} while (ret == -1 && errno == EINTR);
 		DISPATCH_SEMAPHORE_VERIFY_RET(ret);
+#elif USE_FUTEX_SEM
+		do {
+			ret = _dispatch_futex_wait(&dsema->dsema_futex, NULL);
+		} while (ret == -1 && errno == EINTR);
```
MadCoder: you wrote that EINTR loop 3 or 4 times already, _dispatch_futex_wait should handle EINTR itself.

src/semaphore.c
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+_dispatch_futex_signal(dispatch_futex_t *dfx)
+{
+	uint64_t orig;
+	do {
+		orig = dfx->dfx_data;
+		if (slowpath((orig & DISPATCH_FUTEX_VALUE_MASK) ==
+					 DISPATCH_FUTEX_VALUE_MAX)) {
+			DISPATCH_CRASH("semaphore overflow");
+		}
+	} while (!dispatch_atomic_cmpxchg2o(dfx, dfx_data, orig, orig + 1, seq_cst));
```
MadCoder: release atomics are enough, no need for seq_cst.

src/semaphore.c
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+	if (slowpath(orig >> DISPATCH_FUTEX_NWAITERS_SHIFT)) {
+		int ret = _dispatch_futex_syscall(dfx, FUTEX_WAKE, 1, NULL);
+		DISPATCH_SEMAPHORE_VERIFY_RET(ret);
+	}
+	return 0;
+}
+
+// `timeout` is relative, and can be NULL for an infinite timeout--see futex(2).
+int
+_dispatch_futex_wait(dispatch_futex_t *dfx, const struct timespec *timeout)
```
MadCoder: this should handle retries as said before (EINTR)

src/semaphore.c
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+}
+
+// Atomic-decrement-if-positive on the semaphore value, with acquire MO if
+// successful.
+int
+_dispatch_futex_trywait(dispatch_futex_t *dfx)
+{
+	uint64_t orig;
+	while ((orig = dfx->dfx_data) & DISPATCH_FUTEX_VALUE_MASK) {
+		if (dispatch_atomic_cmpxchg2o(dfx, dfx_data, orig, orig - 1, seq_cst)) {
```
MadCoder: acquire is enough

src/semaphore.c
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+_dispatch_futex_wait_slow(dispatch_futex_t *dfx, const struct timespec *timeout)
+{
+	int spins = DISPATCH_FUTEX_NUM_SPINS;
+	// Spin for a short time (if there are no waiters).
+	while (spins-- && !dfx->dfx_data) {
+		dispatch_hardware_pause();
+	}
+	while (_dispatch_futex_trywait(dfx)) {
+		dispatch_atomic_add2o(dfx, dfx_data,
+							  1ull << DISPATCH_FUTEX_NWAITERS_SHIFT, seq_cst);
```
MadCoder: relaxed atomics is enough

src/semaphore.c
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+		int ret = _dispatch_futex_syscall(dfx, FUTEX_WAIT, 0, timeout);
+		dispatch_atomic_sub2o(dfx, dfx_data,
```
MadCoder: acquire barrier is enough

src/semaphore.c
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+		}
+	}
+	return 0;
+}
+
+static int
+_dispatch_futex_syscall(
+		dispatch_futex_t *dfx, int op, int val, const struct timespec *timeout)
+{
+#ifdef DISPATCH_LITTLE_ENDIAN
```
MadCoder: I hate this, dfx_data should be a union instead with a global uint64_t and a struct to access the futex address.

src/semaphore_internal.h
```diff
@@ -29,6 +29,13 @@
 
 struct dispatch_queue_s;
 
+#if USE_FUTEX_SEM
+typedef struct {
+	/* Low 32 bits are the count; high bits are the waiter count. */
+	volatile uint64_t dfx_data;
+} dispatch_futex_t;
+#endif  /* USE_FUTEX_SEM */
```
MadCoder: `_t` types in dispatch are pointers, please make this:

```C
typedef union {
    uint64_t volatile dfx_data;
    struct {
#if LITTLE_ENDIAN
        uint32_t dfx_futex;
        uint32_t dfx_waiters;
#else
        ...
#endif
    };
} *dispatch_futex_t
```

src/semaphore_internal.h
```diff
@@ -90,4 +99,11 @@ _dispatch_put_thread_semaphore(_dispatch_thread_semaphore_t sema)
 	}
 }
 
+#if USE_FUTEX_SEM
+int _dispatch_futex_init(dispatch_futex_t *dfx);
```
MadCoder: This should be

```C
#define DISPATCH_FUTEX_INIT  ((dispatch_futex_t){ 0 })
```

and no need for _dispatch_futex_dispose() I don't think it helps readability

src/semaphore.c
```diff
@@ -716,6 +771,11 @@ _dispatch_thread_semaphore_create(void)
 	int ret = sem_init(s4, 0, 0);
 	DISPATCH_SEMAPHORE_VERIFY_RET(ret);
 	return (_dispatch_thread_semaphore_t)s4;
+#elif USE_FUTEX_SEM
+	dispatch_futex_t *futex = malloc(sizeof(*futex));
```
MadCoder: you don't need that once the TSD patch is in, there should be an inlined `struct dispatch_futex_s` in that struct that is used as the "thread semaphore" to avoid a malloc for 64bits.

src/semaphore.c
MadCoder: actually now that I read the full patch I understand why, but it feels sad that FUTEX_WAKE with `n` can't be use, I assume it's not the end of the world though.

src/semaphore.c
MadCoder: actually unlike other stuff that needs to be in the same struct for performance and also to be able to run a destructor, we don't care for a futex on linux.

given that `__thread` usage is fine on linux, just have a global:

static __thread struct _dispatch_futex_s _dispatch_thread_semaphore;

and avoid this malloc altogether.

src/semaphore.c
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
 #error "No supported semaphore type"
 #endif
 }
+
+#if USE_FUTEX_SEM
+#define DISPATCH_FUTEX_NUM_SPINS 100
+#define DISPATCH_FUTEX_VALUE_MAX INT_MAX
+#define DISPATCH_FUTEX_NWAITERS_SHIFT 32
+#define DISPATCH_FUTEX_VALUE_MASK ((1ull << DISPATCH_FUTEX_NWAITERS_SHIFT) - 1)
```
MadCoder: also these last 3 should be next to the structure definition

src/semaphore.c
MadCoder: also since you loop should be the weak variant and reload as a side effect of the cmpxchg:

```C
	uint64_t orig = dfx->dfx_data;
	do {
		if (slowpath((orig & DISPATCH_FUTEX_VALUE_MASK) ==
					 DISPATCH_FUTEX_VALUE_MAX)) {
			DISPATCH_CRASH("semaphore overflow");
		}
	} while (!dispatch_atomic_cmpxchgvw2o(dfx, dfx_data, orig, orig + 1, &orig, seq_cst));
```

src/semaphore.c
MadCoder: Same as above, should be a cmpxcgh weak loop instead:

```C
uint64_t orig = dfx->dfx_data;
while (orig & DISPATCH_FUTEX_VALUE_MASK) {
    if (dispatch_atomic_cmpxchgvw2o(dfx, dfx_data, orig, orig - 1, &orig, acquire)) {
        return 0;
    }
}
```

configure.ac
MadCoder: What I mean is that I can tell you'll want to use FUTEX for other things, so I wouldn't make this scoped to "semaphores" only.

src/internal.h
```diff
@@ -230,7 +230,11 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
 #include <search.h>
 #if USE_POSIX_SEM
 #include <semaphore.h>
+#elif USE_FUTEX_SEM
```
MadCoder: since USE_FUTEX will probably be used for more than semaphores, I would test for USE_FUTEX first, as it's "more" generic.

src/internal.h
```diff
@@ -242,6 +246,34 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
...
+#include <endian.h>
+#if __BYTE_ORDER == __LITTLE_ENDIAN
+#define DISPATCH_LITTLE_ENDIAN
+#elif __BYTE_ORDER == __BIG_ENDIAN
+#define DISPATCH_BIG_ENDIAN
+#endif
+
+#endif  /* defined(__APPLE__) */
+
+#if !(defined(DISPATCH_BIG_ENDIAN) || defined(DISPATCH_LITTLE_ENDIAN))
```
MadCoder: rather make it an `#else` of all the blocks that define it for various platforms.
also I would have defines for the various usual CPUs as in

```C
#elif defined(__i386__) || defined(__x86_64__) || defined(__i386) || ..
#define DISPATCH_LITTLE_ENDIAN
#else
#  error please define DISPATCH_LITTLE_ENDIAN or DISPATCH_BIG_ENDIAN for your architecture / platform
#endif

File: src/semaphore.c [hunk 21]
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+int
+_dispatch_futex_dispose(dispatch_futex_t *dfx)
+{
+	(void)dfx;
+	return 0;
+}
+
+// Increments semaphore value and reads wait count in a single atomic
+// operation, with release MO. If wait count is nonzero, issues a FUTEX_WAKE.
+int
```
MadCoder: should be void, it never returns failure, does it?

src/semaphore.c
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+{
+	if (fastpath(!_dispatch_futex_trywait(dfx))) {
+		return 0;
+	}
+	return _dispatch_futex_wait_slow(dfx, timeout);
+}
+
+// Atomic-decrement-if-positive on the semaphore value, with acquire MO if
+// successful.
+int
```
MadCoder: dispatch usually prefers `bool` for success/failure rather than the old POSIX -1/0 terminology.

please make it a bool where `true` means success

MadCoder: dispatch usually uses `bool` to mean success, and so please do so and have `true` indicate success

src/semaphore.c
```diff
@@ -804,3 +880,130 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+DISPATCH_NOINLINE
+static int
+_dispatch_futex_wait_slow(dispatch_futex_t *dfx, const struct timespec *timeout)
+{
+	int spins = DISPATCH_FUTEX_NUM_SPINS;
+	// Spin for a short time (if there are no waiters).
+	while (spins-- && !dfx->dfx_data) {
+		dispatch_hardware_pause();
+	}
+	while (_dispatch_futex_trywait(dfx)) {
```
MadCoder: especially here this is very confusing, I read this as "while _futex_trywait works" except that it's the exact opposite. so making it a `bool` result and inverting the condition here is way better

MadCoder:
ok I know this is a lot of comments, but despite that I love where this is going.

I have not completely checked that the futex() based semaphore implementation is correct, and only made remarks for the barriers (which aren't important on intel but could be for other architectures).

Note that dispatch assumes little endian in a LOT of places so I'm not entirely sure it's useful to test, but why not.

Last, as far as atomic barriers go, for a semaphore, the semantics is that:
- signal() must have a release barrier
- wait() must have at least one acquire barrier as late as possible.

This is why there is no need for a barrier at all right before the call to futex() in the wait_slow because the sub2o done on the way out has one and that is enough.

src/semaphore.c
MadCoder: also if you use the union, you would not have to mask anymore, because you'd load into dfx->dfx_data but could read the member which would be significantly more readable

configure.ac
frankeh: In this particular case we want to be symmetric on the USE_xxxxx_SEM, so that should stay.
Will see whether we can be more generic here from which we derive USE_FUTEX_SEM.

src/semaphore.c
frankeh: If you follow the code down to _dispatch_futex_wait_slow
it indicates the resason that <timeout> needs to be recomputed.
If you want I can massage that whole code, but that would require the timeout recomputation
to be brought in here, which seems overdone, instead elevate it back out the caller.
Let me know

src/semaphore.c
frankeh: Unfortunately the atomic.h has basically no explanation. 
dispatch_atomic_cmpxchgvw2o  ::= dispatch_atomic_cmpxchgvw ?

src/internal.h
frankeh: confusing, proper way to do that is let the compiler do it.
I don't know whether APPLE does it that way but properway is:
#if defined(__APPLE__)
#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
#define DISPATCH_LITTLE_ENDIAN
#elif __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
#define DISPATCH_BIG_ENDIAN
#endif
#else 
#include <endian.h>
#if __BYTE_ORDER == __LITTLE_ENDIAN
#define DISPATCH_LITTLE_ENDIAN
#elif __BYTE_ORDER == __BIG_ENDIAN
#define DISPATCH_BIG_ENDIAN
#else
#error "please define DISPATCH_LITTLE_ENDIAN or DISPATCH_BIG_ENDIAN for your
architecture / platform"
#endif  /* defined(__APPLE__) */

configure.ac
MadCoder: fair enough

src/internal.h
MadCoder: I think that libbsd-dev has the definition too btw.

src/semaphore.c
MadCoder: ok fair enough

src/semaphore.c
MadCoder: dispatch_atomic_cmpxchgvw2o takes (object, field, expected, new_value, pointer_to_current_value_on_failure, barrier).

v == return current value on failure
w == weak (as in compare_exchange_weak in c11 parlance).

so when you loop you should use the vw variant. just use the code I gave you it works, in C11 parlance:

```C
dispatch_atomic_cmpxchgvw2o(p, f, orig, orig + 1, &orig, release)
```

is more or less equivalent to

```C
atomic_compare_exchange_weak_explicity(&dfx->dfx_data, &orig, orig + 1, memory_order_release, memory_order_relaxed);
```

MadCoder: yeah the definition is right here:

```C
#define dispatch_atomic_cmpxchgvw(p, e, v, g, m) \
		({ _dispatch_atomic_basetypeof(p) _v = (v), _r = (e); _Bool _b = \
		__c11_atomic_compare_exchange_weak(_dispatch_atomic_c11_atomic(p), \
		&_r, _v, dispatch_atomic_memory_order_##m, \
		dispatch_atomic_memory_order_relaxed); *(g) = (typeof(*(p)))_r;  _b; })
```

the 2o variants are for "reasons" but just about `&p->field` where `p` and `field` are the two first arguments.

that thing above looks weird but does something akin to this (pseudo code):

```C
#define dispatch_atomic_cmpxchgvw(p, e, v, g, m) \
		({ *g = e;  \
		atomic_compare_exchange_weak_explicit(p, g, v, memory_order_##m, memory_order_relaxed); })
```

the "v" variant does the same for strong.
and the non "v" variant is a strong cmpxchg that doesn't get you the current value.

rule of thumb, if you loop, you usually want a `v` variant as it will reload the value for you efficiently, and usually you want the `w` (weak) variant as going to strong on LL/SC architectures such as arm or ppc will itself loop, so if you do, you should piggy back on your own loop instead.

src/semaphore.c
frankeh: this will create a lot of #if-then-else code

Look at ..
DISPATCH_ALWAYS_INLINE
static inline _dispatch_thread_semaphore_t
_dispatch_get_thread_semaphore(void)
{
    _dispatch_thread_semaphore_t sema = (_dispatch_thread_semaphore_t)
            _dispatch_thread_getspecific(dispatch_sema4_key);
    if (slowpath(!sema)) {
        return _dispatch_thread_semaphore_create();
    }
    _dispatch_thread_setspecific(dispatch_sema4_key, NULL);
    return sema;
}

that would have to be ifdef.
Only other thing I can think off is to use as you say above and return the address
of the thread local variable in the allocation ? Is that's what you meant?

you prefer there to separate out the handling of futex?
I went for cleanness here.

MadCoder: > Only other thing I can think off is to use as you say above and return the address
of the thread local variable in the allocation ? Is that's what you meant?

yes

MadCoder: actually I looked at the dispatch source and using a per thread static one doesn't work because the memory has to be reusable because of dispatch_apply and similar codepaths, so you need the malloc as sad as this is. though instead of doing malloc I would use a continuation as it will very often be taken from a very hot pool.

IOW just replace the malloc with a _dispatch_continuation_alloc(), use it (it's way larger than what you need) and free it with _dispatch_continuation_free().

frankeh: Hmmm, just got it to run last night with tests passing, though that doesn't mean much.

So three questions to 
a) is the storage for the futex required beyond the life time of a thread?
b) are there multiple instances over the life of the thread (though one at a time)?
c) will any other thread point at it ?
Otherwise, why can't we use thread local?

src/semaphore_internal.h
frankeh: I tried to code this up ... this makes the code <ugly> IMO
and also not clear from a semantics point of view, since the code needs to massage the full uint64.
For example:  we compare and exchange of the full value, which believe we have to or NOT?
since data has to be volatile (even for the dfx_futex etc), its not obvious that the compiler won't reload 
it though the same union is accessed.

I suggest not to go this route at this time and leave as is. The code is limited to a few lines.
We can discuss later, but want to get this out and done.

DISPATCH_INLINE
bool
_dispatch_futex_trywait(dispatch_futex_t dfx) 
{
    uint64_t orig;
    while ((orig = dfx->dfx_data) && dfx->dfx_futex) {
        if (dispatch_atomic_cmpxchg2o(dfx, dfx_data, orig, orig - 1, seq_cst)) {
            return true;
        }
    }    
    return false;
}

src/semaphore.c
frankeh: so question is whether the 64-bit data structure needs to be manipulated as one unit.
So we can't afford from my understanding to just to manipulate the dfx_futex by itself since somebody else
can modify the dfx_waiters. Very unlikely due to cache coherency, but seems unnessary effort for basically no gain.

src/semaphore.c
MadCoder: it can have nested uses beacuse of dispatch_apply, that's why there is this `_semaphore_put()` thing to take it and leave it back in the TSD.

though you could use stack allocated stuff each time and I *think* it could work, and then nesting just works. it doesn't even really need to be in the TSD.

@das would have to confirm.

src/semaphore.c
MadCoder: yes it needs to be manipulated as a 64bit value. but SYS_futex() only works on uint32_t's that's the guy who wants to see the single word.

src/semaphore_internal.h
MadCoder: that's not what it should be, you should still use the 64bit field everywhere, only "futex" needs to use `&dfx->dfx_futex` for its argument. you wouldnt use the dfx_waiters part of the struct ever, it's mostly for documentation purposes of how the code works, it's important for audit purposes.

src/semaphore.c
frankeh: OK .. let me convert that back then. Its a 5 line change.

src/semaphore_internal.h
frankeh: The problem is that I have to reload it .. Let me massage a bit again, will see how it looks like.
Wanna move on to other problems ..

MadCoder: reload what?? of course you don't have to, you can use the value through the unt64_t, and pass the &dfx->dfx_futex to the futex call alright. it's just an addres, there's nothing to read from it, only the kernel does.

MadCoder: reloading would even be wrong as it could be another value than what you loaded from dfx_data.

frankeh: That's my point .. OK .. I code it up, then we discuss based on the code.
Its easier that way and maybe we just misunderstand each other here.

src/semaphore.c
```diff
@@ -804,3 +889,110 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+DISPATCH_INLINE
+bool
+_dispatch_futex_trywait(dispatch_futex_t dfx)
+{
+	uint64_t orig;
+	while ((orig = dfx->dfx_data) & DISPATCH_FUTEX_VALUE_MASK) {
+		if (dispatch_atomic_cmpxchg2o(dfx, dfx_data, orig, orig - 1, acquire)) {
+			return true;
+		}
+	}
```
MadCoder: this code should be:

```C
	uint64_t orig = dispatch_atomic_load2o(dfx, dfx_data, relaxed);
 	while (orig & DISPATCH_FUTEX_VALUE_MASK) {
		if (dispatch_atomic_cmpxchgv2o(dfx, dfx_data, orig, orig - 1, &orig, acquire)) {
			return true;
		}
	}
	return false;
```

src/semaphore.c
```diff
@@ -804,3 +889,110 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+
+// Increments semaphore value and reads wait count in a single atomic
+// operation, with release MO. If wait count is nonzero, issues a FUTEX_WAKE.
+DISPATCH_INLINE
+bool
+_dispatch_futex_signal(dispatch_futex_t dfx)
+{
+	uint64_t orig;
+	do {
+		orig = dfx->dfx_data;
```
MadCoder: This must be

```C
	uint64_t orig = dispatch_atomic_load2o(dfx, dfx_data, relaxed);
	do {
```

src/semaphore.c
```diff
@@ -804,3 +889,110 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+		if (slowpath((orig & DISPATCH_FUTEX_VALUE_MASK) ==
+					 DISPATCH_FUTEX_VALUE_MAX)) {
+			DISPATCH_CRASH("semaphore overflow");
+		}
+	} while (!dispatch_atomic_cmpxchgvw2o(dfx, dfx_data, orig, orig + 1, &orig, release));
+	if (slowpath(orig >> DISPATCH_FUTEX_NWAITERS_SHIFT)) {
+		int ret = _dispatch_futex_syscall(&dfx->dfx_futex, FUTEX_WAKE, 1, NULL);
```
MadCoder: can't you get EINTR if a debugger is attached? in which case you have to loop

src/semaphore.c
```diff
@@ -804,3 +889,110 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+	// Spin for a short time (if there are no waiters).
+	while (spins-- && !dfx->dfx_data) {
+		dispatch_hardware_pause();
+	}
+	while (!_dispatch_futex_trywait(dfx)) {
+		dispatch_atomic_add2o(dfx, dfx_data,
+							  1ull << DISPATCH_FUTEX_NWAITERS_SHIFT, relaxed);
+		int ret = _dispatch_futex_syscall(&dfx->dfx_futex, FUTEX_WAIT, 0, timeout);
+		dispatch_atomic_sub2o(dfx, dfx_data,
+							  1ull << DISPATCH_FUTEX_NWAITERS_SHIFT, acquire);
```
MadCoder: the indent rule is two tabs so that should be:

```C
		dispatch_atomic_sub2o(dfx, dfx_data,
				1ull << DISPATCH_FUTEX_NWAITERS_SHIFT, acquire);
```

and same above

src/semaphore.c
```diff
@@ -804,3 +889,110 @@ _dispatch_thread_semaphore_wait(_dispatch_thread_semaphore_t sema)
...
+	return _dispatch_futex_wait_slow(dfx, timeout);
+}
+
+DISPATCH_NOINLINE
+static bool
+_dispatch_futex_wait_slow(dispatch_futex_t dfx, const struct timespec *timeout)
+{
+	int spins = DISPATCH_FUTEX_NUM_SPINS;
+	// Spin for a short time (if there are no waiters).
+	while (spins-- && !dfx->dfx_data) {
```
MadCoder: should be

```C
while (spins-- && !dispatch_atomic_load(dfx, dfx_data, relaxed)) {
```

src/semaphore.c
MadCoder: please still move them next to the definition of dispatch_futex_t

MadCoder:
sorry I had missed the extra commit, wasn't notified, I start to believe either some notification go to my spam box or github has a problem.

anyway made extra comments but that is more like what I had in mind indeed!

src/semaphore.c
```diff
@@ -716,6 +785,10 @@ _dispatch_thread_semaphore_create(void)
 	int ret = sem_init(s4, 0, 0);
 	DISPATCH_SEMAPHORE_VERIFY_RET(ret);
 	return (_dispatch_thread_semaphore_t)s4;
+#elif USE_FUTEX_SEM
```
MadCoder: not reall related to that very code but:

`_dispatch_get_thread_semaphore` and `_dispatch_put_thread_semaphore` are meant to amortize the cost of the syscall to create a kernel semaphore.

on linux where a futex is used, caching the malloc() is not useful, as we allocate a continuation which has its own cache.

`_dispatch_put_thread_semaphore` and `_dispatch_get_thread_semaphore` should be the same as `dispose` and `create` when USE_FUTEX_SEM is in use like this:

```C
#if USE_FUTEX_SEM
#define _dispatch_get_thread_semaphore() _dispatch_thread_semaphore_create()
#define _dispatch_put_thread_semaphore() _dispatch_thread_semaphore_dispose()
#else
DISPATCH_ALWAYS_INLINE
static inline _dispatch_thread_semaphore_t
_dispatch_get_thread_semaphore(void)
{
	_dispatch_thread_semaphore_t sema = (_dispatch_thread_semaphore_t)
			_dispatch_thread_getspecific(dispatch_sema4_key);
	if (slowpath(!sema)) {
		return _dispatch_thread_semaphore_create();
	}
	_dispatch_thread_setspecific(dispatch_sema4_key, NULL);
	return sema;
}

DISPATCH_ALWAYS_INLINE
static inline void
_dispatch_put_thread_semaphore(_dispatch_thread_semaphore_t sema)
{
	_dispatch_thread_semaphore_t old_sema = (_dispatch_thread_semaphore_t)
			_dispatch_thread_getspecific(dispatch_sema4_key);
	_dispatch_thread_setspecific(dispatch_sema4_key, (void*)sema);
	if (slowpath(old_sema)) {
		return _dispatch_thread_semaphore_dispose(old_sema);
	}
}
#endif
```

src/semaphore.c
frankeh: Loop only around for line 945 or do we have to regrab the whole from line 936 (including reloading the orig value ? Not clear to me.

MadCoder: ok I'm wrong the man page says:

```
       EINTR  A FUTEX_WAIT operation was interrupted by  a  signal  (see  sig‐
              nal(7)) or a spurious wakeup.
```

so that is fine

src/semaphore.c
frankeh: dispatch_atomic_load  or dispatch_atomic_load2o ?

MadCoder: 2o sorry.

frankeh:
OK .. in case the commit got lost ..

MadCoder:
oooh that's it, I don't get commit notifications. doh...

I'm ok with this, can you please squash this in a single commit & force push and then I'll merge

frankeh:
done .. can you see it ?

MadCoder:
nope I still see 3 commits. you have to update the pull request one way or the other, I assumed that a force push on your repository should do it naturally but maybe not.

frankeh:
Damned ..

2023  git checkout master
2024  git merge --squash futexes
2025  git commit -m "complete futex patch for semaphore implementation"
2033  git push -f origin appledispatch/master 

what's wrong with this ?

MadCoder:
`2033 git push -f origin appledispatch/master` shouldn't that be "futexes" ? your pull request is about "futexes"

frankeh:
Hmm none of that works .. all tell me its upto date.
Will have to consult somebody else.

PR #54, opened 2016-02-24
Enable notification to libpwq before semaphore waiting

frankeh:
Before waiting on a dispatch_semaphore (slow path), notify the thread pool manager
that this thread will go to block.

Depends on the small patch submitted to libpwq
https://github.com/mheily/libpwq/pull/10   (not approved yet).
This patch will wakeup the thread manager (instead of doing this once per second)
which then forces a new worker_thread to be started.

Lack of this patch will result in reduced concurrency at program startup time.
See https://bugs.swift.org/browse/SR-761  for some symptoms.

This patch is not a fix for all, for instance it does not catch sleep() calls, but according
to Swift Foundation implementers its a good intermediate solution before linux kernel solutions
are considered which have significantly longer lead times.


frankeh:
Will provide a clean submission.

PR #55, opened 2016-02-26
resolve all compiler warnings

dgrove-oss:
Resolve all compiler warnings when building libdispatch and
the test suite on linux.  It now compiles cleanly.

Many of the warnings were due to platform-specific sized integers
being used in printf-like contexts.  In libdispatch itself, I took
the approach of adding casts to the values to match the size/type
specified in the printf format string (under the assumption that
the expected output format should not be changed and needs to be
the same on all platforms).

### DIFF
--- a/src/source.c
+++ b/src/source.c
@@ -148,7 +148,7 @@ dispatch_source_create(dispatch_source_type_t type,
 	dk->dk_kevent.ident = handle;
 	dk->dk_kevent.flags |= EV_ADD|EV_ENABLE;
 	dk->dk_kevent.fflags |= (uint32_t)mask;
-	dk->dk_kevent.udata = (uintptr_t)dk;
+	dk->dk_kevent.udata = (typeof(dk->dk_kevent.udata))dk;
 	TAILQ_INIT(&dk->dk_sources);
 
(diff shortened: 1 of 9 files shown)


src/source.c
```diff
@@ -1532,8 +1532,13 @@ struct dispatch_timer_s _dispatch_timer[] =  {
 #define DISPATCH_TIMER_COUNT \
 		((sizeof(_dispatch_timer) / sizeof(_dispatch_timer[0])))
 
+#if __linux__
```
MadCoder: I don't understand that guard, why isn't uintptr_t working on linux, why is it needing a pointer type?

dgrove-oss: On Linux we only have kevent, not kevent64 (os/linux_base.h #defines kevent64 to be kevent).  The type of the udata field of kevent is void* (vs. kevent64 where the type of the udata field is uintptr_t).     It would be nicer to use a typeof here to avoid a #if __linux__ , but I didn't see a good way to do that through all the levels of macros being defined.

MadCoder: ugh fair enough. libkqueue should really be extended :/

PR #58, opened 2016-02-29
inform libpwq thread manager before semaphore wait

frankeh:
Before waiting on a dispatch_semaphore (slow path), notify the thread pool manager
that this thread will go to block.

Depends on the small patch submitted to libpwq which has been integrated. 

This patch will wakeup the thread manager (instead of doing this once per second)
which then forces a new worker_thread to be started.

Lack of this patch will result in reduced concurrency at program startup time.
See https://bugs.swift.org/browse/SR-761 for some symptoms.

This patch is not a fix for all, for instance it does not catch sleep() calls, but according
to Swift Foundation implementers its a good intermediate solution before the  linux kernel solution
is considered which has a longer lead time and requires a kernel extension to be installed.

### DIFF
--- a/src/semaphore.c
+++ b/src/semaphore.c
@@ -383,6 +383,7 @@ again:
 			uint64_t nsec = _dispatch_timeout(timeout);
 			_timeout.tv_sec = (typeof(_timeout.tv_sec))(nsec / NSEC_PER_SEC);
 			_timeout.tv_nsec = (typeof(_timeout.tv_nsec))(nsec % NSEC_PER_SEC);
+			pthread_workqueue_signal_np();
 			ret = slowpath(_dispatch_futex_wait(&dsema->dsema_futex, &_timeout));
 		} while (ret == false && errno == EINTR);
 
@@ -430,6 +431,7 @@ again:
 		DISPATCH_SEMAPHORE_VERIFY_RET(ret);
 #elif USE_FUTEX_SEM
 		do {
+			pthread_workqueue_signal_np();
 			ret = _dispatch_futex_wait(&dsema->dsema_futex, NULL);
 		} while (ret == false && errno == EINTR);
 		DISPATCH_SEMAPHORE_VERIFY_RET(ret);
@@ -648,6 +650,7 @@ again:
 			uint64_t nsec = _dispatch_timeout(timeout);
 			_timeout.tv_sec = (typeof(_timeout.tv_sec))(nsec / NSEC_PER_SEC);
 			_timeout.tv_nsec = (typeof(_timeout.tv_nsec))(nsec % NSEC_PER_SEC);
+			pthread_workqueue_signal_np();
 			ret = slowpath(_dispatch_futex_wait(&dsema->dsema_futex, &_timeout));
 		} while (ret == false && errno == EINTR);
 
@@ -695,6 +698,7 @@ again:
 		DISPATCH_SEMAPHORE_VERIFY_RET(ret);
 #elif USE_FUTEX_SEM
 		do {
+			pthread_workqueue_signal_np();
 			ret = _dispatch_futex_wait(&dsema->dsema_futex, NULL);
 		} while (ret == false && errno == EINTR);
 		DISPATCH_SEMAPHORE_VERIFY_RET(ret);


dgrove-oss:
Hubertus, I think we also need to update the libpwq submodule version so the new function that was merged in upstream will be there.  If you are lucky, doing " git submodule update --remote" in the libdisaptach directory will be enough (then commit/push the changed submodule version for libpwq).   If not, let's do it together tomorrow.

frankeh:
Lost me here. I thought when I check out libdispatch its automatic ?
I hold off with experimenting on it until you are in.

frankeh:
@madcoder:  fixed the libpwq level. This is ready for merge to fix the thread starvation problem reported.

drewcrawford:
LGTM  :+1:   I'm running this patch in development.

MadCoder:
This change breaks OS X, as HAVE_PTHREAD_WORKQUEUES is defined to 1 for us
also this is missing all the other cases that would cause a thread to block

which makes two reasons not to take that patch atm.

drewcrawford:
Perhaps it would be appropriate to add a guard for Linux. 

It seems a strange objection that a patch does not solve *enough* bugs, but I suppose there's no harm in extending it to the other cases.

frankeh:
@madcoder : fully agree that is does not solve the other blocking problems as highlighted in the discussion. 
The questions is whether we can provide a temp solution that can easily be dropped later to have foundation folks to be able to make some progress.
A kernel extension is the right solution if I own the system and have root authority to install, however
I suspect there are a plenty of people who want to give it a shot and if the extension is not available or said person doesn't have authority to install it, would this be at least a solution?
As you know I am looking at the kernel extension.

drewcrawford:
To speak quickly about the kernel extension, Ubuntu 14.04 is supported through 2019, RHEL 7 is supported through 2024.  So kernels without dispatch-specific patches are going to be a fact of life.

I agree that a kernel-based approach is optimal but in the meantime we need a "plan B", because presumably someone  ( :hand: ) will want to run libdispatch on a standard Linux system sometime in the next 5 years.

frankeh:
Drew, I am in touch with Pierre. Its not as bad as I initially thought.
The basic requirement (namely a process change handler) exists in Linux since 2.6.30 or so.
So it can entirely be done as a module, which is a huge advantage.
Will code something up.

drewcrawford:
A docker container [can't load a kernel module](http://askubuntu.com/questions/314817/how-do-i-install-a-kernel-module-in-an-lxc-guest-machine), so we're going to break lots of things with this as our only approach.  In the typical case, no Swift program will be able to be Dockerized, will not be able to run on [ECS](https://aws.amazon.com/ecs/), [CE](https://cloud.google.com/container-engine/), [Tutum](https://www.tutum.co) and others.  This will also break the common Linux-on-OSX development environments of [Vagrant](https://www.vagrantup.com) and [boot2docker](http://boot2docker.io).

While I can figure out how to load kernel modules onto my hosts, my customers can't, so I will probably be required to create/maintain/build/distribute my own userspace solution here.

frankeh:
My hope here was to get a duality in.
If the module (e.g. for hosted container clouds) is loaded, will try to open the device and succeed.
If not we need some user level fall back which this patch at least can provide. We can figure this
out at runtime.

MadCoder:
So, that discussion misses the point, as I said, my first concern is that this change breaks OS X. this is a big no-go.

Also you're missing another point which is that we (Apple engineers) do have 10 years of experience with use of dispatch, and we've seen what it does to rely on very quick thread creation, you get an entire class of other problems such as thread explosion and pool reduction issues. The fact that it "fixes bugs" is not incentive enough if it fixes it the wrong way. I will discuss that change with @das today, but I'm worried about that change.

MadCoder:
btw I think the right way to write this patch is to have a HAVE_PTHREAD_WORKQUEUE_SIGNAL_NP detection from autoconf and to #define that guy to if not, and use that name directly. *if* we take the patch. (again, it's not about being the full solution or not, I have further concerns with the direction of the change in the first place).

src/semaphore.c
```diff
@@ -301,6 +301,14 @@ dispatch_semaphore_signal(dispatch_semaphore_t dsema)
 	return _dispatch_semaphore_signal_slow(dsema);
 }
 
+#if defined(__linux__) && HAVE_PTHREAD_WORKQUEUES && \
```
MadCoder: please remove that whole hunk

src/semaphore.c
```diff
@@ -371,6 +379,7 @@ _dispatch_semaphore_wait_slow(dispatch_semaphore_t dsema,
 			uint64_t nsec = _dispatch_time_to_nanoseconds(timeout);
 			_timeout.tv_sec = (typeof(_timeout.tv_sec))(nsec / NSEC_PER_SEC);
 			_timeout.tv_nsec = (typeof(_timeout.tv_nsec))(nsec % NSEC_PER_SEC);
+			_dispatch_threadmgr_inform_wait();
```
MadCoder: remove

src/semaphore.c
```diff
@@ -383,6 +392,7 @@ _dispatch_semaphore_wait_slow(dispatch_semaphore_t dsema,
 			uint64_t nsec = _dispatch_timeout(timeout);
 			_timeout.tv_sec = (typeof(_timeout.tv_sec))(nsec / NSEC_PER_SEC);
 			_timeout.tv_nsec = (typeof(_timeout.tv_nsec))(nsec % NSEC_PER_SEC);
+			_dispatch_threadmgr_inform_wait();
```
MadCoder: keep

src/semaphore.c
```diff
@@ -425,11 +435,13 @@ _dispatch_semaphore_wait_slow(dispatch_semaphore_t dsema,
 		DISPATCH_SEMAPHORE_VERIFY_KR(kr);
 #elif USE_POSIX_SEM
 		do {
+			_dispatch_threadmgr_inform_wait();
```
MadCoder: remove

src/semaphore.c
```diff
@@ -425,11 +435,13 @@ _dispatch_semaphore_wait_slow(dispatch_semaphore_t dsema,
...
 			ret = sem_wait(&dsema->dsema_sem);
 		} while (ret != 0);
 		DISPATCH_SEMAPHORE_VERIFY_RET(ret);
 #elif USE_FUTEX_SEM
 		do {
+			_dispatch_threadmgr_inform_wait();
```
MadCoder: call pthread_workqueue_signal_np() directly

MadCoder:
so @das and I both think that this direction of hinting the libpwq code from userland is unsustainable long term for a lot of reasons that we can discuss offline from this review.

However, it is clear that it is a needed work around for the bringup so please only do these calls in the FUTEX call that is specific to linux to begin with, and do the raw calls to `pthread_workqueue_signal_np()` in that case. do not use it elsewhere, for the sake of the bring up.

frankeh:
comments are executed.
We can take it out once we have a proper kernel support.

PR #62, opened 2016-03-25
Build fix for Linux overlay with 03242016 compiler.

dgrove-oss:
With the 03242016 development drop, building the
Dispatch overlay on Linux fails because mode_t and
off_t are not defined.  This is a minimal fix to
include the needed header files and get the build
working again.

### DIFF
--- a/dispatch/dispatch.h
+++ b/dispatch/dispatch.h
@@ -26,6 +26,7 @@
 #include <TargetConditionals.h>
 #endif
 #include <sys/cdefs.h>
+#include <sys/types.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <stdbool.h>
--- a/dispatch/io.h
+++ b/dispatch/io.h
@@ -26,6 +26,10 @@
 #include <dispatch/base.h> // for HeaderDoc
 #endif
 
+#ifdef __DISPATCH_BUILDING_SWIFT_MODULE__
+#include <stdio.h>
+#endif
+
 __BEGIN_DECLS
 
 /*! @header
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -122,7 +122,7 @@ SWIFT_OBJECTS=	\
 	$(abs_builddir)/Dispatch.swiftdoc \
 	$(abs_builddir)/Dispatch.o
 
-SWIFTC_FLAGS = -Xcc -fmodule-map-file=$(abs_top_srcdir)/dispatch/module.map -I$(abs_top_srcdir) -parse-as-library -Xcc -fblocks
+SWIFTC_FLAGS = -Xcc -D__DISPATCH_BUILDING_SWIFT_MODULE__=1 -Xcc -fmodule-map-file=$(abs_top_srcdir)/dispatch/module.map -I$(abs_top_srcdir) -parse-as-library -Xcc -fblocks
 
 $(abs_builddir)/Dispatch.o: $(abs_srcdir)/swift/Dispatch.swift
 	$(SWIFTC) $(SWIFTC_FLAGS) -c -o $@ $<


dgrove-oss:
This fix is also included in #61, but since it may take some time for 61 to get merged and the Linux build of dispatch is broken without this small change opening a separate PR.

dgrove-oss:
Merged this changeset to experimental/foundation in #68.  Leaving this pull request open and baking branch undeleted  until a fix gets merged to the master branch to enable it to compile with 03242016 Swift tool chain.

dispatch/dispatch.h
```diff
@@ -32,6 +32,10 @@
 #include <stdarg.h>
 #include <unistd.h>
 #include <fcntl.h>
+#ifndef __APPLE__
+#include <stdio.h>
```
MadCoder: why do you need that in a public header? we kind of try to restrict what we export to a bare minimum, <sys/types.h> I understand, but stdio seems to point toward something else being wrong.

I would also add <sys/types.h> for everyone not only !__APPLE__ it's fine

dgrove-oss:
Coming back to this, I'm actually not sure if this is really a bug that should be fixed by changing dispatch.h or by fixing the clang importer of swiftc or the module maps the importer is using for SwiftGLibc.  This change is not for the C compilaition of libdispatch.  What it is trying to fix is a problem when compiling Dispatch.swift with a swift toolchain post the 03242016 drop. Namely:

```
libtool: compile:  clang -DHAVE_CONFIG_H -I. -I/home/dgrove/swift/swift-corelibs-libdispatch/src -I../config -I.. -I/home/dgrove/swift/swift-corelibs-libdispatch -I/home/dgrove/swift/swift-corelibs-libdispatch/private -I/home/dgrove/swift/swift-corelibs-libdispatch/os -I/home/dgrove/swift/swift-corelibs-libdispatch/libpwq/include -Wall -fvisibility=hidden -momit-leaf-frame-pointer -I/usr/include/kqueue -isystem /usr/include/bsd -DLIBBSD_OVERLAY -fblocks -g -O2 -c /home/dgrove/swift/swift-corelibs-libdispatch/src/swift/swift_wrappers.c  -fPIC -DPIC -o swift/.libs/swift_wrappers.o
/home/dgrove/swift/build/dpg/swift-linux-x86_64/bin/swiftc -Xcc -fmodule-map-file=/home/dgrove/swift/swift-corelibs-libdispatch/dispatch/module.map -I/home/dgrove/swift/swift-corelibs-libdispatch -parse-as-library -Xcc -fblocks -c -o /home/dgrove/swift/build/dpg/libdispatch-linux-x86_64/src/Dispatch.o /home/dgrove/swift/swift-corelibs-libdispatch/src/swift/Dispatch.swift
<module-includes>:1:10: note: in file included from <module-includes>:1:
#include "dispatch.h"
         ^
/home/dgrove/swift/swift-corelibs-libdispatch/dispatch/dispatch.h:59:10: note: in file included from /home/dgrove/swift/swift-corelibs-libdispatch/dispatch/dispatch.h:59:
#include <dispatch/io.h>
         ^
/home/dgrove/swift/swift-corelibs-libdispatch/dispatch/io.h:253:31: error: declaration of 'mode_t' must be imported from module 'SwiftGlibc.POSIX.sys.types' before it is required
        const char *path, int oflag, mode_t mode,
                                     ^
/usr/include/x86_64-linux-gnu/sys/types.h:70:18: note: previous declaration is here
typedef __mode_t mode_t;
                 ^
<module-includes>:1:10: note: in file included from <module-includes>:1:
#include "dispatch.h"
         ^
/home/dgrove/swift/swift-corelibs-libdispatch/dispatch/dispatch.h:59:10: note: in file included from /home/dgrove/swift/swift-corelibs-libdispatch/dispatch/dispatch.h:59:
#include <dispatch/io.h>
         ^
/home/dgrove/swift/swift-corelibs-libdispatch/dispatch/io.h:355:2: error: declaration of 'off_t' must be imported from module 'SwiftGlibc.C.stdio' before it is required
        off_t offset,
        ^
/usr/include/stdio.h:90:17: note: previous declaration is here
typedef __off_t off_t;
```
Do you want to pull in someone from the compiler team to advise?

MadCoder:
off_t is supposed to be in <unistd.h> which is a more reasonnable header to import from dispatch.h IMO (for all platforms)

dgrove-oss:
Looks like the key bit (which other header file to try) got chopped from your comment.

MadCoder:
unistd.h stupid html sanitizer *sigh*

dgrove-oss:
Thanks.  unistd.h is already #included. And as expected my unistd.h does include a definition of off_t.  So I think maybe the fix is in the GLibc module map.   Who would be the right person on the compiler team to drag in to help figure out why the importer isn't seeing the definition of off_t when it processes the header file (but was seeing it before the march 24 drop of the swift compiler).

MadCoder:
I don't know you should ask on one of the swift mailing lists. FWIW I don't think the dispatch build uses modules on linux, so I'm slightly surprised at the issue here.

@mwwa maybe you can help given the work you're doing on the overlay?

dgrove-oss:
@MadCoder   I've played with various options suggested by @gribozavr on swift-dev list today, and I think this change is the least disruptive fix that is possible in the medium term (ie, without getting the system headers changed to be more module friendly).   Hopefully guarding the include of stdio.h will make this attempt more acceptable than the previous one.

MadCoder:
that looks reasonnable to me

PR #64, opened 2016-03-31
Another fix for libpwq configuration

dgrove-oss:
Setting a configure variable that is tested in
a later AM_CONDITIOANL to true, is not the same as
setting it to true


dgrove-oss:
Merged to experimental/foundation in #67.  Leaving this open and not deleting branch until it is also merged to master

drewcrawford:
This resolves SR-1242.

Please merge.

PR #65, opened 2016-03-31
guard pthread_workqueue_signal_np calls

dgrove-oss:
Fixes build when on Linux, but not building with libpwq.
Not clear if this configuration is actually that useful,
but it should either (a) build/run as well as possible or
(b) be treated as a configuration error and fail the build
with a useful error message.  This patch enables option (a).


dgrove-oss:
No longer relevant; calls to pthread_workqueue_signal_np were removed.

PR #66, opened 2016-04-01
SR-1101 : clang 3.9 doesn't allow typeof on bit-field

dgrove-oss:
Avoid using an expression whose type is a bitfield in
macros like slowpath or dispatch_assert that internally
use typeof on their argument expression.


dgrove-oss:
Problem reported by team porting libdispatch to Linux S/390 where they need to use latest version of clang (https://bugs.swift.org/browse/SR-1101 for details).

dgrove-oss:
No longer needed; closing.

PR #67, opened 2016-04-06
Another fix for libpwq configuration

dgrove-oss:
Setting a configure variable that is tested in
a later AM_CONDITIOANL to true, is not the same as
setting it to true


dgrove-oss:
Replicates #64 (master branch) to the experimental branch

PR #68, opened 2016-04-07
Build fix for Linux overlay with 03242016 compiler.

dgrove-oss:
With the 03242016 development drop, building the
Dispatch overlay on Linux fails because mode_t and
off_t are not defined.  This is a minimal fix to
include the needed header files and get the build
working again.


dgrove-oss:
Repeat of #62 for experimental/foundation branch.

PR #70, opened 2016-04-13
change prototype of functions returning nil to use ImplictlyUnwrapped…

dgrove-oss:
…Optional

Fixes build break with 20160412a development drop by restoring !
annotation on the return types of methods that return nil.
Picked ! (not ?) to match swift/public/SDK/Dispatch/Dispatch.swift.


dgrove-oss:
master-branch variant of #69, which is merged to the experimental/foundation branch

PR #76, opened 2016-06-08
libkqueue as git submodule

ianpartridge:
This commit adds the option to build libkqueue into libdispatch as a git submodule.  This is so we can pick up fixes to libkqueue which are not available via the system package.

If the libkqueue git submodule is not present, then the system package is used as before.


INSTALL
```diff
@@ -109,7 +109,7 @@ Prepare your system
  2. Install dtrace (to generate provider.h)
     sudo apt-get install systemtap-sdt-dev
  3. Install libdispatch pre-reqs
-    sudo apt-get install libblocksruntime-dev libkqueue-dev libbsd-dev
+    sudo apt-get install libblocksruntime-dev libbsd-dev
```
MadCoder: I wouldn't *quite* do that, as I think that it won't fly long term with distributions to not use the system's libkqueue. so I wouldn't touch this, but use pkg-config or some method in configure to check that the system libkqueue is recent enough to have the right fixes instead.

ianpartridge: We did an identical change when we added libpwq as a git submodule.

MadCoder: I know, but the situation with libpwq is different: libdispatch is its only client on the system *ever* so having dispatch "disptribute" it is fine.

libkqueue is different, it has other clients on the system and other OSS projects use it, so while I understand the motivation of this patch for swift development purposes, I don't think we can ship like that.

MadCoder: IOW I think it's completely fine to be *able* to use the submodule, but that should be for development purposes only, unlike libpwq

seabaylea: @MadCoder that makes sense. Once we have a final list of fixes needed and know which level of libkqueue those fixes make it in to, we can definitely take the approach of only using a submodule if the system installed levels are too old.

seabaylea:
Merging to get the libkqueue fixes into the 'experimental' working branch

PR #77, opened 2016-06-08
SR-1101 : clang 3.9 doesn't allow typeof on bit-field

dgrove-oss:
Avoid using an expression whose type is a bitfield in
macros like slowpath or dispatch_assert that internally
use typeof on their argument expression.


dgrove-oss:
pull request for experimental/foundation branch so it will build with clang 3.9.

PR #80, opened 2016-06-09
Provide Swift 3 overlay

seabaylea:
Provide a Swift 3 language syntax overlay into experimental/foundation to allow the integration of libdispatch into Foundation and the full Swift toolchain to progress.


seabaylea:
I've built full toolchains on Ubuntu 14.04, 15.04 and 16.04, and tested using some NSURLSession tests. I've also tested with using the Kitura web framework, which invokes Dispatch directly rather than via Foundation.

PR #81, opened 2016-06-15
Linux port of libdispatch-685 merge

dgrove-oss:
Minimal set of changes to get sources merged
from libdispatch-685 to compile and run on Linux.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -295,9 +295,6 @@ AC_MSG_CHECKING([what semaphore type to use]);
 AS_IF([test "x$have_mach" = "xtrue"],
   [AC_DEFINE(USE_MACH_SEM, 1, [Define to use Mach semaphores])
     AC_MSG_RESULT([Mach semaphores])],
-  [test "x$have_futex" = "xtrue"],
-  [AC_DEFINE(USE_FUTEX_SEM, 1, [Define to use Futex semaphores])
-    AC_MSG_RESULT([Futex semaphores])],
   [test "x$have_sem_init" = "xtrue"],
   [AC_DEFINE(USE_POSIX_SEM, 1, [Define to use POSIX semaphores])
     AC_MSG_RESULT([POSIX semaphores])],
--- a/os/linux_base.h
+++ b/os/linux_base.h
@@ -13,92 +13,60 @@
 #ifndef __OS_LINUX_BASE__
 #define __OS_LINUX_BASE__
 
-// #include <sys/event.h>
 #include <sys/user.h>
+#include <sys/param.h>
 
-// marker for hacks we have made to make progress
-#define __LINUX_PORT_HDD__ 1
-
-/*
- * Stub out defines for some mach types and related macros
- */
-
-typedef uint32_t mach_port_t;
-
-#define  MACH_PORT_NULL (0)
-#define  MACH_PORT_DEAD (-1)
-
-#define EVFILT_MACHPORT (-8)
-
-typedef uint32_t mach_error_t;
-
-typedef uint32_t mach_vm_size_t;
-
-typedef uint32_t mach_msg_return_t;
-
-typedef uintptr_t mach_vm_address_t;
-
-typedef uint32_t dispatch_mach_msg_t;
-
-typedef uint32_t dispatch_mach_t;
-
-typedef uint32_t dispatch_mach_reason_t;
-
-typedef uint32_t voucher_activity_mode_t;
-
-typedef uint32_t voucher_activity_trace_id_t;
-
-typedef uint32_t voucher_activity_id_t;
-
-typedef uint32_t _voucher_activity_buffer_hook_t;;
-
-typedef uint32_t voucher_activity_flag_t;
-
-typedef struct
-{
-} mach_msg_header_t;
-
-
-typedef void (*dispatch_mach_handler_function_t)(void*, dispatch_mach_reason_t,
-						 dispatch_mach_msg_t, mach_error_t);
-
-typedef void (*dispatch_mach_msg_destructor_t)(void*);
-
-// Print a warning when an unported code path executes.
-#define LINUX_PORT_ERROR()  do { printf("LINUX_PORT_ERROR_CALLED %s:%d: %s\n",__FILE__,__LINE__,__FUNCTION__); } while (0)
-
-/*
- * Stub out defines for other missing types
- */
-
-#if __linux__
-// we fall back to use kevent
-#define kevent64_s kevent
-#define kevent64(kq,cl,nc,el,ne,f,to)  kevent(kq,cl,nc,el,ne,to)
+#if __GNUC__
+#define OS_EXPECT(x, v) __builtin_expect((x), (v))
+#else
+#define OS_EXPECT(x, v) (x)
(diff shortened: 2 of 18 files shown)


configure.ac
```diff
@@ -295,9 +295,6 @@ AC_MSG_CHECKING([what semaphore type to use]);
 AS_IF([test "x$have_mach" = "xtrue"],
   [AC_DEFINE(USE_MACH_SEM, 1, [Define to use Mach semaphores])
     AC_MSG_RESULT([Mach semaphores])],
-  [test "x$have_futex" = "xtrue"],
```
MadCoder: that is very good, you should also get rid of all the code guarded with `#if USE_FUTEX_SEM` that is still around

os/linux_base.h
```diff
@@ -48,12 +50,25 @@ typedef uint32_t voucher_activity_mode_t;
...
+typedef uint32_t firehose_activity_flags_t;
+
+typedef uint32_t firehose_activity_mode_t;
+
+typedef uint32_t firehose_tracepoint_id_t;
+
+typedef uint32_t firehose_stream_t;
+
+
 typedef struct
```
MadCoder: this is not valid c++ and could be void * instead

os/linux_base.h
```diff
@@ -38,6 +38,8 @@ typedef uint32_t mach_msg_return_t;
 
 typedef uintptr_t mach_vm_address_t;
 
+typedef uint32_t mach_msg_bits_t;
```
MadCoder: all this should move to linux_stubs.h because linux_base.h is a public API header that should not define/export these types that are only used to build dispatch.

os/linux_base.h
```diff
@@ -123,12 +138,67 @@ typedef void (*dispatch_mach_msg_destructor_t)(void*);
 #endif
 #define OS_NOTHROW
 
+#ifdef OS_ASSUME_NONNULL_BEGIN
```
MadCoder: this is not the right definition, doing this will break the overlay, please take the definition from <os/base.h> in an OS X SDK

os/linux_base.h
```diff
@@ -123,12 +138,67 @@ typedef void (*dispatch_mach_msg_destructor_t)(void*);
...
+#undef OS_SWIFT_UNAVAILABLE
+#endif
+#define OS_SWIFT_UNAVAILABLE(errstring)
+
 
 // These and similar macros come from Availabilty.h on OS X
 // Need a better way to do this long term.
 #define __OSX_AVAILABLE_BUT_DEPRECATED(a,b,c,d) //
 #define __OSX_AVAILABLE_BUT_DEPRECATED_MSG(a,b,c,d,msg) //
+#define __OSX_DEPRECATED(a,b,c) //
```
MadCoder: these should go in <dispatch/dispatch.h> like other similar ones

os/linux_base.h
```diff
@@ -123,12 +138,67 @@ typedef void (*dispatch_mach_msg_destructor_t)(void*);
...
+#endif
+#define __TVOS_AVAILABLE(version)
+
+#ifdef __WATCHOS_AVAILABLE
+#undef __WATCHOS_AVAILABLE
+#endif
+#define __WATCHOS_AVAILABLE(version)
+
+#ifndef os_likely
+#define os_likely(cond)  (cond)
```
MadCoder: this should be `__builtin_expect(!!(cond), 1)`, see the macOS SDK (`<os/base.h>`)

src/firehose/firehose_internal.h
```diff
@@ -27,6 +27,7 @@
 #define __MigTypeCheck 1
 #endif
 
+#ifdef OS_FIREHOSE_SPI
```
MadCoder: should be `#if` not ifdef.

src/shims/lock.h
```diff
@@ -95,7 +96,11 @@ typedef pid_t dispatch_lock_owner;
 #define DLOCK_OWNER_MASK			((dispatch_lock)FUTEX_TID_MASK)
 #define DLOCK_WAITERS_BIT			((dispatch_lock)FUTEX_WAITERS)
 #define DLOCK_FAILED_TRYLOCK_BIT	((dispatch_lock)FUTEX_OWNER_DIED)
-#define _dispatch_tid_self()		/* FIXME cached in TSD in the swift port */
+
+#include <unistd.h>
```
MadCoder: please group the `#include`'s together after the `defined(__linux__)` and keep the order of the `#define _dispatch_tid_self()` to mirror the order of defines for OS X

src/source.c
```diff
@@ -622,7 +622,7 @@ _dispatch_source_kevent_unregister(dispatch_source_t ds)
 		}
 	} else if (!ds->ds_is_direct_kevent) {
 		ds->ds_dkev = NULL;
-		dispatch_assert(ds->ds_is_installed);
+		dispatch_assert(ds->ds_is_installed == 1);
```
MadCoder: I'd rather have a cast to bool. if the bitfield is `int` the value may be -1 and this would be wrong.

dgrove-oss:
revised based on feedback; squashed & force-pushed.

src/queue.c
```diff
@@ -5750,3 +5764,15 @@ _dispatch_context_cleanup(void *ctxt)
 	DISPATCH_INTERNAL_CRASH(ctxt,
 			"Premature thread exit while a dispatch context is set");
 }
+
+#if !DISPATCH_COCOA_COMPAT
+// FIXME:  Need real implementation of this + related functions for Linux
+DISPATCH_NOINLINE
+void
+_dispatch_queue_class_override_drainer(dispatch_queue_t dq,
```
MadCoder: that's not the right patch please remove that hunk, and instead do this:

```diff
 DISPATCH_NOINLINE
 void
 _dispatch_queue_wakeup(dispatch_queue_t dq, pthread_priority_t pp,
 		dispatch_wakeup_flags_t flags)
 {
 	dispatch_queue_wakeup_target_t target = DISPATCH_QUEUE_WAKEUP_NONE;
 
 	if (_dispatch_queue_class_probe(dq)) {
 		target = DISPATCH_QUEUE_WAKEUP_TARGET;
 	}
 	if (target) {
 		return _dispatch_queue_class_wakeup(dq, pp, flags, target);
+#if HAVE_PTHREAD_WORKQUEUE_QOS
 	} else if (pp) {
 		return _dispatch_queue_class_override_drainer(dq, pp, flags);
+#endif
 	} else if (flags & DISPATCH_WAKEUP_CONSUME) {
 		return _dispatch_release_tailcall(dq);
 	}
 }
```

MadCoder: the reason is that `_dispatch_queue_class_override_drainer()` only makes sense if you have QoS support which linux doesn't so there's no point having the code in the first place, as you can see _dispatch_queue_class_override_drainer() is not reachable because it is itself under `#if HAVE_PTHREAD_WORKQUEUE_QOS`

dgrove-oss:
Thanks.  updated.  Also had to guard calls to _dispatch_queue_class_override_drainer in source.c

MadCoder:
Makes perfect sense.

PR #82, opened 2016-06-16
TLS: Thread Local Storage 

frankeh:
Bring up previous patch for Thread Local Storage utilization over explicit keys to 685 base.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -67,7 +67,10 @@ void *(*_dispatch_begin_NSAutoReleasePool)(void);
 void (*_dispatch_end_NSAutoReleasePool)(void *);
 #endif
 
-#if !DISPATCH_USE_DIRECT_TSD
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+__thread struct dispatch_tsd __dispatch_tsd;
+pthread_key_t __dispatch_tsd_key;
+#elif !DISPATCH_USE_DIRECT_TSD
 pthread_key_t dispatch_queue_key;
 pthread_key_t dispatch_frame_key;
 pthread_key_t dispatch_cache_key;
@@ -82,7 +85,7 @@ pthread_key_t dispatch_bcounter_key;
 pthread_key_t dispatch_sema4_key;
 pthread_key_t dispatch_voucher_key;
 pthread_key_t dispatch_deferred_items_key;
-#endif // !DISPATCH_USE_DIRECT_TSD
+#endif // !DISPATCH_USE_DIRECT_TSD && !DISPATCH_USE_THREAD_LOCAL_STORAGE
 
 #if VOUCHER_USE_MACH_VOUCHER
 dispatch_once_t _voucher_task_mach_voucher_pred;
@@ -184,10 +187,6 @@ const struct dispatch_tsd_indexes_s dispatch_tsd_indexes = {
 	.dti_voucher_index = dispatch_voucher_key,
 	.dti_qos_class_index = dispatch_priority_key,
 };
-#else // DISPATCH_USE_DIRECT_TSD
-#ifndef __LINUX_PORT_HDD__
-#error Not implemented on this platform
-#endif
 #endif // DISPATCH_USE_DIRECT_TSD
 
 // 6618342 Contact the team that owns the Instrument DTrace probe before
--- a/src/queue.c
+++ b/src/queue.c
@@ -894,6 +894,9 @@ libdispatch_init(void)
 #endif
 #endif
 
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+	_dispatch_thread_key_create(&__dispatch_tsd_key, _libdispatch_tsd_cleanup);
+#else
 	_dispatch_thread_key_create(&dispatch_queue_key, _dispatch_queue_cleanup);
 	_dispatch_thread_key_create(&dispatch_deferred_items_key,
 			_dispatch_deferred_items_cleanup);
@@ -912,6 +915,7 @@ libdispatch_init(void)
 		_dispatch_thread_key_create(&dispatch_sema4_key,
 				_dispatch_thread_semaphore_dispose);
 	}
+#endif
 #endif
 
 #if DISPATCH_USE_RESOLVERS // rdar://problem/8541707
@@ -964,6 +968,59 @@ _dispatch_get_mach_host_port(void)
 	dispatch_once_f(&_dispatch_mach_host_port_pred, NULL,
 			_dispatch_mach_host_port_init);
 	return _dispatch_mach_host_port;
+}
+#endif
+
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+#include <unistd.h>
+#include <sys/syscall.h>
+
+#ifdef SYS_gettid
+DISPATCH_ALWAYS_INLINE
+static inline pid_t
+gettid(void)
+{
+	return (pid_t) syscall(SYS_gettid);
+}
+#else
+#error "SYS_gettid unavailable on this system"
+#endif
+
+#define _tsd_call_cleanup(k, f)  do { \
+		if ((f) && tsd->k) ((void(*)(void*))(f))(tsd->k); \
+    } while (0)
+
+void
+_libdispatch_tsd_cleanup(void *ctx)
+{
+	struct dispatch_tsd *tsd = (struct dispatch_tsd*) ctx;
+
+	_tsd_call_cleanup(dispatch_queue_key, _dispatch_queue_cleanup);
+	_tsd_call_cleanup(dispatch_frame_key, _dispatch_frame_cleanup);
+	_tsd_call_cleanup(dispatch_cache_key, _dispatch_cache_cleanup);
+	_tsd_call_cleanup(dispatch_context_key, _dispatch_context_cleanup);
+	_tsd_call_cleanup(dispatch_pthread_root_queue_observer_hooks_key,
+			NULL);
+	_tsd_call_cleanup(dispatch_defaultpriority_key, NULL);
+#if DISPATCH_PERF_MON && !DISPATCH_INTROSPECTION
+	_tsd_call_cleanup(dispatch_bcounter_key, NULL);
+#endif
+#if DISPATCH_LOCK_USE_SEMAPHORE_FALLBACK
+	_tsd_call_cleanup(dispatch_sema4_key, _dispatch_thread_semaphore_dispose);
+#endif
+	_tsd_call_cleanup(dispatch_priority_key, NULL);
+	_tsd_call_cleanup(dispatch_voucher_key, _voucher_thread_cleanup);
+	_tsd_call_cleanup(dispatch_deferred_items_key,
+			_dispatch_deferred_items_cleanup);
+	tsd->tid = 0;
+}
+
+DISPATCH_NOINLINE
+void
+libdispatch_tsd_init(void)
+{
+	pthread_setspecific(__dispatch_tsd_key, &__dispatch_tsd);
+	__dispatch_tsd.tid = gettid();
 }
 #endif
 
--- a/src/shims/linux_stubs.c
+++ b/src/shims/linux_stubs.c
@@ -57,3 +57,4 @@ unsigned short dispatch_callout__return_semaphore;
 unsigned short dispatch_queue__push_semaphore;
 void (*_dispatch_block_special_invoke)(void*);
 struct dispatch_queue_attr_s _dispatch_queue_attr_concurrent;
+
--- a/src/shims/tsd.h
+++ b/src/shims/tsd.h
@@ -92,10 +92,76 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
 	if (!*k || !d) return;
 	dispatch_assert_zero(pthread_key_init_np((int)*k, d));
 }
+#elif DISPATCH_USE_THREAD_LOCAL_STORAGE
+
+DISPATCH_TSD_INLINE
+static inline void
+_dispatch_thread_key_create(pthread_key_t *k, void (*d)(void *))
+{
+	dispatch_assert_zero(pthread_key_create(k, d));
+}
+
+struct dispatch_tsd {
+	pid_t tid;
+	void *dispatch_queue_key;
+	void *dispatch_frame_key;
+	void *dispatch_cache_key;
+	void *dispatch_context_key;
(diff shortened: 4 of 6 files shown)


src/shims/linux_stubs.c
```diff
@@ -63,3 +64,8 @@ unsigned short dispatch_callout__return_semaphore;
 unsigned short dispatch_queue__push_semaphore;
 void (*_dispatch_block_special_invoke)(void*);
 struct dispatch_queue_attr_s _dispatch_queue_attr_concurrent;
+
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
```
MadCoder: this should move in init.c

src/shims/linux_stubs.c
```diff
@@ -40,8 +40,9 @@ char* mach_error_string(mach_msg_return_t x) {
 void mach_vm_deallocate() { LINUX_PORT_ERROR();  }
 
 mach_port_t pthread_mach_thread_np(void) {
-  return (pid_t)syscall(SYS_gettid);
```
MadCoder: this should not be touched, I just asked @dgrove-oss to remove this in another pull request and that will conflict

src/init.c
```diff
@@ -67,7 +67,7 @@ void *(*_dispatch_begin_NSAutoReleasePool)(void);
 void (*_dispatch_end_NSAutoReleasePool)(void *);
 #endif
 
-#if !DISPATCH_USE_DIRECT_TSD
```
MadCoder: make that

```diff
-#if !DISPATCH_USE_DIRECT_TSD
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+__thread struct dispatch_tsd __dispatch_tsd;
+pthread_key_t __dispatch_tsd_key;
+-#elif !DISPATCH_USE_DIRECT_TSD
```

instead of linux_sbuts.c

src/queue.c
```diff
@@ -835,6 +835,10 @@ _dispatch_root_queues_init_once(void *context DISPATCH_UNUSED)
 	}
 }
 
+#if DISPATCH_USE_THREAD_LOCAL_STORAGE
+	static void _libdispatch_tsd_cleanup(void *ctx); // we need declaration
```
MadCoder: this prototype should live in the shims/tsd.h header and the function live in init.c

src/queue.c
```diff
@@ -967,6 +975,63 @@ _dispatch_get_mach_host_port(void)
...
+gettid(void)
+{
+	return (pid_t) syscall(SYS_gettid);
+}
+#else
+#error "SYS_gettid unavailable on this system"
+#endif
+
+#define _tsd_call_cleanup(k, f)  do { \
+		if (f && tsd->k) ((void(*)(void*))(f))(tsd->k); \
```
MadCoder: `if ((f)` ?

src/queue.c
```diff
@@ -967,6 +975,63 @@ _dispatch_get_mach_host_port(void)
...
+	_tsd_call_cleanup(dispatch_cache_key, _dispatch_cache_cleanup);
+	_tsd_call_cleanup(dispatch_context_key, _dispatch_context_cleanup);
+	_tsd_call_cleanup(dispatch_pthread_root_queue_observer_hooks_key,
+			NULL);
+	_tsd_call_cleanup(dispatch_defaultpriority_key, NULL);
+#if DISPATCH_PERF_MON && !DISPATCH_INTROSPECTION
+	_tsd_call_cleanup(dispatch_bcounter_key, NULL);
+#endif
+#if DISPATCH_LOCK_USE_SEMAPHORE_FALLBACK
+	_tsd_call_cleanup(dispatch_sema4_key,
```
MadCoder: the macro already does the cast, so you can avoid it, also linux doesn't have `DISPATCH_LOCK_USE_SEMAPHORE_FALLBACK` set every anyway, but it's probably good to keep it that way.

src/shims/tsd.h
```diff
@@ -92,10 +92,75 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+
+DISPATCH_TSD_INLINE
+static inline void
+_dispatch_thread_key_create(pthread_key_t *k, void (*d)(void *))
+{
+	dispatch_assert_zero(pthread_key_create(k, d));
+}
+
+struct dispatch_tsd {
+	bool initialized;
```
MadCoder: I would use `tid != 0` to know that it's initialized, that would make less loads in fast-paths where we care.

src/shims/tsd.h
```diff
@@ -92,10 +92,75 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+}
+
+#define _dispatch_thread_getspecific(key) \
+	(_dispatch_get_tsd_base()->key)
+#define _dispatch_thread_setspecific(key, value) \
+	(void)(_dispatch_get_tsd_base()->key = (value))
+
+#define _dispatch_thread_getspecific_pair(k1, p1, k2, p2) \
+		( *(p1) = _dispatch_thread_getspecific(k1), \
+		  *(p2) = _dispatch_thread_getspecific(k2) )
```
MadCoder: please remove the extra space:

```diff
+#define _dispatch_thread_getspecific_pair(k1, p1, k2, p2) \
+		(*(p1) = _dispatch_thread_getspecific(k1), \
+		*(p2) = _dispatch_thread_getspecific(k2))
```

src/shims/tsd.h
```diff
@@ -92,10 +92,75 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+
+#define _dispatch_thread_getspecific_packed_pair(k1, k2, p) \
+		( (p)[0] = _dispatch_thread_getspecific(k1), \
```
MadCoder: ditto

src/shims/tsd.h
```diff
@@ -92,10 +92,75 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+		  (p)[1] = _dispatch_thread_getspecific(k2) )
+
+#define _dispatch_thread_setspecific_pair(k1, p1, k2, p2) \
+		( _dispatch_thread_setspecific(k1,p1), \
+		  _dispatch_thread_setspecific(k2,p2) )
```
MadCoder: ditto

src/shims/tsd.h
```diff
@@ -92,10 +92,75 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+
+#define _dispatch_thread_setspecific_packed_pair(k1, k2, p) \
+		( _dispatch_thread_setspecific(k1,(p)[0]), \
```
MadCoder: ditto

src/shims/tsd.h
```diff
@@ -105,7 +170,8 @@ extern pthread_key_t dispatch_introspection_key;
 #elif DISPATCH_PERF_MON
 extern pthread_key_t dispatch_bcounter_key;
 #endif
-extern pthread_key_t dispatch_cache_key;
+extern pthread_key_t dispatch_sema4_key;
+extern pthread_key_t dispatch_priority_key;
```
MadCoder: eeew thanks

src/shims/tsd.h
```diff
@@ -228,6 +296,8 @@ _dispatch_thread_setspecific_packed_pair(pthread_key_t k1, pthread_key_t k2,
 #if DISPATCH_USE_DIRECT_TSD
 #define _dispatch_thread_port() ((mach_port_t)(uintptr_t)\
 		_dispatch_thread_getspecific(_PTHREAD_TSD_SLOT_MACH_THREAD_SELF))
+#elif DISPATCH_USE_THREAD_LOCAL_STORAGE
+#define _dispatch_thread_port() ((mach_port_t)(_dispatch_get_tsd_base()->tid))
```
MadCoder: I really don't like this name, `_dispatch_thread_port()` is a mach concept, so we should ban the use of this on the linux port, and instead have `_dispatch_tid_self()` in lock.h expand to `(_dispatch_get_tsd_base()->tid)`.

and if there are calls to _dispatch_thread_port() in non mach code, we should most likely port them to `_dispatch_tid_self()` or guard the code with a `#if HAVE_MACH`

IOW that thing here should be in a `#if HAVE_MACH` this way:

```C
#if TARGET_OS_WIN32
#define _dispatch_thread_port() ((mach_port_t)0)
#elif HAVE_MACH
#if DISPATCH_USE_DIRECT_TSD
#define _dispatch_thread_port() ((mach_port_t)(uintptr_t)\
  		_dispatch_thread_getspecific(_PTHREAD_TSD_SLOT_MACH_THREAD_SELF))
#else
#define _dispatch_thread_port() pthread_mach_thread_np(_dispatch_thread_self())
#endif
#endif // HAVE_MACH
```

MadCoder: Technically definining it for Windows is not the right way, but let's leave that part alone. it should also define `_dispatch_tid_self()` eventually if someone decides to re-port dispatch in the future, the same way I'm asking you to do for linux.

src/queue.c
```diff
@@ -967,6 +975,63 @@ _dispatch_get_mach_host_port(void)
...
+
+DISPATCH_EXPORT DISPATCH_NOINLINE
+void
+libdispatch_tsd_init(void)
+{
+	__dispatch_tsd.tid = gettid();
+	pthread_setspecific(__dispatch_tsd_key, &__dispatch_tsd);
+	__dispatch_tsd.initialized = true;
+}
+
```
MadCoder: spurious extra line

src/queue.c
```diff
@@ -967,6 +975,63 @@ _dispatch_get_mach_host_port(void)
...
+#endif
+	_tsd_call_cleanup(dispatch_priority_key, NULL);
+	_tsd_call_cleanup(dispatch_voucher_key, _voucher_thread_cleanup);
+	_tsd_call_cleanup(dispatch_deferred_items_key,
+			_dispatch_deferred_items_cleanup);
+	tsd->tid = -1;
+	tsd->initialized = false;
+}
+
+DISPATCH_EXPORT DISPATCH_NOINLINE
```
MadCoder: DISPATCH_EXPORT here is not valid, it should not be exported.

src/queue.c
```diff
@@ -967,6 +975,63 @@ _dispatch_get_mach_host_port(void)
...
+#endif
+#if DISPATCH_LOCK_USE_SEMAPHORE_FALLBACK
+	_tsd_call_cleanup(dispatch_sema4_key,
+			(void (*)(void *))_dispatch_thread_semaphore_dispose);
+#endif
+	_tsd_call_cleanup(dispatch_priority_key, NULL);
+	_tsd_call_cleanup(dispatch_voucher_key, _voucher_thread_cleanup);
+	_tsd_call_cleanup(dispatch_deferred_items_key,
+			_dispatch_deferred_items_cleanup);
+	tsd->tid = -1;
```
MadCoder: 0 is an invalid pid. -1 isn't afaict.

src/shims/tsd.h
```diff
@@ -92,10 +92,75 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+
+extern __thread struct dispatch_tsd __dispatch_tsd;
+extern pthread_key_t __dispatch_tsd_key;
+extern void libdispatch_tsd_init(void);
+
+DISPATCH_ALWAYS_INLINE
+static inline struct dispatch_tsd *
+_dispatch_get_tsd_base(void)
+{
+	if (slowpath(!__dispatch_tsd.initialized)) {
```
MadCoder: please use `unlikely()` we're trying to phase out `slowpath/fastpath` that don't work well for clang reasons (can explain offline from this review)

MadCoder:
done with the first round of comments.

src/shims/tsd.h
```diff
@@ -92,10 +92,75 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+		libdispatch_tsd_init();
```
MadCoder: Adding a `OS_COMPILER_CAN_ASSUME(__dispatch_tsd.tid != 0)` may help clang to also not call that lazy init several times for the cases when several TSD accesses are performed

src/shims/tsd.h
```diff
@@ -92,10 +92,75 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+	void *dispatch_priority_key;
+	void *dispatch_voucher_key;
+	void *dispatch_deferred_items_key;
+};
+
+extern __thread struct dispatch_tsd __dispatch_tsd;
+extern pthread_key_t __dispatch_tsd_key;
+extern void libdispatch_tsd_init(void);
+
+DISPATCH_ALWAYS_INLINE
```
MadCoder: you should also add `DISPATCH_PURE` here, which can help have the compiler hoist the lazy init to the top.

src/shims/linux_stubs.c
frankeh: Yes we agree, once you merge Dave's change, we will fix this. Right now we need it to get this compile and run.

src/queue.c
frankeh: I don't see what you are requesting here. You want the _tsd_call_cleanup to do the cast
to a ((struct dispatch_tsd*) ctx)  to obtain the equivalence of the tsd?

frankeh:
Besides the other two comments: 
Moving the functions for libdispatch_tsd_cleanup to init.c requires that all the key specific cleanup functions that are static in queue.c now have to be externalized and prototypes need to be provided. 
Is that really what you want ?

frankeh:
Finally DISPATCH_PURE on libdispatch_tsd_init   results in crashes (verified with only that change).
So we didn't do that.

src/queue.c
MadCoder: _tsd_call_cleanup(dispatch_sema4_key... has a cast that is not useful on its second argument.

MadCoder:
ok we can keep libdispatch_tsd_cleanup() in queue.c then.
why did making it PURE crashes? that makes no sense to me (I believe you I'd just be interested to see a crash log)

MadCoder:
the patch looks good beside cleaning up `gettid/_dispatch_thread_port` which we'll do once the other request is merged. thanks

MadCoder:
I just merged #86 so please rebase, squash and address the remaining comments

frankeh:
Rebase squashed based on the comments you had left and we fixed. (thanx)

MadCoder:
\o/

PR #84, opened 2016-06-16
Avoid overwriting config.h with generated header

dgrove-oss:
config.h is generated by the autotools build process.
Having it stored in git causes useless diffs and the
need for git checkout operations when rebasing because
the generated version differs from the version in git.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -6,7 +6,7 @@ AC_PREREQ(2.69)
 AC_INIT([libdispatch], [1.3], [libdispatch@macosforge.org], [libdispatch], [http://libdispatch.macosforge.org])
 AC_REVISION([$$])
 AC_CONFIG_AUX_DIR(config)
-AC_CONFIG_HEADER([config/config.h])
+AC_CONFIG_HEADER([config/config_ac.h])
 AC_CONFIG_MACRO_DIR([m4])
 ac_clean_files=a.out.dSYM
 AM_MAINTAINER_MODE
--- a/src/internal.h
+++ b/src/internal.h
@@ -27,7 +27,11 @@
 #ifndef __DISPATCH_INTERNAL__
 #define __DISPATCH_INTERNAL__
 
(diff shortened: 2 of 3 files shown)


dgrove-oss:
If this doesn't cause problems with the XCode build, it would be nice to not have config.h in git.  If it does cause a problem, it is ok to just reject this pull request.

It is just a minor annoyance (that I hit 10 times this morning when changing branches, rebasing, etc...).

MadCoder:
the problem is that it will cause issues with the xcode based internal port at Apple, but we may be able to carry with that file removed from the git repository (IOW not backport this to Darwin).

@das is that ok with you?

das:
as long as we have the Xcode project in the repo, we should also keep the config.h so that you can build on Darwin with Xcode. If we were to remove the config.h we should also remove the project and its associated xcconfig files etc (as it would be broken anyway)

FWIW I've always kept this as part of the opensource drop to document how we build libdispatch on Darwin (and enable others to potentially replicate that), so I'm not a fan such a removal.

IIRC there is a way to ask autoheader to use a different filename than config.h, ISTM that would be a better way to solve this (and use __has_include() to switch between config.h and that header).

We could also rename the config.h used by the Xcode build (but that name has history associated at this point)

MadCoder:
yeah I was assuming that you would say something like that and do that trick.

the other way is also to play with include paths and have the ./configure generated one shadow the config/config.h one

@dgrove-oss can you try to play tricks in that direction instead please?

MadCoder:
by "that direction" I mean either using include paths based hiding or using a __has_header() kind of thing which is probably cleaner.

dgrove-oss:
sure.  I'll take a look tomorrow.

dgrove-oss:
Thanks for the alternate suggestion; changing the name of the generated file and using has_include was very easy.  Pull request updated.

PR #85, opened 2016-06-16
avoid overflow in dispatch_io test

dgrove-oss:
Test was failing because number of bytes read overflows int32_t.
Changing to a uint32_t is sufficient to avoid overflow.

Changing to int64_t would also work, but is a larger change (since
we then need to provide an OSAtomicAdd64 function on Linux).


MadCoder:
you can also use stdatomic ;)
but that will do

PR #86, opened 2016-06-16
eliminate mach_task_self from linux_stubs

dgrove-oss:
guard stray usage of mach_task_self with #if HAVE_MACH
and remove mach_test_self() from linux_stubs.

### DIFF
--- a/src/data.c
+++ b/src/data.c
@@ -155,10 +155,12 @@ _dispatch_data_destroy_buffer(const void* buffer, size_t size,
 		free((void*)buffer);
 	} else if (destructor == DISPATCH_DATA_DESTRUCTOR_NONE) {
 		// do nothing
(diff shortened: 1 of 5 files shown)


MadCoder:
could you try to get rid of pthread_mach_thread_np() in the same patch? because that one is also a MACH-ism and it would fit very well in that change set

dgrove-oss:
Updated.  

I'm pretty sure using _dispatch_tid_self in place of dispatch_thread_port is the right fix to avoid the expansion out to pthread_mach_thread_np, but if not let me know and I'll try again.

MadCoder:
Yup using `_dispatch_tid_self()` when it's a lock value is fine. it would be conceptually incorrect to use it as the argument to a mach call which you didn't do, so I think it's perfectly fine that way!

PR #87, opened 2016-06-17
resolve all compiler warnings

chsu-ibm:
Add missing declarations in linux_stubs.h for 3 functions:
pthread_mach_thread_np, mach_error_string and mach_vm_deallocate

For empty struct warning from mach_msg_header_t defined in linux_stubs.h,
I suppress the warning by add -Wno-extern-c-compat


MadCoder:
no, this isn't the right way, none of these functions should be called on linux, so the code should be compiled out. #86 is already doing work in that direction, this work is directly conflicting.

PR #89, opened 2016-06-22
Import Swift 3 overlay from swift repo

das:
* import Dispatch overlay and apinotes from swift-3.0-preview-2-branch of github swift repo
* add new overlay sourcefiles to src/Makefile.am (does not build)


parkera:
Awesome!

PR #90, opened 2016-06-22
Disable intermittently failing tests

seabaylea:
The following tests are sometimes failing when run under `make check-TESTS`:
```
dispatch_apply
dispatch_group
dispatch_concur		
dispatch_timer_short		
dispatch_sema			
dispatch_drift
```
All are intermittent, and some of these can hang (e.g. `dispatch_group` and `dispatch_sema`), so we can't just move them to the XFAIL list.

### DIFF
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -32,38 +32,42 @@ UNPORTED_TESTS=					\
 	dispatch_vnode
 
 TESTS=							\
-	dispatch_apply				\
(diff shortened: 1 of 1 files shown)


dgrove-oss:
We've decided to disable these tests on experimental/foundation to avoid blocking progress on CI of libdispatch.  If any of these tests are still problematic on master, we will investigate them there.

PR #97, opened 2016-07-05
Use SWIFT_SDK_OVERLAY_DISPATCH_EPOCH to detect that an overlay is bei…

dgrove-oss:
…ng built

Instead of defining the special macro __DISPATCH_BUILDING_SWIFT_MODULE__
to detect that dispatch headers are being processed by the clang importer,
check to see if SWIFT_SDK_OVERLAY_DISPATCH_EPOCH is defined.

The prime motivation for doing this is to avoid needing to define __DISPATCH_BUILDING_SWIFT_MODULE__ when compiling a Swift program that imports Dispatch.  A secondary benefit is that it makes additional functions unavailable for Swift3 (eg dispatch_once).  

This depends on a pull request (https://github.com/apple/swift/pull/3346) to Swift to change the importer to define SWIFT_SDK_OVERLAY_DISPATCH_EPOCH on non-Darwin platforms.

### DIFF
--- a/dispatch/dispatch.h
+++ b/dispatch/dispatch.h
@@ -44,8 +44,8 @@
 #endif
 
 #ifdef __linux__
-#ifdef __DISPATCH_BUILDING_SWIFT_MODULE__
-#include <stdio.h> // for off_t
+#if __has_feature(modules)
+#include <stdio.h> // for off_t (to match Glibc.modulemap)
 #endif
 #define DISPATCH_LINUX_UNAVAILABLE() \
 		__DISPATCH_UNAVAILABLE("This interface is unavailable on linux systems")
--- a/dispatch/object.h
+++ b/dispatch/object.h
@@ -108,7 +108,7 @@ typedef union {
 #define DISPATCH_RETURNS_RETAINED
 #endif
 
-#if OS_OBJECT_SWIFT3
+#if OS_OBJECT_SWIFT3 && OS_OBJECT_USE_OBJC
 #define DISPATCH_SOURCE_TYPE_DECL(name) \
 		DISPATCH_EXPORT struct dispatch_source_type_s \
 				_dispatch_source_type_##name; \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -149,7 +149,7 @@ SWIFT_GEN_FILES=	\
 	$(SWIFT_OBJ_FILES:%=%.~partial.swiftdoc) \
 	$(SWIFT_OBJ_FILES:%=%.~partial.swiftdeps)
 
-SWIFTC_FLAGS = -Xcc -D__DISPATCH_BUILDING_SWIFT_MODULE__=1 -Xcc -fmodule-map-file=$(abs_top_srcdir)/dispatch/module.modulemap -I$(abs_top_srcdir) -Xcc -fblocks
+SWIFTC_FLAGS = -Xcc -fmodule-map-file=$(abs_top_srcdir)/dispatch/module.modulemap -I$(abs_top_srcdir) -Xcc -fblocks
 
 $(abs_builddir)/swift/%.o:	$(abs_srcdir)/swift/%.swift
 	$(SWIFTC) -frontend -c $(SWIFT_ABS_SRC_FILES) -primary-file $< \


MadCoder:
Seems reasonable but I'd like for @mwwa to double check.

jrose-apple:
This doesn't seem correct to me. Won't it be set whenever *any* Swift code is used, not just when building the overlay?

dgrove-oss:
The problem I'm trying to work around could be a symptom of me not teaching the makefiles how to build the Dispatch and CDispatch modules correctly.  I've tried a bunch of things, but I could have missed something.

The problem is that the current approach of defining DISPATCH_BUILDING_SWIFT_MODULE only when actually building the overlay doesn't work very well.  When a user imports Dispatch into their Swift program, they also need to define DISPATCH_BUILDING_SWIFT_MODULE or the import of the dispatch header files fails.  It appears that when the Dispatch module is imported by a user program, the Swift compiler goes off, looks at the modulemap file that defines the Dispatch and CDispatch modules, and attempts to reimport the C dispatch header files for CDispatch. This import fails because DISPATCH_BUILDING_SWIFT_MODULE isn't defined (unless the user magically knows to do that via a command line argument to swiftc).

If there were a way to build the swiftmodule that would cut off this import when users import the module, that would be great and we could do that instead. 

A related thing that would be great to fix, when the user imports Dispatch, they currently need to add -Xcc -fblocks to the swiftc command line manually.  I'd really like to make that automatic (triggered by importing Dispatch), but don't know a good way of doing that.

jrose-apple:
So DISPATCH_BUILDING_SWIFT_MODULE really just means "Dispatch is going to be used from Swift"? :-/ We've mostly been *avoiding* providing a general query of that kind—so that people don't accidentally break the Swift side of that branch, or accidentally not expose some API—but maybe we need to give up on that. (I wouldn't want to make it Dispatch-specific, though.)

On the related thing: we're already considering just turning on blocks all the time for Swift-on-Linux, but we're wondering if we need to make them AnyObject-compatible first. I'm not in most of those discussions, though.

MadCoder:
@jrose-apple there's no way you can have something like dispatch work with swift with an equivalent of `#ifdef __OBJC__` but for swift because it's predating code that is made swifty, and not a new Swift API that has a C backend.

jrose-apple:
The original idea was that swift-corelibs-dispatch and any other framework not shipped with the OS should *always* be Swifty, at least on non-Apple platforms. In particular, since you're only using this guard to protect against the code being considered non-modular, checking for `__has_feature(modules)` is probably a better idea anyway, because that helps C clients too.

mwwa:
(As an aside: The Darwin version of this check uses it to alter the class hierarchy appearance under Swift, something that really does need to be checking for Swift-and-not-C, so that we don't alter the API that we've historically presented to Obj-C clients)

jrose-apple:
Yeah, we probably will need this at some point. I just want to do it properly when we get there, including discussing it on swift-evolution.

MadCoder:
@jrose-apple the problem with this is that the codebase is shared with macOS where it is not the case, and I don't think we should break people using libdispatch as a C library outside of swift on Linux. It actually is somehow broken today because objects are wrapped, but for the sake of interoperating with say, CF level APIs, being able to address libdispatch in C is valuable even on non macOS platforms.

so I don't think the intent works in practice, however pure and noble it is :)

Just for the sake of porting and testing libdispatch on new platforms it's much easier if you can isolate dispatch from Swift.

MadCoder:
also `__has_feature(modules)` is not what we want because it would inject `<stdio.h>` on Darwin platforms where you build in C with modules.

dgrove-oss:
fwiw, I think having a __SWIFT__ defined by the importer would be quite reasonable.  In the very short term, the __has_feature(modules) does work (as one would expect) and is inside a __linux__.  So we would take that as a an interim fix if your don't want to define the epoch off Darwin and getting __SWIFT__ would take a full swift evolution.   

I'd like to get the Swift3 overlay workable as soon as possible so we can start using it for corelibs-foundation.

MadCoder:
Right since it's inside a `__linux__` guard then let's go with the modules feature.

jrose-apple:
I'll bring up `__SWIFT__` on swift-dev, and we can move it to swift-evolution once we have a direction.

dgrove-oss:
updated to modules features test & squashed to single commit.

PR #98, opened 2016-07-05
release wrapped C objects when Swift object deinitialized

dgrove-oss:
ARC support in wrapping overlay. Add deinit methods to wrapping
Swift classes that call dispatch_release on the wrapped C objects.


MadCoder:
LGTM, @mwwa to confirm

PR #100, opened 2016-07-06
bug fix: DispatchWorkItem init failed to store group parameter

dgrove-oss:


MadCoder:
I won't merge this because we're disputing that it is a good interface. @mwwa  to follow up on this.

dgrove-oss:
ok; makes sense.   I ran into this while porting NSOperation in foundation to use the Swift 3 overlay.  Not blocking (I can test using the unmerged fix for now).

MadCoder:
well you should not use the `group:group` with a WorkItem and we're talking about adding an async() that takes a WorkItem and a group. so you will need some support at some point, but fixing the WorkItem init() feels wrong to us.

dgrove-oss:
ok.  So the likely outcome is that the _group instance field of WorkItem gets removed and we get a n additional async method on queue instead.   I'll pick it up again in the morning (have to head out now).

MadCoder:
yes that's what we're thinking about, @mwwa will lead that part on our end, I think he also has a few minor adjustments coming up.

mwwa:
Yeah. Stay tuned for a bunch of tweaks to the interface sometime next week. We should probably hold off on things like this PR until then, and then sync up all the changes at once.

shahmishal:
@swift-ci Please test

shahmishal:
@swift-ci Please test

seabaylea:
The CI failure should be fixed by #105. Note that when running the tests on a performance box I get the following failures:
```
FAIL: dispatch_concur
HANG: dispatch_timer_short
FAIL: dispatch_readsync
FAIL: dispatch_io
``

dgrove-oss:
@mwwa any update on the overlay API tweaks?

mwwa:
@dgrove-oss Yeah, I just pushed my changes, I won't be opening the PR *quite* yet but these should help.

https://github.com/mwwa/swift/compare/master...mwwa:libdispatch-overlay-fixes

PR #101, opened 2016-07-07
Enable CF runloop support for linux

dgrove-oss:
Here's a first cut at this.   It compiles, but I haven't actually tested it with foundation yet.  Implementation mostly based on the code from experimental/foundation, but I decided to more closely mirror the mach code and stash the eventfd in dq->do_ctxt instead of introducing a static variable just for the main queue. 

I'm not that happy with the HANDLE (windows) vs. int (linux) aspect of the patch.  I partially introduced dispatch_handle_t, but I also should introduce a symbolic constant for 0/NULL and be more careful in the linux-specific code to always use dispatch_handle_t instead of int.  Do we want dispatch_handle_t or similar?  On as similar note, we could define dispatch_handle_t to be mach_port_t if TARGET_OS_MAC and if we could change the name of _dispatch_get_main_queue_port_4CF to _dispatch_get_main_queue_handle_4CF we could avoid duplicating code.

Also, I used TARGET_OS_MAC in queue.c instead of HAVE_MACH.  This is different than other places in queue.c where we guard mach specific code.  Is this correct?

### DIFF
--- a/private/private.h
+++ b/private/private.h
@@ -166,31 +166,43 @@ void _dispatch_prohibit_transition_to_multithreaded(bool prohibit);
  * SPI for CoreFoundation/Foundation ONLY
  */
 
-#define DISPATCH_COCOA_COMPAT (TARGET_OS_MAC || TARGET_OS_WIN32)
+#if TARGET_OS_MAC
+#define DISPATCH_COCOA_COMPAT 1
+#elif defined(__linux__)
+#define DISPATCH_COCOA_COMPAT 1
+#else
+#define DISPATCH_COCOA_COMPAT 0
+#endif
 
 #if DISPATCH_COCOA_COMPAT
+
+#define DISPATCH_CF_SPI_VERSION 20160712
+
+#if TARGET_OS_MAC
+typedef mach_port_t dispatch_runloop_handle_t;
+#elif defined(__linux__)
+typedef int dispatch_runloop_handle_t;
+#else
+#error "runloop support not implemented on this platform"
+#endif
 
 #if TARGET_OS_MAC
 __OSX_AVAILABLE_STARTING(__MAC_10_6,__IPHONE_4_0)
 DISPATCH_EXPORT DISPATCH_CONST DISPATCH_WARN_RESULT DISPATCH_NOTHROW
-mach_port_t
+dispatch_runloop_handle_t
 _dispatch_get_main_queue_port_4CF(void);
+#endif
 
-__OSX_AVAILABLE_STARTING(__MAC_10_6,__IPHONE_4_0)
+__OSX_AVAILABLE(10.12) __IOS_AVAILABLE(10.0)
+__TVOS_AVAILABLE(10.0) __WATCHOS_AVAILABLE(3.0)
 DISPATCH_EXPORT DISPATCH_NOTHROW
-void
-_dispatch_main_queue_callback_4CF(mach_msg_header_t *_Null_unspecified msg);
-#elif TARGET_OS_WIN32
-__OSX_AVAILABLE_STARTING(__MAC_10_6,__IPHONE_4_0)
-DISPATCH_EXPORT DISPATCH_NOTHROW
-HANDLE
+dispatch_runloop_handle_t
 _dispatch_get_main_queue_handle_4CF(void);
 
 __OSX_AVAILABLE_STARTING(__MAC_10_6,__IPHONE_4_0)
 DISPATCH_EXPORT DISPATCH_NOTHROW
 void
-_dispatch_main_queue_callback_4CF(void);
-#endif // TARGET_OS_WIN32
+_dispatch_main_queue_callback_4CF(void *_Null_unspecified msg);
 
 __OSX_AVAILABLE_STARTING(__MAC_10_9,__IPHONE_7_0)
 DISPATCH_EXPORT DISPATCH_MALLOC DISPATCH_RETURNS_RETAINED DISPATCH_WARN_RESULT
--- a/src/queue.c
+++ b/src/queue.c
@@ -74,11 +74,11 @@ static int _dispatch_pthread_sigmask(int how, sigset_t *set, sigset_t *oset);
 #endif
 
 #if DISPATCH_COCOA_COMPAT
-static dispatch_once_t _dispatch_main_q_port_pred;
+static dispatch_once_t _dispatch_main_q_handle_pred;
 static void _dispatch_runloop_queue_poke(dispatch_queue_t dq,
 		pthread_priority_t pp, dispatch_wakeup_flags_t flags);
-static void _dispatch_runloop_queue_port_init(void *ctxt);
-static void _dispatch_runloop_queue_port_dispose(dispatch_queue_t dq);
+static void _dispatch_runloop_queue_handle_init(void *ctxt);
+static void _dispatch_runloop_queue_handle_dispose(dispatch_queue_t dq);
 #endif
 
 static void _dispatch_root_queues_init_once(void *context);
@@ -4063,6 +4063,48 @@ _dispatch_queue_wakeup(dispatch_queue_t dq, pthread_priority_t pp,
 }
 
 #if DISPATCH_COCOA_COMPAT
+
+DISPATCH_ALWAYS_INLINE
+static inline bool
+_dispatch_runloop_handle_is_valid(dispatch_runloop_handle_t handle)
+{
+#if TARGET_OS_MAC
+	return MACH_PORT_VALID(handle);
+#elif defined(__linux__)
+	return handle >= 0;
+#else
+#error "runloop support not implemented on this platform"
+#endif
+}
+
+DISPATCH_ALWAYS_INLINE
+static inline dispatch_runloop_handle_t
+_dispatch_runloop_queue_get_handle(dispatch_queue_t dq)
+{
+#if TARGET_OS_MAC
+	return ((dispatch_runloop_handle_t)(uintptr_t)dq->do_ctxt);
+#elif defined(__linux__)
+	// decode: 0 is a valid fd, so offset by 1 to distinguish from NULL
+	return ((dispatch_runloop_handle_t)(uintptr_t)dq->do_ctxt) - 1;
+#else
+#error "runloop support not implemented on this platform"
+#endif
+}
+
+DISPATCH_ALWAYS_INLINE
+static inline void
+_dispatch_runloop_queue_set_handle(dispatch_queue_t dq, dispatch_runloop_handle_t handle)
+{
+#if TARGET_OS_MAC
+	dq->do_ctxt = (void *)(uintptr_t)handle;
+#elif defined(__linux__)
+	// encode: 0 is a valid fd, so offset by 1 to distinguish from NULL
+	dq->do_ctxt = (void *)(uintptr_t)(handle + 1);
+#else
+#error "runloop support not implemented on this platform"
+#endif
+}
+
 void
 _dispatch_runloop_queue_wakeup(dispatch_queue_t dq, pthread_priority_t pp,
 		dispatch_wakeup_flags_t flags)
@@ -4089,13 +4131,6 @@ _dispatch_runloop_queue_wakeup(dispatch_queue_t dq, pthread_priority_t pp,
 	if (flags & DISPATCH_WAKEUP_CONSUME) {
 		return _dispatch_release_tailcall(dq);
 	}
-}
-#else
-void
-_dispatch_runloop_queue_wakeup(dispatch_queue_t dq, pthread_priority_t pp,
-		dispatch_wakeup_flags_t flags)
-{
-	LINUX_PORT_ERROR();
 }
 #endif
 
@@ -4130,10 +4165,13 @@ _dispatch_root_queue_wakeup(dispatch_queue_t dq,
 static inline void
 _dispatch_runloop_queue_class_poke(dispatch_queue_t dq)
 {
-	mach_port_t mp = (mach_port_t)dq->do_ctxt;
-	if (!mp) {
+	dispatch_runloop_handle_t handle = _dispatch_runloop_queue_get_handle(dq);
+	if (!_dispatch_runloop_handle_is_valid(handle)) {
 		return;
 	}
+
+#if TARGET_OS_MAC
+	mach_port_t mp = handle;
 	kern_return_t kr = _dispatch_send_wakeup_runloop_thread(mp, 0);
 	switch (kr) {
 	case MACH_SEND_TIMEOUT:
@@ -4144,6 +4182,15 @@ _dispatch_runloop_queue_class_poke(dispatch_queue_t dq)
 		(void)dispatch_assume_zero(kr);
 		break;
 	}
+#elif defined(__linux__)
+	int result;
+	do {
+		result = eventfd_write(handle, 1);
+	} while (result == -1 && errno == EINTR);
+	(void)dispatch_assume_zero(result);
+#else
+#error "runloop support not implemented on this platform"
+#endif
 }
 
 DISPATCH_NOINLINE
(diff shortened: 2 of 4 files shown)


src/queue.c
```diff
@@ -4144,6 +4142,17 @@ _dispatch_runloop_queue_class_poke(dispatch_queue_t dq)
 		(void)dispatch_assume_zero(kr);
 		break;
 	}
+#elif defined(__linux__)
+	int efd = (int)dq->do_ctxt;
+
+	if (efd != 0) {
```
MadCoder: 0 is a valid fd, and you will definitely get it if you close STDIN which lots of daemons do, which makes it a bit of an issue. A very gross way to fix that is to store (eventfd + 1) in here

src/queue.c
```diff
@@ -5629,13 +5642,21 @@ _dispatch_runloop_queue_port_init(void *ctxt)
 		(void)dispatch_assume_zero(kr);
 	}
 	dq->do_ctxt = (void*)(uintptr_t)mp;
+#elif defined(__linux__)
+	int efd = eventfd(0, EFD_CLOEXEC | EFD_NONBLOCK);
+	(void)dispatch_assume(efd != 0);
```
MadCoder: that's not the right test, the right test is `efd >= 0` see remark above.

src/queue.c
```diff
@@ -5647,19 +5668,39 @@ _dispatch_runloop_queue_port_dispose(dispatch_queue_t dq)
 	kr = mach_port_mod_refs(mach_task_self(), mp, MACH_PORT_RIGHT_RECEIVE, -1);
 	DISPATCH_VERIFY_MIG(kr);
 	(void)dispatch_assume_zero(kr);
+#elif defined(__linux__)
+	int efd = (int)dq->do_ctxt;
+	if (!efd) {
```
MadCoder: same remark as above.

src/queue.c
```diff
@@ -5673,7 +5714,11 @@ _dispatch_queue_set_mainq_drain_state(bool arg)
 }
 
 void
+#if TARGET_OS_MAC
 _dispatch_main_queue_callback_4CF(mach_msg_header_t *msg DISPATCH_UNUSED)
+#else
+_dispatch_main_queue_callback_4CF()
```
MadCoder: `(void)`

MadCoder:
I don't think we care much about windows, the port is clearly not working, and I would just get rid of that.

maybe the right thing to do is to have a `dispatch_runloop_handle_t` rather than `dispatch_handle_t` which is a bit too generic that abstracts port & fd on linux.

and internally I would have a:

```C
DISPATCH_ALWAYS_INLINE
static inline bool
_dispatch_runloop_handle_is_valid(dispatch_runloop_handle_t handle)
{
#if TARGET_OS_MAC
    return MACH_PORT_VALID(handle);
#elif defined(__linux__)
    return handle >= 0
#else
#error
#endif
}

DISPATCH_ALWAYS_INLINE
static inline dispatch_runloop_handle_t
_dispatch_runloop_queue_get_handle(dispatch_queue_t dq)
{
#if TARGET_OS_MAC
    return ((dispatch_runloop_handle_t)(uintptr_t)dq->dq_ctxt);
#elif defined(__linux__)
    return ((dispatch_runloop_handle_t)(uintptr_t)dq->dq_ctxt) - 1;
#else
#error
#endif
}

DISPATCH_ALWAYS_INLINE
static inline void
_dispatch_runloop_queue_set_handle(dispatch_queue_t dq, dispatch_runloop_handle_t handle)
{
#if TARGET_OS_MAC
    return dq->do_ctxt = (void *)(uintptr_t)handle;
#elif defined(__linux__)
    return dq->do_ctxt = (void *)(uintptr_t)(handle + 1);
#else
#error
#endif
}
```

That way the actual mangling of the handle is abstracted away and should allow you to write clean code whether it's a mach port, a file descriptor, a HANDLE, or maybe even something else in the future.

I think you need the +1/-1 trick for fd's because do_ctxt is initialized to 0 where you would want it to be -1, and that is easier to do that mangling above than to try to initialize runloop queues differently.

do you agree?

src/queue.c
```diff
@@ -23,6 +23,10 @@
 #include "protocol.h"
 #endif
 
+#if defined(__linux__)
```
MadCoder: this kind of `#include` should be in `src/internal.h` with other dependencies that dispatch needs for most of its source.

private/private.h
```diff
@@ -166,10 +166,22 @@ void _dispatch_prohibit_transition_to_multithreaded(bool prohibit);
...
+#elif defined(__linux__)
+#define DISPATCH_COCOA_COMPAT 1
+#else
+#define DISPATCH_COCOA_COMPAT 0
+#endif
 
 #if DISPATCH_COCOA_COMPAT
 
+#if TARGET_OS_WIN32
+typedef HANDLE dispatch_handle_t;
```
MadCoder: see my comment on the general review, I like abstracting the type, but `dispatch_handle_t` is too generic for something that really is about runloops only. I suggested `dispatch_runloop_handle_t`. FWIW I'm not necessarily in love with `handle` as a term, `port` definitely feels wrong. Maybe @das will have a good idea, he's generally good at naming.

I would make it a typedef to mach_port_t on TARGET_OS_MAC too for consistency sake, it will be binary compatible.

dgrove-oss:
Thanks for the quick review! 
The suggestion for dispatch_runloop_handle_t and setters/getters makes sense to me.  I'll update the patch later tonight.

parkera:
Not a blocker for this, but we will care about Windows in the long run - both because we actually do use dispatch on windows today, and also because Swift will get ported there soon enough (if it hasn't quite made it already).

MadCoder:
@parkera I completely agree, what I meant is that @dgrove-oss was trying to maintain code that is not wroking at all. the `_get_handle()` function is actually a prototype only and has no corresponding function in `queue.c`, and in my experience, it's best to delete half-assed code so that porters aren't misled into thinking this may work.

I wasn't meaning that I don't care about windows, HTH.

parkera:
Ok, sounds good.

private/private.h
```diff
@@ -166,31 +166,45 @@ void _dispatch_prohibit_transition_to_multithreaded(bool prohibit);
...
+#endif
 
 #if DISPATCH_COCOA_COMPAT
 
 #if TARGET_OS_MAC
+typedef mach_port_t dispatch_runloop_handle_t;
+#elif defined(__linux__)
+typedef int dispatch_runloop_handle_t;
+#else
+#error
```
MadCoder: `#error` should have an argument, my example code was just the layout, but an actual error string is good :)

src/queue.c
```diff
@@ -4063,6 +4063,46 @@ _dispatch_queue_wakeup(dispatch_queue_t dq, pthread_priority_t pp,
...
+DISPATCH_ALWAYS_INLINE
+static inline bool
+_dispatch_runloop_handle_is_valid(dispatch_runloop_handle_t handle)
+{
+#if TARGET_OS_MAC
+	return MACH_PORT_VALID(handle);
+#elif defined(__linux__)
+	return handle >= 0;
+#else
+#error
```
MadCoder: ditto, and for the 2 other too

MadCoder:
ok from a cursory glance this looks clean and nice, but I'm too fried to do a proper review now.
I'd also like @das to look at it

dgrove-oss:
Squashed to a single commit.  

Confirmed that combined with the changes in https://github.com/apple/swift-corelibs-foundation/pull/434 plus the temporary hack in #100, Foundation links with Dispatch support on Linux and testOperationCount in TestNSOperationQueue passes (however testOperationPriorities hangs).  Did not investigate hang since I don't know if the Foundation side of the support is there for that test case yet.  

Would be nice to enable some of the dispatch tests that require CoreFoundation.   Would it make sense to try to reach into swift-corelibs-foundation from the dispatch build to find CoreFoundation to attempt to build them?  Or do you want to defer to testing in the context of swift-corelibs-foundation?

parkera:
I really want CF to remain a totally private implementation detail of Foundation.

parkera:
Would it be possible to make these tests use Foundation instead? I guess that would require reimplementing them.

das:
i'll take a look asap

MadCoder:
LGTM, @das to rubber stamp too as it affects headers

private/private.h
```diff
@@ -166,31 +166,45 @@ void _dispatch_prohibit_transition_to_multithreaded(bool prohibit);
  * SPI for CoreFoundation/Foundation ONLY
  */
 
-#define DISPATCH_COCOA_COMPAT (TARGET_OS_MAC || TARGET_OS_WIN32)
+#if (TARGET_OS_MAC || TARGET_OS_WIN32)
```
das: I think we should remove the WIN32 here, it is just going to #error below anyway

src/queue.c
```diff
@@ -5594,71 +5638,90 @@ _dispatch_runloop_root_queue_wakeup_4CF(dispatch_queue_t dq)
...
 		(void)dispatch_assume_zero(kr);
 	}
-	dq->do_ctxt = (void*)(uintptr_t)mp;
+#elif defined(__linux__)
+	handle = eventfd(0, EFD_CLOEXEC | EFD_NONBLOCK);
+#else
+#error "runloop support not implemented on this platform"
+#endif	
+	(void)dispatch_assume(_dispatch_runloop_handle_is_valid(handle));
+	_dispatch_runloop_queue_set_handle(dq, handle);
```
das: do you really want to set the handle if it is invalid ? if seems like this wants to be inside an if for the assume above ?

alternatively add a hard failure (CRASH macro) here, similarly to what we have in `_dispatch_kq_init()` when we cannot create the kq file descriptor (e.g. due the process or system being out of fd's)

src/queue.c
```diff
@@ -4063,6 +4063,46 @@ _dispatch_queue_wakeup(dispatch_queue_t dq, pthread_priority_t pp,
...
+}
+
+DISPATCH_ALWAYS_INLINE
+static inline void
+_dispatch_runloop_queue_set_handle(dispatch_queue_t dq, dispatch_runloop_handle_t handle)
+{
+#if TARGET_OS_MAC
+	dq->do_ctxt = (void *)(uintptr_t)handle;
+#elif defined(__linux__)
+	dq->do_ctxt = (void *)(uintptr_t)(handle + 1);
```
das: a one-line comment about why this is here would probably be helpful

private/private.h
```diff
@@ -166,31 +166,45 @@ void _dispatch_prohibit_transition_to_multithreaded(bool prohibit);
...
-mach_port_t
+dispatch_runloop_handle_t
 _dispatch_get_main_queue_port_4CF(void);
 
 __OSX_AVAILABLE_STARTING(__MAC_10_6,__IPHONE_4_0)
 DISPATCH_EXPORT DISPATCH_NOTHROW
 void
 _dispatch_main_queue_callback_4CF(mach_msg_header_t *_Null_unspecified msg);
-#elif TARGET_OS_WIN32
+#else
```
das: maybe we should define these on TARGET_OS_MAC as well and move the opensource CF to call the new entrypoint on all platforms (the callback argument could be void*)

src/queue.c
```diff
@@ -5673,7 +5736,11 @@ _dispatch_queue_set_mainq_drain_state(bool arg)
 }
 
 void
+#if TARGET_OS_MAC
 _dispatch_main_queue_callback_4CF(mach_msg_header_t *msg DISPATCH_UNUSED)
+#else
+_dispatch_main_queue_callback_4CF(void)
```
das: this seems like an unnecessary difference if we made the argument void* and just pass in NULL (given that it is not used)

src/queue.c
```diff
@@ -5594,71 +5638,90 @@ _dispatch_runloop_root_queue_wakeup_4CF(dispatch_queue_t dq)
...
 
 #pragma mark -
 #pragma mark dispatch_main_queue
 
-mach_port_t
+dispatch_runloop_handle_t
+#if TARGET_OS_MAC
 _dispatch_get_main_queue_port_4CF(void)
+#else
+_dispatch_get_main_queue_handle_4CF(void)
```
das: as in the header comment above, IMO this could be the primary version and `_dispatch_get_main_queue_port_4CF` just a legacy alias

src/queue.c
```diff
@@ -5594,71 +5638,90 @@ _dispatch_runloop_root_queue_wakeup_4CF(dispatch_queue_t dq)
...
 	dispatch_queue_t dq = (dispatch_queue_t)ctxt;
-	mach_port_t mp;
-	kern_return_t kr;
+	dispatch_runloop_handle_t handle;
 
 	_dispatch_fork_becomes_unsafe();
-	kr = mach_port_allocate(mach_task_self(), MACH_PORT_RIGHT_RECEIVE, &mp);
+
+#if TARGET_OS_MAC
+	kern_return_t kr;
```
das: keeping the `mach_port_t mp` local in this `#if` and only setting the generic handle var at the end of the block would read more cleanly to me (there is a lot of other code dealing with mp & mach_port_t in this project that this now clashes with style-wise)

src/queue.c
```diff
@@ -5594,71 +5638,90 @@ _dispatch_runloop_root_queue_wakeup_4CF(dispatch_queue_t dq)
...
+	if (!_dispatch_runloop_handle_is_valid(handle)) {
 		return;
 	}
 	dq->do_ctxt = NULL;
-	kern_return_t kr = mach_port_deallocate(mach_task_self(), mp);
+#if TARGET_OS_MAC
+	kern_return_t kr = mach_port_deallocate(mach_task_self(), hand);
 	DISPATCH_VERIFY_MIG(kr);
 	(void)dispatch_assume_zero(kr);
 	kr = mach_port_mod_refs(mach_task_self(), mp, MACH_PORT_RIGHT_RECEIVE, -1);
```
das: this still uses `mp` and will not compile as is. As above, I think this should keep a `mach_port_t mp` local inside this `#if` for symmetry with other such code (and to avoid mistakes like this)

das:
one bug and various small style/convention comments but lgtm otherwise, back to you

private/private.h
MadCoder: @das CF already know that windows prototype takes `(void)` so I'm not sure we can really solve this easily. I would keep it that way I think.

das: I don't see why, passing in an extra unused NULL isn't harmful (i.e. there is no binary compatibility issue) and we don't expect to build CF against older versions of dispatch I assume

(the standard way to avoid the revlock Tony is referring to is to bump DISPATCH_API_VERSION, or add a new macro of that type, and conditionally compile in CF based on that)

das: (though I'm certainly also ok with cleaning this part up in a later set of changes if we think it is too painful to deal with right now, it just seemed like the right opportunity given that CF is not yet making those calls on Linux)

parkera: Right now I'm interested in minimum required changes so we can move forward. =)

das: @MadCoder we don't build CF on windows against these headers, so I don't think that matters

dgrove-oss:
I'd already made changes to private/private.h based on my interpretation of @das comments.  I can undo or do differently, just let me know what you'd like.  Please be sure to look at the availability macro I put on _dispatch_get_main_queue_handle_4CF.  Since this is a new function on MAC_OS, I think it should get these versions, but I'm only guessing.

Once we finalize this, I can update my pull request to foundation to match the final API.

parkera:
Ok, we can deal with any lockstep changes later probably.

das:
Thanks @dgrove-oss that looks exactly like what I had in mind. The new availability annotation should be `__OSX_AVAILABLE(10.12) __IOS_AVAILABLE(10.0) __TVOS_AVAILABLE(10.0) __WATCHOS_AVAILABLE(3.0)` I think
Given the above discussion, it would probably also be useful to add a `#define DISPATCH_CF_SPI_VERSION 20160712` define above this section, that CF can use to switch to the new call without revlock.

private/private.h
```diff
@@ -166,31 +166,40 @@ void _dispatch_prohibit_transition_to_multithreaded(bool prohibit);
...
 _dispatch_get_main_queue_port_4CF(void);
+#endif
 
-__OSX_AVAILABLE_STARTING(__MAC_10_6,__IPHONE_4_0)
-DISPATCH_EXPORT DISPATCH_NOTHROW
-void
-_dispatch_main_queue_callback_4CF(mach_msg_header_t *_Null_unspecified msg);
-#elif TARGET_OS_WIN32
-__OSX_AVAILABLE_STARTING(__MAC_10_6,__IPHONE_4_0)
+__OSX_AVAILABLE_STARTING(__MAC_10_12,__IPHONE_10_0)
```
das:
```
__OSX_AVAILABLE(10.12) __IOS_AVAILABLE(10.0)
__TVOS_AVAILABLE(10.0) __WATCHOS_AVAILABLE(3.0)
```

dgrove-oss:
availability macros updated and DISPATCH_CF_SPI_VERSION #define added.

das:
perfect, thanks

PR #102, opened 2016-07-11
add dispatchMain to wrapping overlay

dgrove-oss:
dispatch_main is meant to be made available via
the global function dispatchMain().


MadCoder:
hah that could be useful.

@mwwa is it what we're doing too on Darwin?

PR #104, opened 2016-07-13
Change DispatchQueue.after() to take a DispatchTimeInterval

karwa:
- Change DispatchQueue.after() to take a DispatchTimeInterval instead of an absolute time
- Rename DispatchQueue.after() that takes an absolute time to DispatchQueue.at(time:)

I couldn't find a way to build/test these changes on OSX. All I know is that this API really bugs me. Especially that DispatchQueue.after() takes an absolute time. That has always bugged me, even with the C API.

I would much rather write `DispatchQueue.main.after(.seconds(1)) { /* some stuff */ }`

### DIFF
--- a/src/swift/Queue.swift
+++ b/src/swift/Queue.swift
@@ -302,32 +302,34 @@ public extension DispatchQueue {
 		}
 	}
 
-	public func after(when: DispatchTime, qos: DispatchQoS = .unspecified, flags: DispatchWorkItemFlags = [], execute work: @convention(block) () -> Void) {
+	public func after(_ duration: DispatchTimeInterval, from: DispatchTime = DispatchTime.now(), qos: DispatchQoS = .unspecified, flags: DispatchWorkItemFlags = [], execute work: @convention(block) () -> Void) {
+		let executeTime = from + duration
 		if #available(OSX 10.10, iOS 8.0, *), qos != .unspecified || !flags.isEmpty {
 			let item = DispatchWorkItem(qos: qos, flags: flags, block: work)
-			dispatch_after(when.rawValue, self.__wrapped, item._block)
+			dispatch_after(executeTime.rawValue, self.__wrapped, item._block)
 		} else {
-			dispatch_after(when.rawValue, self.__wrapped, work)
+			dispatch_after(executeTime.rawValue, self.__wrapped, work)
 		}
 	}
 
 	@available(OSX 10.10, iOS 8.0, *)
-	public func after(when: DispatchTime, execute: DispatchWorkItem) {
-		dispatch_after(when.rawValue, self.__wrapped, execute._block)
+	public func after(_ duration: DispatchTimeInterval, from: DispatchTime = DispatchTime.now(), execute: DispatchWorkItem) {
+		let executeTime = from + duration
+		dispatch_after(executeTime.rawValue, self.__wrapped, execute._block)
 	}
 
-	public func after(walltime when: DispatchWallTime, qos: DispatchQoS = .unspecified, flags: DispatchWorkItemFlags = [], execute work: @convention(block) () -> Void) {
+	public func at(_ time: DispatchWallTime, qos: DispatchQoS = .unspecified, flags: DispatchWorkItemFlags = [], execute work: @convention(block) () -> Void) {
 		if #available(OSX 10.10, iOS 8.0, *), qos != .unspecified || !flags.isEmpty {
 			let item = DispatchWorkItem(qos: qos, flags: flags, block: work)
-			dispatch_after(when.rawValue, self.__wrapped, item._block)
+			dispatch_after(time.rawValue, self.__wrapped, item._block)
 		} else {
-			dispatch_after(when.rawValue, self.__wrapped, work)
+			dispatch_after(time.rawValue, self.__wrapped, work)
 		}
 	}
 
 	@available(OSX 10.10, iOS 8.0, *)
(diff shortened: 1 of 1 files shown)


das:
this should not take a time interval, the deadline is specific to the chosen clock, TimeInterval does not contain that information.

karwa:
@das how about:
- Adding a clock parameter, with a default, or
- Adding a convenience function

instead?

MadCoder:
What is the advantage from

```Swift
DispatchQueue.main.after(when: .now() + .seconds(1))
DispatchQueue.main.after(walltime: .now() + .seconds(1))
```

the number of keystrokes isn't significantly more and it is properly typed

karwa:
@MadCoder It would be:
```
DispatchQueue.main.after(.seconds(1))
DispatchQueue.main.at(time: .now() + .seconds(1))
```

It bugs me because on the relatively infrequent occasions that I need this API, it always involves a trawl through the interfaces, cross-checking types and finding out how to construct something.

The words don't even make grammatical sense: "after when" -- which is it; an "after" (i.e. relative time) or a "when" (i.e. absolute time)? If you look at the level of nuance at which they're picking apart the standard library over on swift-evo, I'm pretty sure the "swifty" way is to disambiguate these cases.

mwwa:
I agree, I don't see a huge advantage over the changes we were already thinking about:
```
func asyncAfter(deadline:qos:flags:work:)
func asyncAfter(wallDeadline:qos:flags:work:)
```

mwwa:
Most of these examples can even be shortened to:
```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 1.0) { /* block */ }
```

karwa:
@mwwa Even if it was renamed to `asyncAfter`, it should still take an interval rather than an absolute time. If you wanted a looser guarantee about timing, it should be:

```
DispatchQueue.main.by(dealine: ...)
```

Or `asyncBy(deadline:)`, or some other variant which is more descriptive about what it's doing. But for as long as it's `after`, IMO it should take a duration (and optionally a point in time that it's relative to).

karwa:
Also, `wallTime` is not necessary. I would refer everybody to the [Swift API Design Guidelines](https://swift.org/documentation/api-design-guidelines/). Specifically points:

- **Take advantage of defaulted parameters when it simplifies common uses**. Any parameter with a single commonly-used value is a candidate for a default.
- **Omit needless words**. Every word in a name should convey salient information at the use site. More words may be needed to clarify intent or disambiguate meaning, but those that are redundant with information the reader already possesses should be omitted. In particular, omit words that merely repeat type information.
- **Name variables, parameters, and associated types according to their roles**, rather than their type constraints.
- **Prefer method and function names that make use sites form grammatical English phrases.**
- **Clarity at the point of use is your most important goal**. Entities such as methods and properties are declared only once but used repeatedly. Design APIs to make those uses clear and concise. When evaluating a design, reading a declaration is seldom sufficient; always examine a use case to make sure it looks clear in context.

MadCoder:
> Also, wallTime is not necessary

Yes it is, again, the `DispatchTimeInterval` type doesn't express which clock is used and this cannot work. There is a difference between `deadline` (monotonic clock) and `wallDeadline` which is calendar clock. One takes system sleep into account and not the other, these are quite important differences.

We know the swift guidelines alright, your proposal fails the first test: there's no good default clock to use here, so you can't omit the deadline / wallDeadline.

karwa:
Well I continue to disagree. I'm pretty sure that if it actually went through a swift-evo discussion this entire new Dispatch API would get torn apart. It doesn't fit Swift at all - it's full of direct translations from C, even when you literally feel like you're speaking gobbledegook (such as "after when").

And no, `wallTime` really is not necessary.  Look at the 2nd bullet point. `wallTime` literally just repeats type information. Sure, there is a distinction between the wall and monotonic clocks, but you actually made that distinction at the type level - look at `src/swift/Time.swift`, literally at the very start of the file. You don't need to carry it around with you every time you describe it - you've moved above C now; we know that's a wall time by virtue of the fact that it's declared as taking a type `DispatchWallTime`. Do you understand me yet?

Also, I believe the wallclock issue is a perfect example of one where there is an obvious default. Is it really such a very common situation that relative times I specify (likely to be on the duration of a handful of seconds) are likely to intersect a sleep event? It's the very definition of a edge-case, and translates in to cryptic and undiscoverable boilerplate that (I would seriously bet) >90% of GCD users don't understand or care about. I would argue that my approach, with an explicit default clock, is far more obvious and clearer at the point of use.

Are the swift corelibs under the domain of swift-evo? I have to apologise, but I disagree so strongly I'm not going to drop it. I have to use this API, it's open-source, so I'm going to insist on opening this discussion to a wider audience.

MadCoder:
dispatch is part of corelibs which is discussed on swift-evo and this proposal was. We have a few adaptations to make for obvious actual bugs in the interface (around DispatchWorkItems), but a pull request is indeed not the right forum to discuss an API surface change.

karwa:
@MadCoder Okay I've posted an evo thread about it. While writing it down, the splitting of the clocks at the type level started to feel very strange and potentially error-prone. I added a bit about possibly unifying them and adding a bool/enum field for the clock (embedded types make this quite elegant IMO). It's also a bit more like the C API in that way (which just has a single `dispatch_time_t`).

MadCoder:
we care about distinct types. how `dispatch_time_t` was built in C is considered a mistake because it makes the type not `Comparable`. We don't want to repeat that mistake

PR #109, opened 2016-07-19
Add libkqueue as a git submodule

seabaylea:
We added libkqueue as a git submodule to `experimental/foundation` under PR #76 to pick up a fix that resolves a crash when utilising Dispatch in NSURLSession.

We still have the same issue in the NSURLSession test, so we need to add the git submodule to `master` as well.


configure.ac
```diff
@@ -183,12 +183,20 @@ esac
...
+AS_IF([test -f $srcdir/libkqueue/configure.ac],
+  [AC_DEFINE(BUILD_OWN_KQUEUES, 1, [Define if building libkqueue from source])
+   ac_configure_args="--disable-libkqueue-install $ac_configure_args"
+   AC_CONFIG_SUBDIRS([libkqueue])
+   build_own_kqueues=true],
+  [build_own_kqueues=false
+   AC_CHECK_HEADER(sys/event.h, [],
+     [AC_SEARCH_LIBS([kqueue], [kqueue], [],
+       [AC_MSG_ERROR([unable to find kqueue() function])]
+     )]
```
dgrove-oss: Why the change from PKG_CHECKS_MODULES to AC_SEARCH_LIBS?  PKG_CHECK_MODULES will define KQUEUE_CFLAGS and KQUEUE_LIBS so you don't have to hardwire them in the makefile

src/Makefile.am
```diff
@@ -60,12 +60,19 @@ EXTRA_libdispatch_la_DEPENDENCIES=
...
-AM_CXXFLAGS=$(PTHREAD_WORKQUEUE_CFLAGS) $(DISPATCH_CFLAGS) $(CXXBLOCKS_FLAGS)
+AM_CXXFLAGS=$(KQUEUE_CFLAGS) $(PTHREAD_WORKQUEUE_CFLAGS) $(DISPATCH_CFLAGS) $(CXXBLOCKS_FLAGS)
 AM_OBJCXXFLAGS=$(DISPATCH_CFLAGS) $(CXXBLOCKS_FLAGS)
 
+if BUILD_OWN_KQUEUES
+  KQUEUE_LIBS=$(top_builddir)/libkqueue/libkqueue.la
+  KQUEUE_CFLAGS=-I$(top_srcdir)/libkqueue/include
+else
+  KQUEUE_CFLAGS=-I/usr/local/include/kqueue
+endif
```
dgrove-oss: can't assume /usr/local/include/kqueue is the right place to get them.

configure.ac
ianpartridge: @seabaylea is porting my changes, here. I'm trying to remember why I changed this. I think I couldn't get M4 to parse the PKG_CHECKS_MODULES correctly when under an AS_IF - and switched to AC_SEARCH_LIBS instead.

dgrove-oss: Not sure if it is accurate, but I did find this page describing how to use the two together: https://autotools.io/pkgconfig/pkg_check_modules.html

seabaylea:
I've switched over/back to using `PKG_CHECK_MODULES`,  which also resolves the `KQUEUE_CFLAGS` issue for non-submodule builds.

MadCoder:
sorry you lost that race and conflict, please update.

seabaylea:
@MadCoder Updated to resolve merge conflict.

PR #120, opened 2016-07-26
Update for SE-0107: Migrate Void->Raw.

atrick:
Unsafe[Mutable]Pointer<Void> is replaced by Unsafe[Mutable]RawPointer.

Migrating often means simply removing casts to/from UnsafePointer<UInt8>
and working directly with UnsafeRawPointer instead.

Unfortunately some API's, like NSData still want UnsafePointer<UInt8>,
which requires "binding" memory to UInt8 before passing it to the API.


atrick:
Depends on swift PR:
https://github.com/apple/swift/pull/3724

PR #121, opened 2016-07-27
[Data] Fixed crash when getting pointer to empty DispatchData

karwa:
- corrected comment for read-only subscript

### DIFF
--- a/src/swift/Data.swift
+++ b/src/swift/Data.swift
@@ -187,7 +187,7 @@ public struct DispatchData : RandomAccessCollection {
 		return copyRange.count
 	}
 
-	/// Sets or returns the byte at the specified index.
+	/// Returns the byte at the specified index.
 	public subscript(index: Index) -> UInt8 {
 		var offset = 0
 		let subdata = CDispatch.dispatch_data_copy_region(__wrapped.__wrapped, index, &offset)
--- a/src/swift/Data.swift
+++ b/src/swift/Data.swift
@@ -74,12 +74,13 @@ public struct DispatchData : RandomAccessCollection {
 	}
 
 	public func withUnsafeBytes<Result, ContentType>(
-		body: (UnsafePointer<ContentType>) throws -> Result) rethrows -> Result
+		body: (UnsafePointer<ContentType>?) throws -> Result) rethrows -> Result
 	{
 		var ptr: UnsafeRawPointer? = nil
 		var size = 0
 		let data = CDispatch.dispatch_data_create_map(__wrapped.__wrapped, &ptr, &size)
-		let contentPtr = ptr!.bindMemory(
+		assert(ptr != nil || self.isEmpty)
+		let contentPtr = ptr?.bindMemory(
 			to: ContentType.self, capacity: size / MemoryLayout<ContentType>.stride)
 		defer { _fixLifetime(data) }
 		return try body(contentPtr)


karwa:
I don't see a way around making the pointer in the handler optional. There's no sensible pointer to give if we don't have any data, and we still need to invoke the handler in order to pass-through the return value.

mwwa:
Given that I let this lie around, the branch needs updating for the UnsafeRawPointer changes of recent times.

karwa:
@mwwa rebased

karwa:
@mwwa / @MadCoder any word on this? A similar API change would be required for the SDK overlays (making the pointer optional)

karwa:
Bug still exists in Swift 3.0 :(

dgrove-oss:
@mwwa How do we progress this?

mwwa:
```withUnsafeBytes<ResultType, ContentType>``` mirrors ```Foundation.Data```'s signature. I suspect the approach needed here is to skip the block if the data is null.

**Comment by karwa** [comment]
If we do that, we'd need to change the result type to optional. One of them needs to change.

**Comment by MadCoder** [comment]
we discussed this at lunch with @mmwa, a way to solve this is to make `Dispatch.DataEmpty` a subclass of `Dispatch.Data` and have `withUnsafeBytes` on that subclass pass a non NULL pointer.

**Comment by karwa** [comment]
@MadCoder @mwwa A pointer to what, then? Swift's pointer model changed during Swift 3 so that non-nil pointers always point _to_ something.

I was just having a look at what `Foundation.Data` does, and they also violate this. An empty `Data` returns a nil pointer as a non-optional, leading to a crash with the following line:
```
Data().withUnsafeBytes { (ptr : UnsafePointer<UInt8>) in print("\(ptr.pointee)") }
```

So it's likely that both APIs will need to change. CC @phausler

**Comment by MadCoder** [comment]
a pointer to any memory that is valid: `""` or `self` or whatever. this can be achieved like I said with a very simple subclass part of the cluster to represent the empty Datums

**Comment by karwa** [comment]
That sounds quite ridiculous to me. Since the type is generic, users will be able to call `pointee` and get back a total junk object, possibly full of invalid pointers, and will have to manually check if the data is empty (Dispatch)Data in order to know if the object is valid or not. That flies in the face of Swift standard practice (nullable pointers should be optional) and policy (safe-by-default), and I don't think is acceptable to anybody.

We already have a mechanism for this - the pointer in the closure should be an optional. It makes semantic sense and all of these issues just go away:

```
// ptr is optional, so does not crash
Data().withUnsafeBytes { (ptr : UnsafePointer<UInt8>?) in print("\(ptr?.pointee)") }
```

In the example above, if the pointer is not optional, what value do you propose `ptr.pointee` to return? Baring in mind that it must be a valid value for every type `T`?

**Comment by karwa** [comment]
Also, an additional question as I was working with these types: `dispatch_data_t` is toll-free bridged to `NSData` since iOS9; NSData gained `enumerateBytes` functions to deal with non-contiguously-backed memory because of it, and corelibs-foundation's `Data` value-type retains those functions and mentions possible backing by a `dispatch_data_t` -- but AFAICT there is no way to actually exploit this from Swift: there is no initialiser for `Data` which takes a `DispatchData`, and `dispatch_data_t` is not available from Swift.

Is this something which should be added? Or is it no longer supported?

**Comment by MadCoder** [comment]
that's not junk as you're allowed to only read `0` bytes from that pointer. any pointer value in your address space will do. I don't see a problem with that, it's not even UB in C:

```C
memcpy(NULL, NULL, 0); // UB
memcpy("", "", 0); // fine even if gross as "" is supposed to be in __TEXT,__cstring which is RO memory
```

karwa:
@MadCoder yes, but what if I have code like the following?

```
class MyClass {}
struct MyStruct {
    let aValue : MyClass
}
// Crash: MyStruct.aValue is nil but not optional
/*(Dispatch)*/Data().withUnsafeBytes { (ptr : UnsafePointer<MyStruct>) in print("\(ptr.pointee.aValue)") }
```

UnsafePointer<T>, non-optional, allows you to just plain deference the pointer. That's safe because potentially-nil pointers are now optional in Swift 3, so a non-optional UnsafePointer<T> always points to a valid T (within the boundaries of `Unsafe`). Since we can't possibly provide a valid T without any data, I don't see how it can't be optional.

Looking at the Swift standard library, Array<T> has a sort-of similar functionality, but it uses UnsafeManagedBufferPointer instead; UMBP's collection interface allows it to be empty, though, whereas UnsafePointer can't. I think that's right -- the Swift pointer types are a bit difficult to figure out precisely. Maybe best to ask @dabrahams or another standard-library person what the correct, Swiftiest thing to do here is?

dabrahams:
@karwa I'm not really clear on what the question is or what you're trying to accomplish here.

@MadCoder we knowingly gave up the ability to treat a null pointer as a valid zero-length array in Swift 3, even though it's a valid C idiom.

MadCoder:
@dabrahams that was exactly my expectation and I don't think it's even valid in the standard but let's not rathole into that discussion.

The discussion here is that the empty DispatchData returns `{ NULL, 0 }` for its pointer & lengths, and I'm just saying that the easy fix here is to return any non NULL pointer that is in your address space instead of `NULL`, and he pretends it would break swift which I don't understand at all, it would seem surprising swift's unsafe buffers should support length 0 with just about any valid pointer in your address space since that pointer would never be dereferenced. am I correct in my expectations?

dabrahams:
@MadCoder I agree that is an appropriate fix, but it needs to be a properly aligned pointer for the pointee type, which might not be so easy to come up with.

dabrahams:
@MadCoder to clarify, the new pointer rules are intentionally written to allow dynamic checking of things you'd never look at in a release build by a sanitizer akin to Address Sanitizer.  Forming an `UnsafePointer<T>` that isn't aligned for `T` is one of those things.

MadCoder:
@dabrahams makes sense, using the `__dso_header` will be 4k aligned at least which is enough for everybody, and is readable so should fly with sanitizers. here is our answer then.

dabrahams:
@MadCoder I 100% agree. @atrick what do you think about exposing this as a static property `Unsafe[Mutable]BufferPointer<T>.empty`?

atrick:
Unsafe[Mutable]BufferPointer<T>.empty makes perfect sense to me.

karwa:
@MadCoder I didn't say it would break Swift, I said it's strange that you can just create an empty data an dereference a pointer from it to some random data. It seems to go against what I think the pointer model is there to achieve, and it just seems strange in general to me to return { something, 0 }; but I don't pretend to be an expert.

It doesn't seem like anybody else is in favour of making the pointer optional though, so I don't mind.

dabrahams:
@atrick care to write up the proposal? 😉

atrick:
Ok @dabrahams. It's on me now: https://bugs.swift.org/browse/SR-2976.
I don't want to send API proposals until we can version-gate source-breaking changes.

dabrahams:
on Mon Oct 17 2016, Andrew Trick <notifications-AT-github.com> wrote:

> Ok @dabrahams. It's on me now: https://bugs.swift.org/browse/SR-2976.
> I don't want to send API proposals until we can version-gate
> source-breaking changes.

#if SWIFT_VERSION_4  will work for now, won't it?

-- 
-Dave

atrick:
@dabrahams my concern was that we don't have a build-bot building with that -D flag. However, @graydon confirmed that I should be able to start using `#if swift(>=4)` and test cases built with -swift-version=4 will be able to see those symbols in our current stdlib build.

karwa:
This would be an additive change to the standard library and fix known crashing bugs in Foundation and Dispatch. Why does it need version-gating?

atrick:
Sorry to hijack this discussion. I don't know exactly what the proposed fix is now, since it has diverged from the PR. If it can be fixed without changing API signatures, then source compatibility is obviously moot. Or if there are crashing bugs, and no correct way to write the code against the current API, then it should just be fixed in Swift 3.

The proposed general UnsafeBufferPointer.empty API in SR-2976 is not necessary to fix the bug addressed by this PR. It does need to be gated under Swift 4 because any program using this additive API's is no longer compatible with Swift 3.

jrose-apple:
I'm against this whole idea. Neither Data nor DispatchData should have been vending an UnsafePointer to begin with; UnsafeBufferPointer is a much more appropriate type for a borrowed range of raw data. Given that we *did* release the old non-optional API, we should have those vend dummy pointers as was suggested, but it's not something that belongs in the standard library, and especially not on UnsafeBufferPointer, because it's not the right answer for any new API.

phausler:
vending a "dummy pointer" would definitely cause a mismatch between Data and NSData since the objective-c variant cannot vend out that pointer.

jrose-apple:
Well, it certainly *could,* but I don't think it actually matters. Data and NSData do not promise to vend the same raw pointer from their contents. Heck, there isn't a correct way to keep pointers vended from Data alive; that's why it uses the `with…` form.

karwa:
Looks like we already have a way to get a dummy pointer in the standard library (is this old code? Trying to find where any of this is used, just leads to dead-ends). If it's still useful, maybe we can reuse this API somehow: https://github.com/apple/swift/blob/master/stdlib/public/core/Pointer.swift#L49

That said I'd +1 deprecating the API with `UnsafePointer` and replacing with `UnsafeBufferPointer`, for symmetry with the standard library collections.

dabrahams:
@jrose-apple When you write "I'm against this whole idea" after a long discussion where several ideas have been raised, it's a little hard to know what you mean.

karwa:
Filed a bug to track this: https://bugs.swift.org/browse/SR-3097

shahmishal:

PR #123, opened 2016-07-28
Update INSTALL instructions

dgrove-oss:
Add description for building libdispatch with Swift API
either as part of building a Swift toolchain from source
using utils/build-script or using a pre-built Swift toolchain.

Update Linux-specific instructions to try to avoid some
common problems people have been reporting.


parkera:
Thanks!

PR #124, opened 2016-07-29
fix issues preventing overlay from compiling due to SE-0025

jpsim:
Without this change, the overlay fails to compile when running Swift's `utils/build-script --libdispatch` as of apple/swift@805e06e, which landed earlier today.


MadCoder:
@mwwa do we need that in the compiler too for Darwin?

PR #126, opened 2016-07-30
[SE-0111] Update for removal of argument labels from function types

DougGregor:


MadCoder:
@mwwa do we need to backport?

PR #130, opened 2016-07-30
[SE-0103][noescape by default] Migration to @noescape by default feature.

gonzalolarralde:
[SE-0103][1] makes `@noescape` the default annotation for blocks, adding the `@escaping` annotation in return to specify the opposite behaviour.

This changeset adds the annotation as it was required by the compiler.

[1]: https://github.com/apple/swift-evolution/blob/master/proposals/0103-make-noescape-default.md


shahmishal:
@swift-ci Please test

PR #131, opened 2016-07-31
Fixed the tuple label error

rfdickerson:
Resolves the build errors:

```
swift-corelibs-libdispatch/src/swift/Wrapper.swift:94:17: error: tuple element cannot have two labels
                                  handler: (_ error: Int32) -> Void) {
```

---
This change is [Reviewable](https://reviewable.io/reviews/apple/swift-corelibs-libdispatch/131)

### DIFF
--- a/src/swift/Data.swift
+++ b/src/swift/Data.swift
@@ -86,13 +86,13 @@ public struct DispatchData : RandomAccessCollection {
 	}
 
 	public func enumerateBytes(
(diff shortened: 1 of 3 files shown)


MadCoder:
You're reverting #126 here which is my understanding was to adapt for SE-0111. Update your compiler.

rfdickerson:
Oh ok, sorry about that. I was on Swift 7/29, assumed that this change was helping with that.

PR #133, opened 2016-08-04
Update deliveries value atomically to avoid potential race condition

seabaylea:
The libdispatch CI has an intermittent failure in `dispatch_read2` with the following reported in the log file:
```
[BEGIN] dispatch_read deliveries
	Actual: 319
	Expected: 320
[FAIL] dispatch_read deliveries (dispatch_read2.c:386)
	dispatch_read2.c:386
```
it looks like this could be cause by a race condition on `deliveries`, so switching to use atomic updates.

### DIFF
--- a/tests/dispatch_read2.c
+++ b/tests/dispatch_read2.c
@@ -343,19 +343,20 @@ test_writes_reads_eagain(void) // rdar://problem/8333366
 		test_errno("pipe", errno, 0);
 		test_stop();
 	}
-	const size_t chunks = 320;
+	const int32_t chunks = 320;
 	const size_t siz_chunk = 32, siz = siz_chunk * chunks;
 
 	dispatch_queue_t q = dispatch_get_global_queue(0,0);
 	dispatch_group_t g = dispatch_group_create();
-	__block size_t siz_acc = 0, deliveries = 0;
+	volatile __block int32_t deliveries = 0;
+	__block size_t siz_acc = 0;
 	__block void (^b)(dispatch_data_t, int);
 	b = Block_copy(^(dispatch_data_t data, int err) {
 		if (err) {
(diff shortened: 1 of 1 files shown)


MadCoder:
no, that can't be the bug, dispatch_read serializes everything through a single dispatch queue, so there is serialization already. FWIW we also have that issue on OS X and I have never taken the time to track it down yet.

seabaylea:
Ok, so does that means that the blocks are executed in serial, not just the reads occurring in serial and enqueueing the blocks with the correct data, which can then be executed in parallel?

MadCoder:
it's more about async I/O in general to avoid wasting thraeds while the data is being fetched by the kernel. but as you can see `dispatch_read(*fd, siz, q, b);` the queue argument mean that the results will be asynced to `q` so everything is serialized

what I think happens for real is that we do 240 writes, but one of the read coalesces two writes in a single read, and that the test assumptions about how the kernel works are wrong.

seabaylea:
Ok - if that's the case, are you happy for me to disable the test so that we can get the CI running cleanly? 

Long term we could look at modifying the test to only check if the data size is correct if we think the number of deliveries might legitimately be smaller than expected.

MadCoder:
sure, FWIW it's just a hunch that the test is wrong, I have never taken the time to double check.

seabaylea:
Ok - created PR #135 to disable the test.

MadCoder:
so for this test, the thing is, it expects that because it does a ` 		usleep(10000);` between writes, reads won't coalesce. this is actually not always the case.

the equality test should be made a test_less_or_equal instead, it's not the important test anyway.

MadCoder:
closing as this hasn't made progress. as I said the right fix is to move to a `<=` assertion instead of `==`

PR #138, opened 2016-08-09
Added Toolchain section in INSTALL.md

amraboelela:


MadCoder:
if @dgrove-oss is ok with it I'll merge

INSTALL.md
```diff
@@ -121,33 +113,71 @@ clang and blocks support:
...
+    dash-dash
+
+    install-libdispatch
+    install-foundation
+    reconfigure
+    ```
+
+2. Run:
+
+    utils/build-toolchain local.swift
```
seabaylea: This should be wrapped in a code block

INSTALL.md
```diff
@@ -121,33 +113,71 @@ clang and blocks support:
...
-	./configure
-	make
+    ```
+    sh autogen.sh
+    ./configure
+    make
     make install
+    ```
+
+#### Building toolchain in Linux
```
seabaylea: I think it actually makes sense for this section to be moved up under the "Configuring and installing libdispatch (general comments)" section as it follows on from how to build libdispatch into an existing Swift toolchain.

It should be marked that this is Linux only - but it might be worth stating that's because Swift on macOS platforms uses the system installed libdispatch, so its not required.

seabaylea:
@amraboelela this is looking good - just a couple of minor comments.

dgrove-oss:
LGTM; thanks!

PR #139, opened 2016-08-10
SR-2309: embed BlocksRuntime in libdispatch to eliminate external dep…

dgrove-oss:
…endency

Eliminate the external dependency on libBlocksRuntime.so by
copying the blocks runtime code from swift-corelibs-foundation/closure
and building it into libdispatch.  Added a configure option to control
whether or not the blocks runtime is embedded; defaults to embedded on
Linux and not embedded on other platforms.  Since the blocks runtime is
embedded; we can take the link directive out of the module.modulemap file.

This change enables dispatch to be built and run on a system without
the libblocksruntime-dev package installed.

### DIFF
--- a/dispatch/module.modulemap
+++ b/dispatch/module.modulemap
@@ -2,7 +2,6 @@ module Dispatch {
 	requires blocks
 	export *
 	link "dispatch"
-	link "BlocksRuntime"
 }
 
 module DispatchIntrospection [system] [extern_c] {
@@ -16,5 +15,4 @@ module CDispatch [system] [extern_c] {
 	export *
 	requires blocks
 	link "dispatch"
-	link "BlocksRuntime"
 }
--- a/m4/blocks.m4
+++ b/m4/blocks.m4
@@ -11,6 +11,21 @@ AC_ARG_WITH([blocks-runtime],
 )
 
 #
+# Configure argument to enable/disable using an embedded blocks runtime
+#
+AC_ARG_ENABLE([embedded_blocks_runtime],
+  [AS_HELP_STRING([--enable_embedded_blocks_runtime],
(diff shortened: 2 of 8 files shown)


MadCoder:
Can you remove the mention to libblocksruntime-dev from INSTALL then please? or maybe have it removed from the ohter pull-request that already is changing it right now? whichever you prefer.

dgrove-oss:
yes.  I was going to update the list of dependencies after the other pull request landed.  I don't want to cause a conflict for @amraboelela.   We also should update to say that libpwq and libkqueue are optional.  I'll do both once the dust settles on INSTALL.md

MadCoder:
sounds reasonable

RLovelett:
Perhaps I do not understand the implications of this change but doesn't this change make [this un-necessary](https://github.com/apple/swift-corelibs-libdispatch/blob/4c588e919b7617f600ad4d07779d1c7c5043e3e9/m4/blocks.m4#L28-L42)?

dgrove-oss:
Hi Ryan,  sorry if I got the help-string wrong.  I'll take a look and open a PR to fix it.   On linux configure should enable the embedded blocks runtime by default (so you don't need the configure flag).

RLovelett:
It is no trouble @dgrove-oss. I merely wanted to bring it to your attention.

I realized that I did not actually need the flag after a few hours of playing around over the weekend. The reason I was exploring the flag in the first place was because I couldn't get it to build with the bundled blocks runtime.

After banging my head against a wall for a few hours I realized I was checking out a _fork_ of the repository that did not have these changes. Suffice it to say I had a long weekend. :hurtrealbad:

Thank you for the work you are contributing here.

PR #144, opened 2016-08-12
fixes for compilation/linking problems on linux

dgrove-oss:
The Makefile.am change to add -ldl for the blocks runtime was something I missed in #139 (it is only needed when building libdispatch with an embedded blocks runtime without including the Swift API).

The rest are minimal fixes to get C compilation to succeed.


src/shims/linux_stubs.h
```diff
@@ -25,6 +25,8 @@ typedef uint32_t mach_port_t;
 #define  MACH_PORT_NULL (0)
 #define  MACH_PORT_DEAD (-1)
 
+#define VOUCHER_CURRENT		((OS_OBJECT_BRIDGE voucher_t)(void *)~2ul)
```
MadCoder: please add:

```
#if OS_VOUCHER_ACTIVITY_SPI
```
guards around the uses of it instead

src/shims/linux_stubs.h
```diff
@@ -43,11 +45,21 @@ typedef uint32_t dispatch_mach_reason_t;
 
 typedef uint32_t voucher_activity_mode_t;
 
+typedef uint32_t firehose_activity_id_t;
```
MadCoder: I think that hunk would do it in voucher.c instaed of adding these back:

```
+#if OS_VOUCHER_ACTIVITY_SPI
 void*
 voucher_activity_get_metadata_buffer(size_t *length)
 {

...
 void
 voucher_activity_initialize_4libtrace(voucher_activity_hooks_t hooks)
 {
        (void)hooks;
 }
+#endif // OS_VOUCHER_ACTIVITY_SPI
```

dgrove-oss:
redone as suggested and squashed.  Should we try to get the modulemap problem sorted out in this pull request too? (ie, don't merge yet).

MadCoder:
I think the modulemap is a different enough issue that goes beyond just simple `#if` guards so I would make it a separate pull request, but YMMV.

MadCoder:
(or a separate commit at least anyway)

dgrove-oss:
ok.  then I think we are done with this one.   How do you want to manage getting the fixes onto both the master and the swift3 branches?

MadCoder:
I think @das should cherry-pick your commits on his branch, so I'll assign this to him to do that

.gitignore
```diff
@@ -29,3 +29,5 @@ config
 configure
 libtool
 .dirstamp
+dispatch/module.modulemap
```
MadCoder: these should start with /

else you will ignore any file named `dispatch/module.map` from the repository which is not what you mean here, see `gitignore(5)`:

```
       o   A leading slash matches the beginning of the pathname. For example,
           "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".
```

MadCoder:
that all looks very good to me but for the gitignore nit.

@das: when you have time, can you cherry pick the changes here on your two pull requests for master & swift-3 ?

dgrove-oss:
gitignore fixed.   This builds for me, so cherry pick at will.

I'm now starting to debug to figure out why dispatch_api hangs.  Have about an hour left this afternoon.

MadCoder:
👍

das:
I merged to das-darwin-006-merge-master and am cherry-picking onto das-darwin-006-merge-swift-3.0

PR #147, opened 2016-08-15
fix help string for --enable-embedded-blocks-runtime

dgrove-oss:


das:
needs a matching PR for swift-3.0-branch

PR #148, opened 2016-08-15
update bug reporting URL and project homepage

dgrove-oss:


das:
needs a matching PR for swift-3.0-branch

PR #152, opened 2016-08-17
bugfix in Linux wrapper: DispatchGroup.leave should call leave, not e…

dgrove-oss:
…nter!

### DIFF
--- a/src/swift/Wrapper.swift
+++ b/src/swift/Wrapper.swift
@@ -63,7 +63,7 @@ public class DispatchGroup : DispatchObject {
 	}
 
 	public func leave() {
-		dispatch_group_enter(__wrapped)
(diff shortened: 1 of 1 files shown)


MadCoder:
wow, we need this for swift-3.0

@mwwa do we have the same bug in the compiler?

mwwa:
@MadCoder No, the compiler uses APINotes to rename ```dispatch_group_leave``` -> ```DispatchGroup.leave(self:)```

dgrove-oss:
I'm pretty sure it was my own stupid cut & paste error when writing the wrapper.  Scary.  We really need to get some tests of the Swift API.    I found this while debugging a disabled test case in TestNSOperationQueue from the Foundation test suite (test_OperationPriorities was hanging because leave wasn't wired correctly).

PR #155, opened 2016-08-19
Fix building on armv7

hpux735:
This patch enables compilation from build-script on armv7.  I don't expect this patch to be without controversy; rather I'd like this to be the opening point for some discussion about the specific implementation.

In particular, I'd like to draw attention to this PR at foundation: https://github.com/apple/swift-corelibs-foundation/pull/399/files which addresses val_list in the same way as presented here.

Furthermore, PAGE_SIZE is, for some reason, undefined on ARM.  The definition provided here is an assumption, and kind of a hack.  I'd happy to receive some guidance whether there is a better way to move forward.

Thanks!

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -147,6 +147,12 @@ AC_ARG_WITH([swift-toolchain],
    case $target_os in
       linux*)
 	    os_string="linux"
+	    case $target_cpu in
+		    armv7l*)
+			target_cpu="armv7"
+			;;
+			*)
+            esac
 	    ;;
 	  *)
         os_string=$target_os
--- a/dispatch/dispatch.h
+++ b/dispatch/dispatch.h
@@ -49,7 +49,9 @@
 
 #if defined(__linux__) && defined(__has_feature)
 #if __has_feature(modules)
+#if !defined(__arm__)
 #include <stdio.h> // for off_t (to match Glibc.modulemap)
+#endif
 #endif
 #endif
 
--- a/src/io.c
+++ b/src/io.c
@@ -24,6 +24,10 @@
 #define DISPATCH_IO_DEBUG DISPATCH_DEBUG
 #endif
 
+#ifndef PAGE_SIZE
+#define PAGE_SIZE getpagesize()
+#endif
+
 #if DISPATCH_DATA_IS_BRIDGED_TO_NSDATA
 #define _dispatch_io_data_retain(x) _dispatch_objc_retain(x)
 #define _dispatch_io_data_release(x) _dispatch_objc_release(x)


src/io.c
```diff
@@ -24,6 +24,10 @@
 #define DISPATCH_IO_DEBUG DISPATCH_DEBUG
 #endif
 
+#ifndef PAGE_SIZE
+#define PAGE_SIZE 4096
```
MadCoder: PAGE_SIZE is in <sys/user.h> on linux I think.

dispatch/dispatch.h
```diff
@@ -49,7 +49,7 @@
 
 #if defined(__linux__) && defined(__has_feature)
 #if __has_feature(modules)
-#include <stdio.h> // for off_t (to match Glibc.modulemap)
+//#include <stdio.h> // for off_t (to match Glibc.modulemap)
```
MadCoder: there was a patch to explicitly add it so I doube this is a good idea

configure.ac
```diff
@@ -147,6 +147,12 @@ AC_ARG_WITH([swift-toolchain],
    case $target_os in
       linux*)
 	    os_string="linux"
+	    case $target_cpu in
```
MadCoder: why is this needed? libtool should know what to do

src/io.c
MadCoder: yup just checked, it's where it is, don't ask me how I can ever remember things like that...

configure.ac
hpux735: Because it gets it wrong.  Much earlier in the evolution of swift on arm, we decided to canonicalize the representation of armv7 to "armv7" rather than the alternate variant "armv7l"  You'll find similar conversions other places.  This simply brings libdispatch inline with other components.

dispatch/dispatch.h
hpux735: Understood.  We have to solve the duplicate definition of va_list somehow.  See https://github.com/apple/swift-corelibs-foundation/pull/399/files for more context.  The same is required for Foundation.

src/io.c
hpux735: Very interesting.  PAGE_SIZE is not defined in sys/user.h on arm.  It's referenced in a few places, but I can't see how it's defined in those cases.

```
 wdillon@tegra-ubuntu  /usr/include  ack-grep "PAGE_SIZE"
scsi/sg.h
188:#define SG_SCATTER_SZ (8 * 4096)  /* PAGE_SIZE not available to user */
190:   The value must be a power of 2 and <= (PAGE_SIZE * 32) [131072 bytes on
191:   i386]. The minimum value is PAGE_SIZE. If scatter-gather not supported
194:   PAGE_SIZE by calling getpagesize() defined in unistd.h . */

elf.h
1882:#define PF_HP_PAGE_SIZE   		0x00100000

arm-linux-gnueabihf/bits/xopen_lim.h
50:   PAGE_SIZE	Size of bytes of a page.

arm-linux-gnueabihf/bits/confname.h
135:#define    	_SC_PAGE_SIZE  			_SC_PAGESIZE

asm-generic/shmparam.h
4:#define SHMLBA PAGE_SIZE     	 /* attach addr a multiple of this */

linux/resource.h
71:#define MLOCK_LIMIT 	((PAGE_SIZE > 64*1024) ? PAGE_SIZE : 64*1024)

linux/vhost.h
58:#define VHOST_PAGE_SIZE 0x1000

linux/kvm.h
352:   	((PAGE_SIZE - sizeof(struct kvm_coalesced_mmio_ring)) / \
499:#define KVM_PPC_PAGE_SIZES_MAX_SZ  	8
509:   	struct kvm_ppc_one_page_size enc[KVM_PPC_PAGE_SIZES_MAX_SZ];
512:#define KVM_PPC_PAGE_SIZES_REAL    		0x00000001
519:   	struct kvm_ppc_one_seg_page_size sps[KVM_PPC_PAGE_SIZES_MAX_SZ];

linux/binfmts.h
14:#define MAX_ARG_STRLEN (PAGE_SIZE * 32)
```

MadCoder: then you should use getpagesize() or sysconf(_SC_PAGESIZE) for the define. I assume arm doesn't define it because there are both 4k and 16k devices out there and it's not tied to the ABI.

dispatch/dispatch.h
MadCoder: commenting out is not the right patch anyway, this is an API header, you just don't do that. either the full hunk is to be removed, or it should stay. @dgrove-oss or @seabaylea should advise here.

configure.ac
MadCoder: huh then libtool should be fixed but oh well.

hpux735: How would libtool know what convention Swift uses for the architecture directory?

dispatch/dispatch.h
hpux735: Sure.  Like I said, this is a starting point for the conversation.  I know that the header is necessary for some cases, and I'd like understand that so we can come up with a solution.  This is the minimal change that fixes it for arm.  There are a ton of possible solutions of varying invasiveness.

MadCoder: a pull request is not the right discussion forum IMO, you should move that discussion to the corelibs-dev mailing list.

configure.ac
MadCoder: hah, fair enough

dispatch/dispatch.h
hpux735: Ok, I'll bring it up on list. I do like PRs because it's less abstract. :)

hpux735:
Thanks, @MadCoder !

PR #157, opened 2016-08-19
More changes for SE-0103 (@noescape by default).

gonzalolarralde:
The following declarations that were `@escaping` (default, actually) before `SE-0103` and that I think they should remain `@escaping`.
- [`dispatch_data_create(…, destructor)`](https://github.com/apple/swift-corelibs-libdispatch/blob/95875d7bed3f66ed8186f9f77bad8ad23f80d152/src/swift/Private.swift#L72) along with [`DispatchData.Deallocator.custom`](https://github.com/apple/swift-corelibs-libdispatch/blob/master/src/swift/Data.swift#L36-L38)
- [`dispatch_queue_set_specific(…, destructor)`](https://github.com/apple/swift-corelibs-libdispatch/blob/master/src/swift/Private.swift#L216) (there's no associated typealias / parameter for this one, it's just being called with [`Queue._destructDispatchSpecificValue`](https://github.com/apple/swift-corelibs-libdispatch/blob/1a7ff3f3e1073eb3352a56ab121ccfa712c42cef/src/swift/Queue.swift#L368-L372) as parameter)

And finally,
- [`dispatch_source_set_event_handler(…, handler)`](https://github.com/apple/swift-corelibs-libdispatch/blob/master/src/swift/Private.swift#L270)
- [`dispatch_source_set_cancel_handler(…, handler)`](https://github.com/apple/swift-corelibs-libdispatch/blob/master/src/swift/Private.swift#L276)
- [`dispatch_source_set_registration_handler(…, handler)`](https://github.com/apple/swift-corelibs-libdispatch/blob/master/src/swift/Private.swift#L324)

along with
- [`DispatchSourceHandler`](https://github.com/apple/swift-corelibs-libdispatch/blob/55261225184e49c6a42c38bbedb144c2610def4a/src/swift/Wrapper.swift#L209)
- [`_DispatchBlock`](https://github.com/apple/swift-corelibs-libdispatch/blob/1a7ff3f3e1073eb3352a56ab121ccfa712c42cef/src/swift/Block.swift#L103)

Finally, the `rescue` parameter in `DispatchQueue._syncHelper` was marked as `@escaping` due to a manual migration, but there's nothing keeping a reference to it.

Please let me know if this patch works for you, or if there's anything that needs to be changed. /cc @slavapestov @dgrove-oss 

Thanks.

### DIFF
--- a/src/swift/Queue.swift
+++ b/src/swift/Queue.swift
@@ -235,7 +235,7 @@ public extension DispatchQueue {
 		fn: (DispatchWorkItem) -> (), 
 		flags: DispatchWorkItemFlags,
 		execute work: () throws -> T,
-		rescue: @escaping ((Swift.Error) throws -> (T))) rethrows -> T
+		rescue: ((Swift.Error) throws -> (T))) rethrows -> T
 	{
 		var result: T?
 		var error: Swift.Error?


src/swift/Block.swift
```diff
@@ -104,8 +104,8 @@ public class DispatchWorkItem {
 /// uses @convention(block). This is to avoid unnecessary bridging between
 /// C blocks and Swift closures, which interferes with dispatch APIs that depend
 /// on the referential identity of a block. Particularly, dispatch_block_create.
-internal typealias _DispatchBlock = @convention(block) () -> Void
-internal typealias dispatch_block_t = @convention(block) () -> Void
+internal typealias _DispatchBlock = @escaping @convention(block) () -> Void
+internal typealias dispatch_block_t = @escaping @convention(block) () -> Void
```
MadCoder: that seems wrong, dispatch_sync() takes a dispatch_block_t and is definitely not escaping

gonzalolarralde: I get your point, but for what I understand that is a change that should concern the C part of libdispatch. Today `dispatch_block_t` is being declared in `object.h` like this:

```
typedef void (^dispatch_block_t)(void);
```

and `internal typealias dispatch_block_t` doesn't seem to be more than a redeclaration of the existing C type. `dispatch_block_t` is being used both in `dispatch_sync` and `dispatch_async` for that matter, that means that it can be both `@escaping` and `@noescape`, and from both options `@escaping` is not the correct, but the safe option.

A different approach may be having a `dispatch_noescape_block_t` with `__attribute__((noescape))`, but that will create a difference with the darwin API.

For what I understand here, `internal typealias dispatch_block_t` is just briging the C type `oid (^dispatch_block_t)(void)` to the Swift scope, and it's being only used internally.

Am I getting this wrong? What would be the correct solution given the context?

Thanks.

MadCoder: my point is that I find it troubling to annotate the type, as if the overlay has to do a
```
dispatch_sync(q) { ... }
```

some day for some weird reason, we will get a poorer performance for it. The importer should already have annotated the C functions (that are properly `__attribute__((noescape))` where it makes sense) properly already anyway, so I think the one on dispatch_block_t is wrong.

@mwwa  do I make sense?

mwwa: @gonzalolarralde Does the ```@noescape``` from ```dispatch_sync``` itself still override the declaration of the block type itself? If so I suspect we're safe, or at least we're no worse than we were before.

src/swift/Block.swift
```diff
@@ -104,8 +104,8 @@ public class DispatchWorkItem {
 /// uses @convention(block). This is to avoid unnecessary bridging between
 /// C blocks and Swift closures, which interferes with dispatch APIs that depend
 /// on the referential identity of a block. Particularly, dispatch_block_create.
-internal typealias _DispatchBlock = @convention(block) () -> Void
-internal typealias dispatch_block_t = @convention(block) () -> Void
+internal typealias _DispatchBlock = @escaping @convention(block) () -> Void
```
slavapestov: This will fail to compile on swift-3.0-branch as of yesterday. You cannot put @escaping anywhere except for a function parameter type.

Basically, if your parameter is written as just `_DispatchBlock`, it will be non-escaping. If you write `@escaping _DispatchBlock`, it is escaping.

src/swift/Block.swift
slavapestov: Also note that `@noescape` is now deprecated, and no longer has any functional effect.

src/swift/Wrapper.swift
```diff
@@ -206,7 +206,7 @@ internal class __DispatchData : DispatchObject {
 	}
 }
 
-public typealias DispatchSourceHandler = @convention(block) () -> Void
+public typealias DispatchSourceHandler = @escaping @convention(block) () -> Void
```
slavapestov: Again, please test with the latest compiler. You should get a diagnostic here.

src/swift/Block.swift
gonzalolarralde: @slavapestov ok, understood! I'm thinking on adding the following macro declaration to match the already existing `DISPATCH_NOESCAPE` solution existing in [base.h](https://github.com/apple/swift-corelibs-libdispatch/blob/ab16f5e62859ff2f54996b8838f8304a8d125102/dispatch/base.h#L164), and keep compatibility with different compiler versions/implementations:

```
#if __has_attribute(escaping)
#define DISPATCH_ESCAPING __attribute__((__escaping__))
#define DISPATCH_NOESCAPE
#if __has_attribute(noescape)
#define DISPATCH_ESCAPING
#define DISPATCH_NOESCAPE __attribute__((__noescape__))
#else
#define DISPATCH_ESCAPING
#define DISPATCH_NOESCAPE
#endif
```

@mwwa / @MadCoder I think this answers the question. I'm gonna keep working on this, unless someone is already doing it internally.

Thanks again for the comments.

src/swift/Block.swift
gonzalolarralde: @slavapestov thanks! just fixed it.

I'm also getting an error now for the block being optional. I submitted a bug, this seemed to be an issue also with `@noescape`, but with other implications now maybe, if those blocks are defaulted to noescaping: [SR-2444](https://bugs.swift.org/browse/SR-2444)

MadCoder:
@mmwa do we have the similar patch for the compiler?
@gonzalolarralde please squash in a single commit

gonzalolarralde:
@MadCoder Done

gonzalolarralde:
Based on this [swift#4905](https://github.com/apple/swift/pull/4905) and this [SR-2444](https://bugs.swift.org/browse/SR-2444) I will remove `@escaping` from the optional block, as they're already escaping by default, and there's no way to mark them as no-escape.

gonzalolarralde:
@MacCoder Ok, done!

PR #159, opened 2016-08-19
Disable the pingpong tests for now since it is causing failures on the build bots

phausler:


phausler:
@swift-ci Please test linux

seabaylea:
You might want to merge #158 instead as it disables two other failing tests as well.

PR #162, opened 2016-08-24
Android support

gonzalolarralde:
This changeset adds support for Android cross-compilation with the same restrictions applied to `swift` project.
## Options
- `--disable-build-tests`: tests are failing to be built due to the lack of `spawn`. Until this is fixed, the flag will be useful to complete compilation
- `--with-android-ndk`: Android NDK path to be used as a base for cross toolchain, SDK and support files.
- `--with-android-ndk-gcc-version`: NDK GCC version to be used (defaults to `4.9`)
- `--with-android-api-level`: API version to be used for compilation
- `--enable-android`: Enabled Android cross-compilation. In this implementation, as it is being done right now in `swift`, this is enforcing `arm-linux-androideabi` `armeabi-v7a`. 
## Issues

There are a few things in this changeset that I'd like to iterate, assuming there are better or more generic solution that can be applied:
- [x] `src/shims/android` includes: In this directory I created [`sys/sysctl.h`](https://github.com/gonzalolarralde/swift-corelibs-libdispatch/blob/android-support/src/shims/android/sys/sysctl.h) + [`syscall.h`](https://github.com/gonzalolarralde/swift-corelibs-libdispatch/blob/android-support/src/shims/android/syscall.h) to redirect headers to their locations in Android. -- _update: moved to compiler conditionals instead of include overlay. [commit](https://github.com/apple/swift-corelibs-libdispatch/pull/162/commits/e67e77f9eda8b87eca90795364e4bc4cd0839582)_
- [x] `src/shims/android` libraries: `libpthread.a` + `librt.a` are dummy libraries because of the [difference described here](https://gist.github.com/leto/663362#file-gistfile1-txt-L85). To avoid this both `libkqueue` and `libpwq` should be modified to avoid asking libtool to link against them.
- [x] [`configure.ac` and `src/Makefile.am`](https://github.com/gonzalolarralde/swift-corelibs-libdispatch/commit/2b2cb83a1810fe2f8dd37f5c64d0ce00fdcafeda): This the first time I deal with `autotools`, so I think there should be a better way to implement this. `(C|CXX|LD)FLAGS` params I think will be needed explicitly declared somewhere to enable cross compilation, but maybe in `AM_$0_FLAGS`? I think there are some overwrites in the `am` files that were causing troubles, but they can be fixed if this is the right direction. Removing flags from `ac_configure_args` like that with `sed` seems like a terrible idea, maybe both `libkqueue` and `libpwq` need to be extended somehow to support this in a better way?
- [x] and more…
## Build

This is an example call to compile `libdispatch` for Android. This is similarly formatted to the call that build script in `swift` does to compile `libdispatch`.

```
$ sh autogen.sh
$ export SWIFT_ANDROID=<...>
$ export NDK_PATH=<...>
$ env \
CC="$SWIFT_ANDROID/build/Ninja-ReleaseAssert/llvm-linux-x86_64/bin/clang" \
CXX="$SWIFT_ANDROID/build/Ninja-ReleaseAssert/llvm-linux-x86_64/bin/clang++" \
SWIFTC="$SWIFT_ANDROID/build/Ninja-ReleaseAssert/swift-linux-x86_64/bin/swiftc" \
$SWIFT_ANDROID/swift-corelibs-libdispatch/configure \
--with-swift-toolchain="$SWIFT_ANDROID/build/Ninja-ReleaseAssert/swift-linux-x86_64/" \
--with-build-variant=release \
--enable-android \
--host=arm-linux-androideabi \
--with-android-ndk=$NDK_PATH \
--with-android-api-level=21 \
--disable-build-tests
$ make
```

_Side note: `va_list` issue mentioned in #155 is affecting the build process. There's no patch for that issues here, to avoid mixing problems that may be addressed in different ways._
## Outcome

This compilation process creates the shared library with the following ELF header:

```
ELF Header:
  Magic:   7f 45 4c 46 01 01 01 00 00 00 00 00 00 00 00 00 
  Class:                             ELF32
  Data:                              2's complement, little endian
  Version:                           1 (current)
  OS/ABI:                            UNIX - System V
  ABI Version:                       0
  Type:                              DYN (Shared object file)
  Machine:                           ARM
  Version:                           0x1
  Entry point address:               0x0
  Start of program headers:          52 (bytes into file)
  Start of section headers:          1060700 (bytes into file)
  Flags:                             0x5000200, Version5 EABI, soft-float ABI
  Size of this header:               52 (bytes)
  Size of program headers:           32 (bytes)
  Number of program headers:         9
  Size of section headers:           40 (bytes)
  Number of section headers:         50
  Section header string table index: 49
```

and dependencies

```
 0x00000001 (NEEDED)                     Shared library: [libstdc++.so]
 0x00000001 (NEEDED)                     Shared library: [libm.so]
 0x00000001 (NEEDED)                     Shared library: [libc.so]
 0x00000001 (NEEDED)                     Shared library: [libdl.so]
```

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -41,6 +41,7 @@ libdispatch_la_SOURCES=		\
 	trace.h					\
 	voucher_internal.h		\
 	firehose/firehose_internal.h \
+	shims/android_stubs.h	\
 	shims/atomic.h			\
 	shims/atomic_sfb.h		\
 	shims/getprogname.h		\
@@ -148,7 +149,7 @@ SWIFT_SRC_FILES=\
 SWIFT_ABS_SRC_FILES = $(SWIFT_SRC_FILES:%=$(abs_srcdir)/%)
 SWIFT_OBJ_FILES = $(abs_builddir)/swift/swift_overlay.o
 
-SWIFTC_FLAGS = -Xcc -fmodule-map-file=$(abs_top_srcdir)/dispatch/module.modulemap -I$(abs_top_srcdir) -Xcc -fblocks
+SWIFTC_FLAGS+= -Xcc -fmodule-map-file=$(abs_top_srcdir)/dispatch/module.modulemap -I$(abs_top_srcdir) -Xcc -fblocks
 if DISPATCH_ENABLE_OPTIMIZATION
 SWIFTC_FLAGS+=-O
 endif
@@ -180,4 +181,3 @@ BUILT_SOURCES=$(MIG_SOURCES) $(DTRACE_SOURCES)
 nodist_libdispatch_la_SOURCES=$(BUILT_SOURCES)
 CLEANFILES=$(BUILT_SOURCES) $(SWIFT_GEN_FILES)
 DISTCLEANFILES=pthread_machdep.h pthread System mach objc
-
--- a/src/internal.h
+++ b/src/internal.h
@@ -155,6 +155,7 @@
 #endif
 
 /* private.h must be included last to avoid picking up installed headers. */
+#include <pthread.h>
 #include "os/object_private.h"
 #include "queue_private.h"
 #include "source_private.h"
@@ -245,7 +246,11 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
 #include <sys/event.h>
 #include <sys/mount.h>
 #include <sys/queue.h>
+#ifdef __ANDROID__
+#include <linux/sysctl.h>
+#else
 #include <sys/sysctl.h>
+#endif /* __ANDROID__ */
 #include <sys/socket.h>
 #include <sys/time.h>
 #include <sys/mman.h>
--- a/src/queue.c
+++ b/src/queue.c
@@ -976,6 +976,7 @@ _dispatch_get_mach_host_port(void)
 #include <unistd.h>
 #include <sys/syscall.h>
 
+#ifndef __ANDROID__
 #ifdef SYS_gettid
 DISPATCH_ALWAYS_INLINE
 static inline pid_t
@@ -985,7 +986,8 @@ gettid(void)
 }
 #else
 #error "SYS_gettid unavailable on this system"
-#endif
+#endif /* SYS_gettid */
+#endif /* ! __ANDROID__ */
 
 #define _tsd_call_cleanup(k, f)  do { \
 		if ((f) && tsd->k) ((void(*)(void*))(f))(tsd->k); \
--- a/src/shims/getprogname.h
+++ b/src/shims/getprogname.h
@@ -23,11 +23,18 @@
 #define __DISPATCH_SHIMS_GETPROGNAME__
 
 #if !HAVE_GETPROGNAME
+
+#ifdef __ANDROID__
+extern const char *__progname;
+#endif /* __ANDROID */)
+
 static inline char *
 getprogname(void)
 {
 # if HAVE_DECL_PROGRAM_INVOCATION_SHORT_NAME
 	return program_invocation_short_name;
+# elif defined(__ANDROID__)
+	return __progname;
 # else
 #   error getprogname(3) is not available on this platform
 # endif
--- a/src/shims/linux_stubs.h
+++ b/src/shims/linux_stubs.h
@@ -71,7 +71,9 @@ typedef void (*dispatch_mach_msg_destructor_t)(void*);
 #endif
 
 // SIZE_T_MAX should not be hardcoded like this here.
-#define SIZE_T_MAX (0x7fffffff)
+#ifndef SIZE_T_MAX
+#define SIZE_T_MAX (~(size_t)0)
+#endif
 
 // Define to 0 the NOTE_ values that are not present on Linux.
 // Revisit this...would it be better to ifdef out the uses instead??
--- a/Makefile.am
+++ b/Makefile.am
@@ -12,6 +12,10 @@ if BUILD_OWN_KQUEUES
   MAYBE_KQUEUES = libkqueue
 endif
 
+if BUILD_TESTS
+  MAYBE_TESTS = tests
+endif
+
 SUBDIRS=						\
 	dispatch					\
 	$(MAYBE_PTHREAD_WORKQUEUES)	\
@@ -20,7 +24,7 @@ SUBDIRS=						\
 	os							\
 	private						\
 	src							\
-	tests
+	$(MAYBE_TESTS)
 
 EXTRA_DIST=					\
 	README.md				\
--- a/configure.ac
+++ b/configure.ac
@@ -11,6 +11,10 @@ AC_CONFIG_MACRO_DIR([m4])
 ac_clean_files=a.out.dSYM
 AM_MAINTAINER_MODE
 
+AC_CANONICAL_BUILD
+AC_CANONICAL_HOST
+AC_CANONICAL_TARGET
+
 #
 # Command line argument to specify build variant (default to release).
 # Impacts default value of CFLAGS et al. so must come before AC_PROG_CC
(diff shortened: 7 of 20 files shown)


src/shims/android/syscall.h
```diff
@@ -0,0 +1 @@
+#import <sys/syscall.h>
```
MadCoder: please just have

```
#ifdef __ANDROID__
#include <sys/syscall.h>
#endif
```

in the right place instead of these overlaying of headers

src/shims/getprogname.h
```diff
@@ -23,11 +23,18 @@
 #define __DISPATCH_SHIMS_GETPROGNAME__
 
 #if !HAVE_GETPROGNAME
+
+#if __ANDROID__
```
MadCoder: should be ifdef

src/shims/getprogname.h
```diff
@@ -23,11 +23,18 @@
...
+extern const char *__progname;
+#endif /* __ANDROID */)
+
 static inline char *
 getprogname(void)
 {
 # if HAVE_DECL_PROGRAM_INVOCATION_SHORT_NAME
 	return program_invocation_short_name;
+# elif __ANDROID__
```
MadCoder: should be elif defined()

src/internal.h
```diff
@@ -154,6 +154,8 @@
 #define DISPATCH_PURE_C 1
 #endif
 
+#include <pthread.h>
```
MadCoder: why? you should have it from "private/private.h" already

src/internal.h
```diff
@@ -368,6 +370,17 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
 			TRASHIT((head)->tqh_last); \
 		} while (0)
 
+#ifndef TAILQ_FOREACH_SAFE
```
MadCoder: all this should be in a "shims/androind_stubs.h" like linux has

src/queue.c
```diff
@@ -976,6 +976,7 @@ _dispatch_get_mach_host_port(void)
 #include <unistd.h>
 #include <sys/syscall.h>
 
+#ifndef __ANDROID__
```
MadCoder: does android already defines gettid() ? because dispatch won't work without it

src/internal.h
gonzalolarralde: @MadCoder `pthread_attr_t` is required by `private/queue_private.h`. In other platforms, this type gets declared by an include in `sys/types.h` to `pthreadtypes.h`.

Android says in `sys/types.h:106`

```
/* while POSIX wants these in <sys/types.h>, we
 * declare then in <pthread.h> instead */
#if 0
typedef  .... pthread_attr_t;
typedef  .... pthread_cond_t;
typedef  .... pthread_condattr_t;
typedef  .... pthread_key_t;
typedef  .... pthread_mutex_t;
typedef  .... pthread_once_t;
typedef  .... pthread_rwlock_t;
typedef  .... pthread_rwlock_attr_t;
typedef  .... pthread_t;
#endif
```

so we need to include it before `private/queue_private.h`. I can put it in that file instead if preferred.

src/queue.c
gonzalolarralde: @MadCoder `gettid` is already defined in `unistd.h:57` for API 21 (and lower also) as `extern pid_t  gettid(void) __pure2`. Android declaration is not `inline` as `src/queue.c`, but it shouldn’t have a performance impact in this context, right?

src/internal.h
MadCoder: oh I see, then I would move that `#include` right after this comment:

```
/* private.h must be included last to avoid picking up installed headers. */
```

src/queue.c
MadCoder: perfect, I just wanted to make sure. we don't care about the performance because this is cached once per thread.

src/Makefile.am
```diff
@@ -64,10 +65,15 @@ DISPATCH_CFLAGS=-Wall $(VISIBILITY_FLAGS) $(OMIT_LEAF_FP_FLAGS) \
 if DISPATCH_ENABLE_ASSERTS
 DISPATCH_CFLAGS+=-DDISPATCH_DEBUG=1
 endif
+
+if ANDROID
+ANDROID_LIBS= -L$(top_srcdir)/src/shims/android
```
gonzalolarralde: @MadCoder now that header overlays are gone I'm even more uncomfortable to having this dummy files (`libpthread.a` and `librt.a`) in `src/shims/android`. I was thinking on `vendors/android/lib(pthread|rt).a` or `deps/android/lib(pthread|rt).a`. What do you think?

More details on why this is required [here](https://github.com/apple/swift-corelibs-libdispatch/pull/162/commits/2b2cb83a1810fe2f8dd37f5c64d0ce00fdcafeda)

MadCoder: you should do pull requests for both these projects to be merged in their upstreams and update the submodules here instead.

these .a's are really gross IMNSHO.

src/shims/linux_stubs.h
```diff
@@ -71,7 +71,9 @@ typedef void (*dispatch_mach_msg_destructor_t)(void*);
 #endif
 
 // SIZE_T_MAX should not be hardcoded like this here.
+#ifndef SIZE_T_MAX
 #define SIZE_T_MAX (0x7fffffff)
```
MadCoder: while you're at it, can you change this as `(~(size_t)0)` please, this just makes me cringe each time I see it :P

MadCoder:
I think the .a issue should be resolved with other upstreams.
@dgrove-oss or @seabaylea or other linux porters should look at the sanity of the autoconf part, I didn't even read it that carefully.

src/shims/linux_stubs.h
gonzalolarralde: haha sure

src/Makefile.am
gonzalolarralde: Trust me, I'm not proud about them. The submodules are not in the `apple` organization, so I wasn't sure how that may affect the chances + delay to include this. I'll start working on that tho.

dgrove-oss: @frankeh is a committer on libpwq, so processing a pull request there should be fast.  We usually get good time turn around time from libkqueue (hours to days).  So, it's worth trying to do the right thing and maybe dropping this hunk from the PR

dgrove-oss:
The autoconf/automake looks plausible to me.   I tried the disable-tests argument to configure and a vanilla build on linux and both work fine.  I assume the big android block in configure.ac does what you want (didn't check it).

src/Makefile.am
gonzalolarralde: Great! I'll send a PR to both projects then. Thanks @dgrove-oss :)

gonzalolarralde:
Ok, dependency PRs created.

- [ ] [mheily/libkqueue#15](https://github.com/mheily/libkqueue/issues/15)
- [ ] [mheily/libpwq#17](https://github.com/mheily/libpwq/issues/17)

@frankeh would you mind to take a look?

Thanks.

gonzalolarralde:
@MadCoder @dgrove-oss Changes in `libpwq` and `libkqueue` has been approved and merged. I just rebased the branch and updated the submodule heads. Let me know if there's anything else I should do. I can squash the commits also.

Cheers

MadCoder:
the patch looks reasonnable, please squash in a single commit with a good message indeed

dgrove-oss:
We realized yesterday that the libpwq and libkqueue change actually broke the build on linux because the default for bionic-libc was backwards.   Fix has been merged to libpwq and is pending for libkqueue.  We'll need to wait for the libkqueue fix to be merged, then update the submodule heads again before this can be merged.

gonzalolarralde:
@dgrove-oss I just seen it, so sorry about that. Thanks for submitting the patch. I'll wait until it is merged to update the heads and squash everything.

dgrove-oss:
I've submitted #179 to update the submodule versions in isolation.  All the changes we need were merged into the upstream repositories over the weekend.  Once #179 is merged, you can drop the submodule changes from this PR and squash the rest of it to a single commit to prepare to merge.

gonzalolarralde:
@dgrove-oss @MadCoder Squashed :)

MadCoder:
@dgrove-oss can you check it doesn't break linux? I'm not familiar with the CI env. I'm ok with the change otherwise

dgrove-oss:
pull request builds fine for me on ubuntu 16.04 and tests pass.  

 I don't think I can trigger a CI test (not in apple organization), but @MadCoder I think if you put @swift-ci please test in a comment it will schedule a CI test for you.

modocache:
@swift-ci Please test

modocache:
@swift-ci please test

gonzalolarralde:
```
TestFoundation: src/common/knote.c:155: knote_disable: Assertion `!(kn->kev.flags & 0x0008)' failed.
/home/buildnode/jenkins/workspace/swift-corelibs-libdispatch-PR-Linux@2/swift/utils/build-script-impl: line 249: 40378 Aborted                 (core dumped) "$@"
/home/buildnode/jenkins/workspace/swift-corelibs-libdispatch-PR-Linux@2/swift/utils/build-script: fatal error: command terminated with a non-zero exit status 134, aborting
Build step 'Execute shell' marked build as failure
```

This is the error the CI is reporting, however I don't see any relation with the changes in this PR. Is it possible that this is a failure somewhere else? Let me know if I need to do anything else.

Thanks!

gonzalolarralde:
Rebased. Is there any change I should work on? Please let me know. @modocache make sense to re-trigger CI test now?

modocache:
@swift-ci please test

modocache:
Sorry I didn't see this sooner, @gonzalolarralde! I can't wait to see this in! :)

johnno1962:
Looks like a problem related to libkqueue https://github.com/mheily/libkqueue/blob/b9e015210bb3e432f44afd054b24a8ceb99eb8e7/src/common/knote.c#L155


TestFoundation: src/common/knote.c:155: knote_disable: Assertion `!(kn->kev.flags & 0x0008)' failed.
/home/buildnode/jenkins/workspace/swift-corelibs-libdispatch-PR-Linux/swift/utils/build-script-impl: line 251:  3921 Aborted                 "$@"
/home/buildnode/jenkins/workspace/swift-corelibs-libdispatch-PR-Linux/swift/utils/build-script: fatal error: command terminated with a non-zero exit status 134, aborting

modocache:
@swift-ci please test

gonzalolarralde:
Rebased.

hpux735:
@gonzalolarralde  Do you have a companion PR for integrating with build-script in the swift project, or does this have to be done manually outside of the greater build?

If it is separate, is it possible to build a dispatch-aware foundation using this work?

By the way, this is fantastic!  Thanks for doing all this!!

gonzalolarralde:
Hi @hpux735, I don't have a companion PR yet, however I've defined a set of options similar to the options needed to call build-script when compiling for Android.

However, I've been working on a dev environment that is compiling all the libraries, and enabling dispatch in foundation.

The project is here: https://github.com/gonzalolarralde/swifty-robot-environment/

The specific compilation script for foundation with all the dependencies + dispatch here: https://github.com/gonzalolarralde/swifty-robot-environment/blob/master/util/prepare_environment/070_build_corelibs_foundation.sh

Some shortcuts have been taken in this scripts. Sorry about that :) I'll try to put some time to add this to build-script and build-script-impl (not sure if the maintainers are accepting changes in this last one due to the Python refactor).

Cheers!

hpux735:
😁👌 Thanks!!

PR #163, opened 2016-08-24
SR-2356 : Remove temporary APIs from Linux overlay

dgrove-oss:
Remove temporary APIs from Linux Dispatch overlay;
swift-corelibs-foundation builds without them and
they are not present in the Darwin Dispatch overlay.


dgrove-oss:
I'll create a matching PR for the swift-3 branch as soon as this one is merged to master.

PR #168, opened 2016-09-09
Correct signedness handling of DispatchTimeInterval

dgrove-oss:
Replicate e32016ffd765020c696bc5d1e42606074305f2db
from the Darwin overlay to the wrapping overlay.

### DIFF
--- a/src/swift/IO.swift
+++ b/src/swift/IO.swift
@@ -88,7 +88,7 @@ public extension DispatchIO {
 	}
 
 	public func setInterval(interval: DispatchTimeInterval, flags: IntervalFlags = []) {
-		dispatch_io_set_interval(self.__wrapped, interval.rawValue, flags.rawValue)
+		dispatch_io_set_interval(self.__wrapped, UInt64(interval.rawValue), flags.rawValue)
 	}
 
 	public func close(flags: CloseFlags = []) {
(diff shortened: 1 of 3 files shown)


dgrove-oss:
Do we need a swift 3 branch version of this patch as well?

mwwa:
This looks good, there's no need to port to the swift repo as the changes are already there. Is that what you meant? (Or do you mean a swift 3 branch of libdispatch?)

mwwa:
@swift-ci please test

dgrove-oss:
I meant the swift3 branch of libdispatch. I think this is an internal implementation change that doesn't impact the exposed API, so we don't need it in swift3-branch, but I wanted to confirm that with you.

mwwa:
Gotcha, yeah that's what I thought too.

dgrove-oss:
The PR test failed with an assertion in libkqueue when running a Foundation test case.  I've built and run locally half a dozen times and can't reproduce.  It's possible I'm missing something, but it might have been a glitch on the test machine.  I did notice at the end of the log that Jenkins complained:
```
[xUnit] [ERROR] - Clock on this slave is out of sync with the master, and therefore 
I can't figure out what test results are new and what are old.

```

mwwa:
Ok, I'll go ahead and merge.

PR #171, opened 2016-09-17
[Diagnostics QOI]: Fix the fix-it for dispatch_apply

karwa:


MadCoder:
@mwwa do we need to port to the compiler too?

@karwa please open a pull request for the swift-3 branch too

mwwa:
Yes we do.

PR #174, opened 2016-09-22
remove unused/incorrect internal init from DispatchIO

dgrove-oss:
DispatchIO should not have an internal init(dispatch_queue_t)
that stores a dispatch_queue_t into a field of type dispatch_io_t.
Bug is only in the wrapping (Linux) overlay; not in Dispatch.apinotes.


MadCoder:
we need a PR for swift-3 too

PR #175, opened 2016-09-22
SR-2656:  extra releases of dispatch_data_t objects on Linux

dgrove-oss:
Fixes an over-release problem in the wrapping overlay that
occurred in callback blocks that take DispatchData structs.
Just going through the glue layer between the Swift and C APIs
should not result in a net change in the reference count of
the dispatch_data_t object that the CDispatch layer is giving
us to hand to the Swift callback block.  However, when the
temporary DispatchData struct we created in the wrapping layer
(and its wrapped __DispatchData object) went out of scope at
the end of the callback, the dispatch_data_t was released
by __DispatchData's deinit resulting in an unintended net -1
from the glue code.

We either need to add a compensating retain before entering the
callback block or suppress the release in the deinit. This patch
suppresses the release by adding an internal init that can
create DispatchData's with borrowed dispatch_data_t objects that
are not released on deinitialization.

### DIFF
--- a/src/swift/Data.swift
+++ b/src/swift/Data.swift
@@ -65,7 +65,11 @@ public struct DispatchData : RandomAccessCollection {
 	}
 
 	internal init(data: dispatch_data_t) {
-		__wrapped = __DispatchData(data: data)
+		__wrapped = __DispatchData(data: data, owned: true)
+	}
+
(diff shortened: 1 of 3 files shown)


MadCoder:
will need a branch for swift-3 too

PR #179, opened 2016-09-25
update libkqueue and libpwq submodules

dgrove-oss:
Update to latest libpwq and libkqueue versions.
This picks up changes for:
  (1) enabling android build
  (2) fixing signal handling (SR-2655)


MadCoder:
you probably want a swift-3 branch

PR #181, opened 2016-09-26
Follow HTTP redirects

ryandesign:
Fixes the `curl` commands by instructing it to follow HTTP redirects.


saltman007web (commented):
review

PR #184, opened 2016-10-12
Swift3: Add dependency on swiftc for %.o.%.swift rules

dgrove-oss:
Detect that swiftc has changed since the Swift file has
been compiled and force it to be recompiled.  The explicit
rm of the target before regenerating it is needed because if
the new compiler generates exactly the same bits as the old
one the .o file isn't updated and the target gets stuck as
being out-of-date.  An explicit rm avoids that problem.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -169,7 +169,8 @@ if DISPATCH_ENABLE_OPTIMIZATION
 SWIFTC_FLAGS+=-O
 endif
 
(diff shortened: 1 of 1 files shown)


dgrove-oss:
https://bugs.swift.org/browse/SR-2922 motivates also merging to Swift3 branch.

erg:
LGTM.

PR #186, opened 2016-10-24
SR-3002: DispatchData.enumerateBytes on Linux leaks the block

dgrove-oss:
To avoid leaking the block parameter of enumerateBytes,
we need to directly call CDispatch.dispatch_data_apply
(not _swift_dispatch_data_apply). However, we need to tell
the compiler that the internal closure that translates between
the C and Swift layers will not cause the captured block
parameter to escape.  If SR-2313 were implemented, we could use
withoutActuallyEscaping.  Since withoutActuallyEscaping is not available,
we instead use unsafeBitCast to bypass the compiler's analysis.

### DIFF
--- a/src/swift/Data.swift
+++ b/src/swift/Data.swift
@@ -91,11 +91,13 @@ public struct DispatchData : RandomAccessCollection {
 	public func enumerateBytes(
 		block: (_ buffer: UnsafeBufferPointer<UInt8>, _ byteIndex: Int, _ stop: inout Bool) -> Void) 
 	{
-		_swift_dispatch_data_apply(__wrapped.__wrapped) { (_, offset: Int, ptr: UnsafeRawPointer, size: Int) in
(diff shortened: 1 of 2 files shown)


dgrove-oss:
@MadCoder any comments on this one?  If the patch looks ok I will go ahead and create a PR to the Swift3 branch as well.

MadCoder:
sorry for some reason it fell through the cracks.

it looks reasonnable to me @mwwa what's your take? do we need the same in the compiler's overlay?

dgrove-oss:
I believe it is a Linux-only problem; initial bug report reported it leaked on Linux but not on Darwin.

dgrove-oss:
I created the matching Swift3 PR: https://github.com/apple/swift-corelibs-libdispatch/pull/189.  If we do get a Swift 3.0.2 release it would be nice to get this fix into it.

dgrove-oss:
@MadCoder @mwwa   ok to merge this to master?  it went into swift3 branch a while ago.

PR #187, opened 2016-11-02
configure support for extended test suite

dgrove-oss:
Add a configure argument to enable extended test suite.
The extended test suite is disabled by default (and thus
will not be run in vanilla Swift CI jobs).

Move yet another intermittently failing test,
dispatch_suspend_timer, out of the list of default tests.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -175,6 +175,18 @@ AC_CHECK_PROG(use_gold_linker, ld.gold, true, false)
 AM_CONDITIONAL(USE_GOLD_LINKER, $use_gold_linker)
 
 #
+# Enable an extended test suite that includes
+# tests that are too unreliable to be enabled by
+# default in the Swift CI environment, but are still
+# useful for libdispatch developers to be able to run.
+#
+AC_ARG_ENABLE([extended-test-suite],
+  [AS_HELP_STRING([--enable-extended-test-suite],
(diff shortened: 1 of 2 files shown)


dgrove-oss:
@erg: in response to email-chain on swift-dev about dispatch_suspend_timer disturbing Swift CI.

erg:
Each Linux Jenkins CI box runs two executors each with 50% of the CPU resources. I found a utility called `stress` that will max out the cores on a machine. It's installable through ``brew install stress`` or from https://people.seas.harvard.edu/~apw/stress/.

If the dispatch tests pass while run in a loop for a few minutes with ``stress --cpu 8 --timeout 60`` (where 8 is the max cores for your machine) then that should be good enough to ensure they pass on the CI machine.

The problem with this patch is I would want to disable these tests for PR testing (really only the `dispatch_suspend_timer` test is breaking things now), but keep them enabled for everything else so we don't have regressions in libdispatch as you suggested in the email thread.

What do you think?

erg:
Actually, looks good since it's just a few more flaky tests. It would be nice to make them all reliable somehow and reenable them.

Thank you!

erg:
Could also do a PR with this change to swift-3.0-branch? Thanks!

dgrove-oss:
Sure, I'll get a Swift 3 branch PR together later this afternoon.

PR #188, opened 2016-11-10
Swift 3: configure support for extended test suite

dgrove-oss:
Add a configure argument to enable extended test suite.
The extended test suite is disabled by default (and thus
will not be run in vanilla Swift CI jobs).

Move yet another intermittently failing test,
dispatch_suspend_timer, out of the list of default tests.


dgrove-oss:
@erg Swift3 branch version of #187

PR #195, opened 2016-12-28
Cleanups

compnerd:


src/internal.h
```diff
@@ -356,14 +356,10 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
 		do { (elm)->field.tqe_prev = NULL; } while (0)
 
 #if DISPATCH_DEBUG
-// sys/queue.h debugging
-#if defined(__linux__)
-#define QUEUE_MACRO_DEBUG 1
```
MadCoder: this breaks people using libbsd-dev and doesn't hurt, please do not remove.
what you probably want to do is to both set QUEUE_MACRO_DEBUG and TRASHIT to support both (?)

src/shims/getprogname.h
```diff
@@ -24,9 +24,9 @@
 
 #if !HAVE_GETPROGNAME
 
-#ifdef __ANDROID__
+#if defined(__ANDROID__)
```
MadCoder: please keep the `#ifdef` only the removal of the `)` below is needed

src/shims/linux_stubs.h
```diff
@@ -16,6 +16,16 @@
 #ifndef __DISPATCH__STUBS__INTERNAL
 #define __DISPATCH__STUBS__INTERNAL
 
+#if !defined(TAILQ_FOREACH_SAFE)
```
MadCoder: please keep consistent style and use `#ifndef`

src/shims/linux_stubs.h
```diff
@@ -16,6 +16,16 @@
...
+#define TAILQ_FOREACH_SAFE(var, head, field, temp)                             \
+  for ((var) = TAILQ_FIRST((head));                                            \
+       (var) && ((temp) = TAILQ_NEXT((var), field), 1); (var) = (temp))
+#endif
+
+#if !defined(TRASHIT)
```
MadCoder: please don't do that, TRASHIT is only meant to be set in debug mode

src/shims/linux_stubs.h
```diff
@@ -16,6 +16,16 @@
 #ifndef __DISPATCH__STUBS__INTERNAL
 #define __DISPATCH__STUBS__INTERNAL
 
+#if !defined(TAILQ_FOREACH_SAFE)
+#define TAILQ_FOREACH_SAFE(var, head, field, temp)                             \
+  for ((var) = TAILQ_FIRST((head));                                            \
```
MadCoder: please use tabs for alignment

src/shims/linux_stubs.h
compnerd: Just following the android pattern, where the `android_stubs.h` always define it.

MadCoder:
@compnerd :

> Just following the android pattern, where the android_stubs.h always define it.

I got the message through mail but couldn't find where it was made after your force push.
I didn't realize Android was doing that and it's wrong, TRASHIT and other similar macros are about poisoning the `<sys/queue.h>` macros, and does have a performance penalty cost for some things that are quite in hotpaths where you don't want it unless you debug. The Android port is wrong and ought to be fixed, would you mind doing it on the same pull request? the rest looks good

compnerd:
@MadCoder sure, I can squash it into the same change.

compnerd:
Gentle post-holiday ping for @MadCoder

src/shims/linux_stubs.h
MadCoder: you haven't handled this review feedback.

Please use `#ifndef` and squash in a single commit before merge, thanks

compnerd:
Address the last bit of the review comments that I had missed.

compnerd:
@swift-ci please test

compnerd:
Hmm, seems that I dont have swift-ci rights to this project.  Mind running this through @MadCoder?  BTW, whats the darwin-pending tag?

MadCoder:
@swift-ci please test

compnerd:
It seems that the failure is unrelated?

PR #198, opened 2017-01-04
SR-3477: replace unsafeBitCast by withoutActuallyEscaping

dgrove-oss:
Now that withoutActuallyEscaping is implemented, use it instead
of unsafeBitCast to allow passing a closure capturing a non-escaping
block down to the C implementation of dispatch_data_apply (which does
not actually escape its block argument).


MadCoder:
@swift-ci please test

dgrove-oss:
CI failure was on swiftpm test.  both dispatch and foundation tests passed.  Should be ok to merge.

PR #199, opened 2017-01-12
consider affinity when computing active cpus on Linux

dgrove-oss:
If pthread_getaffinity_np is available, then use the cpuset
of the current thread to determine the number of active cpus.


MadCoder:
why is that better?

dgrove-oss:
It gets the right answer if the program is being run under numactl with a restricted cpu set.  sysconf doesn't consider scheduler affinity.

MadCoder:
I see, makes sense, could you please add a comment above to explain this shortly in the .c file ? because someone is bound to wonder at some point.

dgrove-oss:
good point; added a comment & squashed commit.

PR #200, opened 2017-01-16
don't hold references to unowned DispatchData objects (SR-3628)

weissi:
In `DispatchIO.read`, the `dispatch_data_t` object received from `dispatch_io_read` will be released once the block passed to `dispatch_io_read` returns.

In the Swift overlay (`src/swift/IO.swift`) the code marks that data object correctly as "borrowedData"

	public func read(offset: off_t, length: Int, queue: DispatchQueue, ioHandler: @escaping (_ done: Bool, _ data: DispatchData?, _ error: Int32) -> Void) {
		dispatch_io_read(self.__wrapped, offset, length, queue.__wrapped) { (done: Bool, data: dispatch_data_t?, error: Int32) in
			ioHandler(done, data.flatMap { DispatchData(borrowedData: $0) }, error)
		}
	}

that however leads to use-after-frees if the Swift `DispatchData` is kept around for longer.

The change I made will `retain` the underlying `dispatch_data_t` if it's "borrowed" which I believe is correct and fixes [SR-3628](https://bugs.swift.org/browse/SR-3628).

### DIFF
--- a/src/swift/Wrapper.swift
+++ b/src/swift/Wrapper.swift
@@ -188,7 +188,6 @@ extension DispatchSource : DispatchSourceProcess,
 
 internal class __DispatchData : DispatchObject {
 	internal let __wrapped:dispatch_data_t
-	internal let __owned:Bool
 
 	final internal override func wrapped() -> dispatch_object_t {
 		return unsafeBitCast(__wrapped, to: dispatch_object_t.self)
@@ -196,13 +195,13 @@ internal class __DispatchData : DispatchObject {
 
 	internal init(data:dispatch_data_t, owned:Bool) {
 		__wrapped = data
-		__owned = owned
+		if !owned {
+			_swift_dispatch_retain(unsafeBitCast(data, to: dispatch_object_t.self))
+		}
 	}
 
 	deinit {
-		if __owned {
-			_swift_dispatch_release(wrapped())
-		}
+		_swift_dispatch_release(wrapped())
 	}
 }
 
@@ -335,3 +334,6 @@ internal enum _OSQoSClass : UInt32  {
 
 @_silgen_name("_swift_dispatch_release")
 internal func _swift_dispatch_release(_ obj: dispatch_object_t) -> Void
+
+@_silgen_name("_swift_dispatch_retain")
+internal func _swift_dispatch_retain(_ obj: dispatch_object_t) -> Void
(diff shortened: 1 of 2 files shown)


dgrove-oss:
borrowedData was added to Data to compensate for memory leaks that I think were related to iteration.   I will go back and look at that PR / SR to see how this proposed change interacts with that.

dgrove-oss:
ok.  This change makes sense to me.   

When I "fixed" the over-release problem of [SR-2656](https://bugs.swift.org/browse/SR-2656) by introducing borrowed in #175 I didn't account for when the __DispatchData that was borrowing the dispatch_data_t outlived the "real" owner. I was too focused on trying to avoid the overhead of an "extra" retain/release operation.  But, it really can't be avoided (as shown by the test case for this bug).

src/swift/Wrapper.swift
```diff
@@ -188,21 +188,20 @@ extension DispatchSource : DispatchSourceProcess,
 
 internal class __DispatchData : DispatchObject {
 	internal let __wrapped:dispatch_data_t
-	internal let __owned:Bool
 
 	final internal override func wrapped() -> dispatch_object_t {
 		return unsafeBitCast(__wrapped, to: dispatch_object_t.self)
 	}
 
 	internal init(data:dispatch_data_t, owned:Bool) {
```
MadCoder: that `owned` tag should really be `transferOwnership: Bool` but the patch LGTM and I should have realized when we reviewed the earlier patch :/

MadCoder:
@swift-ci please test

weissi:
build seems healthy to me but then failed because of

> Build timed out (after 166 minutes). Marking the build as failed.

weissi:
@swift-ci please test

weissi:
@MadCoder are there any special permissions required to '@swift-ci please this'?

MadCoder:
@dgrove-oss I'm not 100% sure, do we need this for another branch than master? we want that for whatever the next swift 3 update will be.

dgrove-oss:
We should get it into the swift-3.1-branch.  I don't think it is worth putting into the swift-3.0-branch.  
@weissi can you easily do the PR to the swift-3.1 branch or would you like me to take care of it?

weissi:
@dgrove-oss sure, on it!

weissi:
@dgrove-oss #201

PR #202, opened 2017-01-21
Fix + and - operations on DispatchTime and DispatchWallTime

moiseev:
... for the cases where seconds parameter is NaN or infinite.
rdar://problem/29764171

Similar to https://github.com/apple/swift/pull/6961


moiseev:
@swift-ci Please test

moiseev:
The CI build is timing out after the libdispatch gets built. Merging this in. The timeout issue to be investigated.

PR #206, opened 2017-01-26
implement pthread_workqueue within libdispatch

dgrove-oss:
Provide an implementation of the pthread_workqueue
functionality as a fully integrated component of
libdispatch.  Integration of the workqueue implementation
into the dispatch code base simplifies the process of
evolving the APIs between the two layers and thus
prepares for future optimization and enhancements.

Initially, the integrated pthread_workqueue is only
enabled by default on Linux. Most of the internals are
built on pthread APIs and thus should be fairly portable.
However, Linux-specific code is used to dynamically estimate
the number of runnable worker threads by reading /proc.
Porting the thread pool management code to non-Linux
platforms would entail providing similar functionality
for those platforms (or otherwise restructuring the manager).


src/queue.c
```diff
@@ -159,7 +160,7 @@ struct dispatch_root_queue_context_s {
 #if HAVE_PTHREAD_WORKQUEUES
 			qos_class_t dgq_qos;
 			int dgq_wq_priority, dgq_wq_options;
-#if DISPATCH_USE_LEGACY_WORKQUEUE_FALLBACK || DISPATCH_USE_PTHREAD_POOL
+#if DISPATCH_USE_LEGACY_WORKQUEUE_FALLBACK || DISPATCH_USE_PTHREAD_POOL || HAVE_INTERNAL_PTHREAD_WORKQUEUE
```
MadCoder: 80 col wrap

src/queue.c
```diff
@@ -719,7 +720,7 @@ _dispatch_root_queues_init_workq(int *wq_supported)
 		result = !r;
 	}
 #endif // HAVE_PTHREAD_WORKQUEUE_SETDISPATCH_NP
-#if DISPATCH_USE_LEGACY_WORKQUEUE_FALLBACK || DISPATCH_USE_PTHREAD_POOL
+#if DISPATCH_USE_LEGACY_WORKQUEUE_FALLBACK || DISPATCH_USE_PTHREAD_POOL || HAVE_INTERNAL_PTHREAD_WORKQUEUE
```
MadCoder: ditto

src/queue.c
```diff
@@ -747,6 +748,15 @@ _dispatch_root_queues_init_workq(int *wq_supported)
 				result = result || dispatch_assume(pwq);
 			}
 #endif // DISPATCH_USE_LEGACY_WORKQUEUE_FALLBACK
+#if HAVE_INTERNAL_PTHREAD_WORKQUEUE
+			if (!disable_wq && qc->dgq_wq_priority != WORKQ_PRIO_INVALID) {
+				int priority = qc->dgq_wq_priority;
+				int overcommit = qc->dgq_wq_options & WORKQ_ADDTHREADS_OPTION_OVERCOMMIT;
```
MadCoder: 80 columns

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
+/*
+ * Copyright (c) 2008-2017 Apple Inc. All rights reserved.
```
MadCoder: that 2008 looks incorrect

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+ * watermark by waking an existing suspended worker thread or spawning an
+ * additional one if no suspended workers are available and the total
+ * number of normal workers is below the allowed maximum number of workers.
+ * The management subsystem will respond to being above the target high
+ * watermark by enqueuing a high-priority workitem that causes the worker
+ * thread that dequeues it to suspend itself.
+ */
+
+
+/////////////////////////////////
```
MadCoder: we never use `////` like that, the rule is `/* */` for things outside of functions and `//` comment when inline inside a function

Here what you mean is:

```
#pragma mark - ...
```

which is an Xcode thing but is easy to wire in vim ;)

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+
+#endif // DISPATCH_ENABLE_PWQ_KEXT
+
+/////////////////////////////////
+// Workqueue internal data structures and functions
+/////////////////////////////////
+
+// Debug messages for threadpool operations
+static bool threadpool_verbose;
+#define verbose_printf(fmt,...) do {								\
```
MadCoder: doesn't _dispatch_debug() already does the same thing more or less?

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+	if (unlikely(threadpool_verbose))								\
+		fprintf(stderr, "libdispatch_threadpool: "fmt"\n", ##__VA_ARGS__); \
+} while (0)
+
+struct work {
+	STAILQ_ENTRY(work) item_entry;
+	void (*func)(void *);
+	void *func_arg;
+};
+typedef struct work * dispatch_workitem_t;
```
MadCoder: dispatch would rather do a:

```
typedef struct dispatch_work_item_s {
} dispatch_work_item_s, *dispatch_work_item_t;
```

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+
+typedef struct _dispatch_workqueue {
```
MadCoder: ditto the tag should be `dispatch_rorkqueue_s`

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+	STAILQ_HEAD(,work) item_listhead;
+	pthread_spinlock_t sl;
+	int priority;
```
MadCoder: the two fields here should use a combined pthread_priority_t longer term *wink*

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+ * The overcommit pool uses a simple coarse locking policy:
+ *   all overcommit pool operations are performed while
+ *   holding the mutex.
+ *
+ * This enables a simple, but non-scalable, implementation
+ * that is adequate as long as operations on the overcommit
+ * pool are relatively infrequent.
+ */
+typedef struct {
+	uint mask;
```
MadCoder: uint32_t

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+ *
+ * The normal pool uses a combination of atomic
+ * operations on a volatile mask word that encodes which
+ * workqueues are non-empty and per-queue synchronization
+ * (currently implemented via pthread_spin_locks).
+ * This is slightly more scalable than the overcommit queue, but
+ * is may be an area that needs future performance enhancement.
+ */
+typedef struct {
+	volatile uint mask;
```
MadCoder: uint32_t

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+	dispatch_workqueue_t queues[WORKQ_NUM_PRIORITIES];
```
MadCoder: work *queues* is confusing, this is really about managing a thread pool for a given priority bucket.

I think the whole type should rather be named:

`dispatch_wq_thread_pool_t` and be referred as a `thread_pool`. or somesuch (just a suggestion, but queue is so overloaded in dispatch that I would shy away from using the term even more)

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+	int32_t low_target = num_cpus;
+	char* env_low_target = getenv("LIBDISPATCH_THREADPOOL_LOW_TARGET");
+	if (unlikely(env_low_target)) {
+		int32_t tmp = atoi(env_low_target);
+		if (tmp > 0) {
+			low_target = tmp;
+		}
+	}
+	int32_t high_target = 2 * low_target;
+	char* env_high_target = getenv("LIBDISPATCH_THREADPOOL_HIGH_TARGET");
```
MadCoder: `const char *env_low_target`

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+
+static void
+_dispatch_workqueue_init_once(void *context DISPATCH_UNUSED)
+{
+	int r;
+
+	// Initialize thread pool size parameters
+	int32_t num_cpus = dispatch_hw_config(active_cpus);
+	int32_t low_target = num_cpus;
+	char* env_low_target = getenv("LIBDISPATCH_THREADPOOL_LOW_TARGET");
```
MadCoder: same that below, use const, and the star sticks to the variable name not the type

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+		if (overcommit_pool.num_spares > 0) {
+			pthread_cond_signal(&overcommit_pool.spare_workers);
+			overcommit_pool.num_spares--;
+		} else {
+			_dispatch_spawn_thread(_dispatch_overcommit_worker_main);
+		}
+		pthread_mutex_unlock(&overcommit_pool.mutex);
+	} else {
+		// normal pool uses finer-grained wq locking and atomic memory ops.
+		pthread_spin_lock(&workq->sl);
```
MadCoder: please don't sue pthread_spin_lock() there are perfectly fine implementations of a better lock in dispatch itself (os/lock.h the unfair lock guy) you should use that instead.

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+int
+dispatch_workqueue_additem_np(dispatch_workqueue_t workq,
+		void (*workitem_func)(void *), void * workitem_arg)
+{
+	dispatch_workitem_t wi = _dispatch_alloc_workitem(workitem_func, workitem_arg);
+	unsigned int wq_index_bit = (0x1 << workq->priority);
+
+	if (unlikely(workq->overcommit)) {
+		// overcommit pool uses trival concurrency control: all operations done holding pool mutex
+		pthread_mutex_lock(&overcommit_pool.mutex);
```
MadCoder: Improvement for later: there's a way to gain concurrency here by using an event count as opposed to a mutex+condvar. We can follow up offline from this review to discuss it, and that doesn't require to be done in the first impl

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+
+
+static void
+_dispatch_workqueue_add_control_workitem(void *op_code)
+{
+	dispatch_workitem_t wi = _dispatch_alloc_workitem(NULL, op_code);
+	dispatch_workqueue_t workq = normal_pool.queues[0];
+	unsigned int wq_index_bit = 1; // highest priority queue
+
+	pthread_spin_lock(&workq->sl);
```
MadCoder: unfair lock please.

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+	r = sigfillset(&mask);
+	(void)dispatch_assume_zero(r);
+	r = _dispatch_pthread_sigmask(SIG_BLOCK, &mask, NULL);
+	(void)dispatch_assume_zero(r);
+
+	r = pthread_mutex_lock(&overcommit_pool.mutex);
+	(void)dispatch_assume_zero(r);
+
+	while (true) {
+		unsigned int idx = ffs(overcommit_pool.mask);
```
MadCoder: `__builtin_ffs()` will avoid a funcall

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+
+		/* Wait for more work to be available. */
+		overcommit_pool.num_spares++;
+		r = pthread_cond_wait(&overcommit_pool.spare_workers, &overcommit_pool.mutex);
+	}
+}
+
+
+static void*
+_dispatch_normal_worker_main(void *context DISPATCH_UNUSED)
```
MadCoder: I dislike having that much code duplicated, can't we do something where you have a generic `DISPATCH_ALWAYS_INLINE` function that takes a `bool overcommit` and handles the semantics differences inline, and then call this symbol from `_dispatch_normal_worker_main` and `_dispatch_overcommit_worker_main`? it will yield exactly the same code since the compiler will inline and constant fold the `overcommit` but it'll make maintenance much easier.

look at `_dispatch_queue_*_drain` for an example of that optimization technique.

src/workqueue/workqueue.c
MadCoder: We tend to instead have a `#define DISPATCH_WQ_DEBUG` that is set either inidividually or if `DISPATCH_DEBUG` is set so that it is compiled out by default.

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+	// should we output debug messages for pool scheduling/sizing decisions?
+	threadpool_verbose = getenv("LIBDISPATCH_THREADPOOL_VERBOSE") != NULL;
+
+#if DISPATCH_ENABLE_PWQ_KEXT
+	// Initialize kext subsystem.
+	kext_active = _dispatch_workqueue_management_init_kext();
+
+	if (kext_active) {
+		_dispatch_spawn_thread(dispatch_thread_monitor_main_kext);
+	} else
```
MadCoder: just do an early return and remove the weird `} else` and indent below.

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+static int
+_dispatch_worker_register(void)
+{
+	int rc;
+
+#if DISPATCH_ENABLE_PWQ_KEXT
+	if (kext_active) {
+		_dispatch_worker_register_kext();
+		rc = 0;
+	} else
```
MadCoder: ditto

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+	return rc;
+}
+
+static void
+_dispatch_worker_unregister(void)
+{
+#if DISPATCH_ENABLE_PWQ_KEXT
+	if (kext_active) {
+		_dispatch_worker_register_kext();
+	} else
```
MadCoder: ditto

MadCoder: (and two times more below)

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+			// Unable to open file.
+			// Must mean worker exited uncleanly (without executing _dispatch_worker_unregister())
+			// Clean up by removing pid and decrementing number of workers in pool
+			verbose_printf("Unable to open /proc/%d/stat; removing worker from monitoring list", worker_pid);
+			int last = scheduler.num_registered_workers-1;
+			scheduler.registered_workers[i] = scheduler.registered_workers[last];
+			scheduler.registered_workers[last] = 0;
+			scheduler.num_registered_workers--;
+		} else {
+			if (fcntl(fd, F_SETFL, O_NONBLOCK) == 0) {
```
MadCoder: doesn't linux support O_NONBLOCK as an argument to open() these days?

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,849 @@
...
+	r = _dispatch_pthread_sigmask(SIG_BLOCK, &mask, NULL);
+	(void)dispatch_assume_zero(r);
+
+	bool disabled = getenv("LIBDISPATCH_THREADPOOL_DISABLE_MANAGER") != NULL;
+	if (unlikely(disabled)) {
+		verbose_printf("user-level threadpool manager exiting");
+		return NULL;
+	}
+
+	while(true) {
```
MadCoder: we tend to use `for (;;) {` for this

src/workqueue/workqueue.c
MadCoder: given my naming remark, this should actually likely actually be `dispatch_workq_item_t`

MadCoder:
Actually I think re-reading myself I like `dispatch_workq_*` better than `dispatch_wq_`. That way, `workq` refer to the C module, and `workqueue` spelled out to a specific thread / execution context.

a `workq_item` is what a `workqueue` (thread) runs.

a `workq_pool` is the pool of work queues.

etc...

src/workqueue/workqueue_internal.h
```diff
@@ -0,0 +1,56 @@
...
+ * relying on these interfaces WILL break.
+ */
+
+#ifndef __DISPATCH_WORKQUEUE_INTERNAL__
+#define __DISPATCH_WORKQUEUE_INTERNAL__
+
+typedef struct _dispatch_workqueue * pthread_workqueue_t;
+
+/* Work queue priority attributes. */
+#define WORKQ_HIGH_PRIOQUEUE             0
```
MadCoder: I'm not 100% sure that's the right interface, I think we should build something on QoS instead of the legacy interface.

dgrove-oss:
I pushed a revision that tries to address most of the review comments.  The main things I know I didn't do in this revision are (a) dealing with the code duplication in the normal/overcommit worker main functions (b) the legacy API (6 priority, not QoS) and (c) some of the overuse of 'queues'.

Stepping back, I think the next big step after this PR is to try to move forward to an API where we avoid the extra level of indirection of having the workq level allocating its own workitems that are just wrapping a root_queue_drain function on a dq and have workq go after the root_queues more directly. If we can do that, than the duplicated code changes quite a bit (and maybe mostly disappears).  I can take a look in more depth, but I'm tempted to leave the duplication alone since I'm hoping that in a week or so all of that code is different anyways.   But, if you disagree I don't mind attempting a cleanup to cut down on the duplication tomorrow.

hassila:
> Integration of the workqueue implementation
into the dispatch code base simplifies the process of
evolving the APIs between the two layers and thus
prepares for future optimization and enhancements

Just a short comment based on some previous experience here (just as food for thought for such future optimisations and enhancements): it would be nice to let e.g. a class reserve the space needed for enqueueing it, to avoid potential memory allocations both in dispatch and in the work queue code - it would be helpful for certain interesting use cases.

Anyway, nice to see this WIP!

src/workqueue/workqueue.c
```diff
@@ -0,0 +1,833 @@
...
+	void *func_arg;
+} dispatch_workq_item_s, *dispatch_workq_item_t;
+
+typedef struct dispatch_workq_workqueue_s {
+	STAILQ_HEAD(,dispatch_workq_item_s) item_listhead;
+	dispatch_unfair_lock_s lock;
+	int priority;
+	bool overcommit;
+} dispatch_workq_workqueue_s, *dispatch_workq_workqueue_t;
+
```
dgrove-oss: I'm not thrilled with the name dispatch_workq_workqueue_t, but am at a bit of a loss for anything better.  Fortunately, this should be a temporary type.  If we can eliminate the dispatch_workq_item_t, we won't need this type either (we won't need to have queues of dispatch_workq_item_t's).

MadCoder: I would call this a `dispatch_workq_thread_context_t` or even `dispatch_workq_context_t`.

MadCoder: Or given you use "worker" to name threads below, `dispatch_worq_worker_context_t`

dgrove-oss: went with dispatch_workq_context_t for now (missed the edited comment).  

I don't expect to actually need this type in the long run.  For the initial merge, I was staying pretty close to the libpwq structure which has a redundant layer of queuing.  I was planning on eliminating that redundancy in a second round of changes.  We can talk on Monday about whether this plan makes sense or if we should jump to something closer to the final structure in one big step.

dgrove-oss:
I'll work on rebasing this PR to current master and updating to use dispatch_priority_t tomorrow.  Then we can figure out what whether it makes sense to merge something like this in the near future, or if we want to go directly to a more optimized version that eliminates the extra level of workitems/queuing that were present in libpwq and replicated here.

In either case, I don't think we should rush this into Swift 3.1.0 (maybe 3.1.1 if there is one).

MadCoder:
I think it's too late for Swift 3.1. My experience for a thread pool is that an evil you know is better than brand new code, we should though start right away for Swift 4. and if there's a swift 3.1+ then we can always backport it at that time.

dgrove-oss:
I did rebase this to post epoll-merge master today.   But, unless we are in a rush to eliminate the libpwq dependency, I think it probably makes sense for me to rework to non-legacy APIs and try to eliminate the extra level of workitems/queuing before you review further.

MadCoder:
Agreed.

MadCoder:
oh and btw, now that we have `src/event/*` the workqueue code should go into `src/event/workqueue` and have if needed a `src/workqueue_internal.h` header

the eventing and threading subsystem are quite tied together, and that puts everything that is strongly tied to the kernel in its own place.

`src/*.c` is really meant for the libdispatch API surface and pure userland constructs as much as possible.

dgrove-oss:
@MadCoder, I've done a rework to something that I think is more plausible.  Much less new code, and the extra level of queuing/workitems inherited from libpwq is gone. The oversubscription heuristics may still need tweaking to maximize performance especially at small core counts (I'm running some experiments over the weekend), but I think the approach/algorithms are ready for review when you have time.

The basic idea is to use the pthread_root_queue functionality already in libdispatch augmented with new code that periodically determines how many worker threads are actually runnable. If it looks like we are under target, then additional workers are gradually created to modestly oversubscribe the thread pool.

The main changes in queue.c are making dgq_thread_pool_size signed to enable using negative values to encode the degree of oversubscription and some other minor adjustments.

As written, it is only monitoring/oversubscribing the default non-overcommit queue.  But, if the approach is acceptable and we want to be able to oversubscribe the other non-overcommit queues the approach should generalize ok.

MadCoder:
@dgrove-oss sounds nice, I won't likely have the time to do the review it requires before mid-next week, but I'll try to skim through it over the week-end.

dgrove-oss:
@MadCoder, I got busy with a few other things for the last month, but I have time again for the next few weeks.  If you could find time to review, I'd really like to push through the libpwq removal work so we have plenty of time to shake out any problems before the next release. 

Thanks!

MadCoder:
I was very busy too but should have some bandwidth early next week, I'll have a closer look then

src/queue.c
```diff
@@ -3992,16 +4037,21 @@ _dispatch_global_queue_poke_slow(dispatch_queue_t dq, unsigned int n)
...
+#else
+		bool pool_full = t_count == 0;
+#endif
+		if (pool_full) {
 			_dispatch_root_queue_debug("pthread pool is full for root queue: "
 					"%p", dq);
 			return;
 		}
-		j = i > t_count ? t_count : i;
+		j = (int32_t)i > t_count ? t_count : (int32_t)i;
```
MadCoder: please avoid this cast by making `i` signed, which will mean fixing `_dispatch_global_queue_poke` to take an int too, all the way down to the root of that need which is `da_thr_cnt`. this is fine as dispatch really only can manage in the thousands of threads, so making this signed is not an issue, and that would avoid these ugly looking casts

src/queue.c
```diff
@@ -3943,6 +3955,39 @@ _dispatch_runloop_queue_poke(dispatch_queue_t dq, dispatch_qos_t qos,
...
+			if (r != EAGAIN) {
+				(void)dispatch_assume_zero(r);
+			}
+			_dispatch_temporary_resource_shortage();
+		}
+	} while (--n);
+}
+
+void
+_dispatch_pthread_root_queue_oversubscribe(dispatch_queue_t dq, int n)
```
MadCoder: (and now that I understand the point the name is actually fine)

src/event/workqueue.c
```diff
@@ -0,0 +1,287 @@
...
+#if DISPATCH_ENABLE_PWQ_KEXT
+	if (_dispatch_workq_kext_active) {
+		_dispatch_workq_worker_register_kext();
+		return;
+	}
+#endif
+	dispatch_workq_manager_t mgr = _dispatch_workq_get_default_manager();
+	int tid = syscall(SYS_gettid);
+	_dispatch_unfair_lock_lock(&mgr->registered_worker_lock);
+	for (int i=0; i<mgr->num_registered_workers; i++) {
```
MadCoder: It is even more critical that you will likely when adding feature to the pool want access to others of these TSDs anyway.

src/event/workqueue.c
```diff
@@ -0,0 +1,287 @@
...
+	dispatch_queue_t dq = _dispatch_workq_get_default_root_queue();
+	bool work_available = _dispatch_workq_root_queue_has_work(dq);
+
+	if (!work_available) {
+		_dispatch_debug("workq: %s is empty; doing nothing", dq->dq_label);
+		return;
+	}
+
+	_dispatch_workq_count_runnable_workers(mgr);
+	int32_t count = _dispatch_pthread_root_queue_thread_pool_size(dq);
```
MadCoder: (it goes with my general review remark about the fact you need to count threads in a pre-posted way anyway to avoid in-flight thread explosion)

dgrove-oss:
Thanks Pierre!   I'll work through the detailed comments; on first read I think I understand your points.

On the overall comments, I agree we need to have a regulator for all the priority buckets.  Since it looks like the overall approach is plausible, I'll go ahead and do that as part of the next revision. I was stalling on tackling all the buckets to make sure that trying to use dispatch's pthread_root_queue code wasn't going to be a non-starter. 

On overall points 2 & 3, I agree there are timing holes in the code where it could easily create extra threads (or be slow by 1 timer invocation of the monitor to notice that the system was totally idle due to blocked workers and needed to be kicked).  The libpwq code did do the more complicated in-flight accounting you described to contain thread explosions.   I should be able to do something similar.

MadCoder:
FWIW for the regulators, overcommit don't need one, the root queue implementation as it is is "good enough". the regulator is only really meaningful for non overcommit buckets.

*however* the overcommit workqueue threads may want to be accounted for globally to have a global thread limit (like Darwin has) that spans several buckets.

Our belief today TBH is that the limit for threads and concurrency should encompass any thread that is at a higher priority. Meaning that if you already are using NCPU runnable threads at the highest priority, other non overcommit buckets should have nothing.

The current approach you're taking doesn't allow to do that, but I wouldn't try to do it at the first try either, I'm just explaining what we believe to be the right direction overall should it influence some choices you're taking.

src/event/workqueue_internal.h
```diff
@@ -0,0 +1,53 @@
...
+#ifndef __DISPATCH_WORKQUEUE_INTERNAL__
+#define __DISPATCH_WORKQUEUE_INTERNAL__
+
+/* Work queue priority attributes. */
+#define WORKQ_HIGH_PRIOQUEUE             0
+#define WORKQ_DEFAULT_PRIOQUEUE          1
+#define WORKQ_LOW_PRIOQUEUE              2
+#define WORKQ_BG_PRIOQUEUE               3
+#define WORKQ_BG_PRIOQUEUE_CONDITIONAL   4
+#define WORKQ_HIGH_PRIOQUEUE_CONDITIONAL 5
```
dgrove-oss: They are here to match the legacy configuration that libpwq provided; code in queue.c is expecting them to be defined.  I agree we should move away from all of these to QoS (and that ideally we would do that as pretty much the next thing after this gets merged).

dgrove-oss:
I've made a pass through and handled almost all of the review comments.  The open items are:

1. Adding the counting of pending thread creates to avoid an explosion of thread creations.  I had a "plumbing" question for this.  One natural way to do this would be to add a pending creations count to the dispatch_root_queue_context_s when HAVE_INTERNAL_PTHREAD_WORKQUEUE is defined. The thing that is slightly awkward is that this type is only visible inside of queue.c.  Should I go ahead and add the counter there and then expose a utility method to workqueue to check to see if there are pending thread creations on the monitored queue, or can you think of a better way to structure this? 

2. Undoing the oversubscription of a pool when blocked threads come back. (ie, backing off when it looks like we have way too many runnable workers).  libpwq handled this by injecting a synthetic workitem into the queue that caused the thread that got it to return from a routine that was similar in function to _dispatch_root_queue_drain_one.  This gave the regulator a chance to tell the thread to suspend itself without adding logic to the hot path.   Right now, there is no backoff in the code (as long as the queue is non-empty, all of the workers keep chugging along).   Should I pursue something like the libpwq approach of injecting a workitem into the overloaded queue, or do you have another approach you think would be better?

Thanks!

dgrove-oss:
I want to test more thoroughly on Monday, but I realized the pending thread logic was actually already in place for the pthread_root_queues.  I just needed to add the right cmpxchg on dgq_pending in the new oversubscribe routine and it should just work.

src/event/workqueue_internal.h
MadCoder: yeah that's kind of what I realized, let's not make the change bigger than required.

MadCoder:
I think that the pthread root queues should have all it takes, and if we need more, we can certainly move the extra struct definition that is stuck at the end of a dispatch_queue_s to workqueue_internal.h if that helps with things, to me "workqueue" is "below" src/queue.c in the same way event/event* is below source.c, so having workqueue.c call into queue.c is inverted layering.

However what you're doing so far feels ok to me.

I had a hunch that dgq_pending has what you need, and I expect that when you give back threads if you're above the original limit you can just cap the counter back.

IOW if the limit was 10, and you oversubscribed by say 5, so you have 15 threads at the moment, as they go back, eventually the count will become 10, then 11, then ... then 15, well you want to drop anything that's above the limit of 10 (which may need you to keep a copy of that original limit if it doesn't already exist as a thing).

hassila:
Just a short comment on 2 for your consideration - not regarding the mechanism used for removing the threads, but rather the decision logic on when to remove them and how quickly. 

The implementation we did in libpwq used hysteresis for the backoff of threads to avoid degenerate behaviour when you have workloads that ramp up/down the thread count within fairly short time spans. 

We just had a simple aggregation of the number of threads that were oversubscribed per time unit and had a heuristic of "x idle-thread-seconds" accumulated required for backoff of one thread- in practice it meant that if you had many extra threads idling, we would more aggressively deallocate them in the beginning and then tapering off as we got closer to the desired number.

Essentially a similar model used for heat system regulation (accumulating the delta between actual and desired temperatures over time and taking action due to certain thresholds).

This approach essentially smoothed out the thread creation/deallocation problems we observed for certain workloads (initially we just naively deallocated threads immediately when there was a surplus) and it proved quite good in practice - hope the explanation makes sense.

dgrove-oss:
@hassila thanks for the description of libpwq's control strategy; makes sense

dgrove-oss:
@MadCoder I've reworked the code and its ready for you to take another look.   

In addition to the localized fixes, I've made some adjustments to how the workq_monitor system interacts with the global queues to address the oversubscription/pending thread create issue.  As a result, the delta got smaller and I think easier to understand/maintain.   The key change is that I added a 'floor' parameter to _dispatch_global_queue_poke enabling it to be used both for "normal" poke operations (floor of 0) and for oversubscription requests (floor < 0) thus funneling all the thread creation requests through a single code path. 

Right now, the only mechanism for reducing the thread pool size is the 5 second timeout on idle threads.  Looking through the code in queue.c, I was wondering if a possible approach would be to adapt the HAVE_PTHREAD_WORKQUEUE_NARROWING path to also work on Linux with the workq_monitor (if the monitor sees overload, it would post a narrow request)?   Unless you think it is critical to get this support into the initial changes, I'd be inclined to handle this in a follow up change.

dgrove-oss:
@MadCoder, would be great if you could take a look at the current proposed changes this week and let me know if there is more that is needed before it can be tested/merged.  I would like to get this (and ideally the followup changes to modernize priority/QoS API on Linux) into the Swift 4 beta.  Thanks!

MadCoder:
I don't think it's urgent to handle the HAVE_PTHREAD_WORKQUEUE_NARROWING case for now

src/event/workqueue.c
```diff
@@ -0,0 +1,258 @@
...
+ *
+ * @APPLE_APACHE_LICENSE_HEADER_END@
+ */
+
+#include "internal.h"
+
+#if HAVE_INTERNAL_PTHREAD_WORKQUEUE
+
+// forward looking typedef; not yet defined in dispatch
+typedef pid_t dispatch_tid;
```
MadCoder: this requires C11, and is defined in lock.h maybe we should reorder lock.h wrt workqueue_internal.h (?)

MadCoder: I mean this is weird that you see dispatch_unfair_lock_s below but not that, something is off

src/event/workqueue.c
```diff
@@ -0,0 +1,258 @@
...
+	/* The desired number of runnable worker threads */
+	int32_t target_runnable;
+
+	/*
+	 * Tracking of registered workers; all accesses must hold lock.
+	 * Invariant: registered_tids[0]...registered_tids[num_registered_tids-1]
+	 *   contain the dispatch_tids of the worker threads we are monitoring.
+	 */
+	dispatch_unfair_lock_s registered_tid_lock;
+	dispatch_tid *registered_tids;
```
MadCoder: I still believe this should point to the TSD struct of threads (which has the tid inside) and have the TSD struct know about the index it is at in the array of registered threads, that will help you for things such as narrowing or similar in the future.

src/event/workqueue.c
```diff
@@ -0,0 +1,258 @@
...
+			_dispatch_debug("workq: %s under utilization target; poking with floor %d",
+					dq->dq_label, floor);
+			_dispatch_global_queue_poke(dq, 1, floor);
+		}
+	}
+}
+
+// temporary until we switch over to QoS based interface.
+static dispatch_queue_t
+get_root_queue_from_legacy_priority(int priority) {
```
MadCoder: bracket on the next line please

src/event/workqueue.c
```diff
@@ -0,0 +1,258 @@
...
+		mon->target_runnable = target_runnable;
+	}
+
+	// Create monitoring timer that will periodically run on dispatch_mgr_q
+	dispatch_source_t ds = dispatch_source_create(DISPATCH_SOURCE_TYPE_TIMER,
+			0, 0, &_dispatch_mgr_q);
+	dispatch_source_set_timer(ds, dispatch_time(DISPATCH_TIME_NOW, 0),
+			NSEC_PER_SEC, 0);
+	dispatch_source_set_event_handler_f(ds, _dispatch_workq_monitor_pools);
+	dispatch_activate(ds);
```
MadCoder: this timer will appear as leaked, because the timer_refs doesn't back reference it. the trick we use to not have that happen is to do a `dispatch_set_context(ds, ds)`

src/queue.c
```diff
@@ -672,23 +677,30 @@ _dispatch_root_queues_init_workq(int *wq_supported)
...
 	if (slowpath(pool_size) && pool_size < thread_pool_size) {
 		thread_pool_size = pool_size;
 	}
 	qc->dgq_thread_pool_size = thread_pool_size;
+	if (qc->dgq_kworkqueue == (void*)(~0ul)) {
+		// when _dispatch_root_queues_init is invoked from
+		// _dispatch_global_queue_poke_slow we need to clear
+		// dgq_pending here because qc->dgq_kworkqueue wasn't
+		// initialized to (~0ul) when we went through _dispatch_global_queue_poke
+		qc->dgq_pending = 0;
```
MadCoder: I would do this in `_dispatch_root_queues_init_workq` when we do set it to `~0ull` explaining:

```C
// because the fastpath of _dispatch_global_queue_poke didn't know yet
// that we're using the internal pool implementation we have to undo it setting dgq_pending
```

That would make more sense to me

src/queue.c
```diff
@@ -1979,8 +1991,9 @@ _dispatch_pthread_root_queue_create(const char *label, unsigned long flags,
 	dispatch_pthread_root_queue_context_t pqc;
 	dispatch_queue_flags_t dqf = 0;
 	size_t dqs;
-	uint8_t pool_size = flags & _DISPATCH_PTHREAD_ROOT_QUEUE_FLAG_POOL_SIZE ?
-			(uint8_t)(flags & ~_DISPATCH_PTHREAD_ROOT_QUEUE_FLAG_POOL_SIZE) : 0;
+	uint8_t pool_size_u = flags & _DISPATCH_PTHREAD_ROOT_QUEUE_FLAG_POOL_SIZE ?
+			(int8_t)(flags & ~_DISPATCH_PTHREAD_ROOT_QUEUE_FLAG_POOL_SIZE) : 0;
+	int32_t pool_size = (int32_t)pool_size_u;
```
MadCoder: why not directly:

```C
	int32_t pool_size = flags & _DISPATCH_PTHREAD_ROOT_QUEUE_FLAG_POOL_SIZE ?
			(int8_t)(flags & ~_DISPATCH_PTHREAD_ROOT_QUEUE_FLAG_POOL_SIZE) : 0;
```

It should not warn because the RHS is a `uint8_t` that promotes to `int` just fine

src/queue.c
```diff
@@ -5458,6 +5491,9 @@ _dispatch_worker_thread(void *context)
 	dispatch_root_queue_context_t qc = dq->do_ctxt;
 	dispatch_pthread_root_queue_context_t pqc = qc->dgq_ctxt;
 
+	int pending = (int)os_atomic_dec2o(qc, dgq_pending, relaxed);
```
MadCoder: I know it's pre-existing but dgq_pending should be made an int to avoid this tasteless cast.

src/queue.c
```diff
@@ -5458,6 +5491,9 @@ _dispatch_worker_thread(void *context)
...
+	dispatch_assert(pending >= 0);
```
MadCoder: And here I would make this a hard crash with a

```C
if (unlikely(pending < 0)) {
    DISPATCH_INTERNAL_CRASH(pending, "Pending thread request underflow");
}
```

src/queue.c
```diff
@@ -3945,10 +3958,10 @@ _dispatch_runloop_queue_poke(dispatch_queue_t dq, dispatch_qos_t qos,
 
 DISPATCH_NOINLINE
 static void
-_dispatch_global_queue_poke_slow(dispatch_queue_t dq, unsigned int n)
+_dispatch_global_queue_poke_slow(dispatch_queue_t dq, int n, int floor)
```
MadCoder: @das this is where the part I'd like you to have a look at starts, to the end of the file. For Darwin `floor` will always be 0. For Linux it can be a negative value that allows to oversubscribe the pool further.

src/event/workqueue.c
dgrove-oss: Perhaps the internal and open source versions of lock.h are out of synch?  

The only typedef for pid_t in the current opensource version is https://github.com/apple/swift-corelibs-libdispatch/blob/master/src/shims/lock.h#L99 which defines dispatch_lock_owner as another name for pid_t.

src/event/workqueue.c
dgrove-oss: I can easily believe this is the right thing to do eventually.  Since the only time the current code needs to actually map from a thread's tid to its index in this table is in dispatch_worker_unregister (when the thread is exiting and we need to remove its tid), it didn't seem worth implementing this yet.

src/event/workqueue.c
MadCoder: oooh right we haven't synced that yet, fine do it like that and I'll remove the typedef when we merge next time.

src/event/workqueue.c
MadCoder: fair enough

dgrove-oss:
@MadCoder, I was experimenting with options for removing libpwq entirely.  This commit https://github.com/dgrove-oss/swift-corelibs-libdispatch/commit/626f8a891986f937e0b62df7bc0f23c8541b8b55 is one possibility. 

The idea is if you aren't on Darwin (no pthread_workqueue) and you don't have dispatch_workq_monitor functionality implemented (currently all non-Linux platforms) then fall back to pthread root_queues where every queue is treated as an overcommit queue.

The other obvious alternative would be to leave in support for libpwq, but I was hoping to reduce the number of pthread-workqueue related configurations.

Any thoughts?  Should I merge the commit into this PR, do as a followup PR, take another approach?

dgrove-oss:
@MadCoder I've squashed down to a single commit to prepare for merge.  Any feedback from @das on the changes in _dispatch_global_queue_poke_slow?   

Do you want me to include https://github.com/dgrove-oss/swift-corelibs-libdispatch/commit/626f8a891986f937e0b62df7bc0f23c8541b8b55 in this PR, do that change in a follow up PR, or do something different?

Hoping we are close; really want to get this + QoS cleanups into Swift 4 beta.

das:
I'll take a look today

dgrove-oss:
@das  ping.  Thanks.

dgrove-oss:
@das Please review the queue.c changes so we can progress this change.  Thanks.

dgrove-oss:
@swift-ci please test

dgrove-oss:
@das - any feedback on the changes to queue.c?
@MadCoder - could you request the CI infrastructure to test the PR?  I tried, but I don't have the right privileges to request a CI test.

configure.ac
```diff
@@ -320,21 +320,41 @@ AS_IF([test -n "$apple_libpthread_source_path" -a -n "$apple_xnu_source_osfmk_pa
...
+  [case $target_os in
+      linux*)
+        enable_internal_libpwq=yes
+	    ;;
+      *)
+        enable_internal_libpwq=no
+   esac]
+)
+AS_IF([test "x$enable_internal_libpwq" = "xyes"],
+  [AC_DEFINE(HAVE_INTERNAL_PTHREAD_WORKQUEUE, 1, [Use libdispatch's own implementation of pthread_workqueue API])
```
das: IMO we should call this DISPATCH_USE_INTERNAL_WORKQUEUE since it isn't a pthread thing anymore (which is a different library altogether on Darwin at least), and also not a HAVE check (which is generally about presence of interfaces external to the project)

configure.ac
```diff
@@ -320,21 +320,41 @@ AS_IF([test -n "$apple_libpthread_source_path" -a -n "$apple_xnu_source_osfmk_pa
...
-   )]
+     [build_own_pthread_workqueues=false
+      AC_CHECK_HEADERS([pthread/workqueue_private.h pthread_workqueue.h],
+        [AC_DEFINE(HAVE_PTHREAD_WORKQUEUES, 1, [Define if pthread work queues are present])
+         have_pthread_workqueues=true],
+        [have_pthread_workqueues=false]
+      )]
+  )]
 )
+AM_CONDITIONAL(HAVE_INTERNAL_PTHREAD_WORKQUEUES, $have_internal_pthread_workqueues)
```
das: DISPATCH_USE_INTERNAL_WORKQUEUE as well ? (also avoids the confusing singular/plural difference between the define and the conditional)

src/queue.c
```diff
@@ -32,6 +32,7 @@
 #endif
 #if HAVE_PTHREAD_WORKQUEUES && (!HAVE_PTHREAD_WORKQUEUE_QOS || DISPATCH_DEBUG) && \
 		!HAVE_PTHREAD_WORKQUEUE_SETDISPATCH_NP && \
+		!HAVE_INTERNAL_PTHREAD_WORKQUEUE && \
```
das: I'm not very clear on why HAVE_PTHREAD_WORKQUEUES is true for the internal workqueue case, that doesn't seem quite right, that define is intended to guard the use of the specific pthread_workqueue API being used (which the internal version doesn't emulate). It would be easier to understand if the cases where you need something that is guarded by HAVE_PTHREAD_WORKQUEUES just added a || HAVE_INTERNAL_PTHREAD_WORKQUEUE instead (and then you wouldn't need the confusing addition here)

src/queue.c
```diff
@@ -635,7 +639,8 @@ _dispatch_root_queues_init_workq(int *wq_supported)
 		result = !r;
 	}
 #endif // HAVE_PTHREAD_WORKQUEUE_SETDISPATCH_NP
-#if DISPATCH_USE_LEGACY_WORKQUEUE_FALLBACK || DISPATCH_USE_PTHREAD_POOL
+#if DISPATCH_USE_LEGACY_WORKQUEUE_FALLBACK || DISPATCH_USE_PTHREAD_POOL || \
+		HAVE_INTERNAL_PTHREAD_WORKQUEUE
```
das: shouldn't be necessary, you already have DISPATCH_USE_PTHREAD_POOL defined

src/queue.c
das: resp DISPATCH_USE_INTERNAL_WORKQUEUE as discussed elsewhere

src/queue.c
```diff
@@ -664,6 +669,10 @@ _dispatch_root_queues_init_workq(int *wq_supported)
 			}
 #endif // DISPATCH_USE_LEGACY_WORKQUEUE_FALLBACK
 			qc->dgq_kworkqueue = pwq ? pwq : (void*)(~0ul);
+			// because the fastpath of _dispatch_global_queue_poke didn't
+			// know yet that we're using the internal pool implementation
+			// we have to undo its setting of dgq_pending
+			qc->dgq_pending = 0;
```
das: don't you want to do this only if `!pwq` ? (to match the check in _dispatch_global_queue_poke). Otherwise you'll wipe out the pending info for the legacy fallback above

src/queue.c
```diff
@@ -5475,6 +5509,15 @@ _dispatch_worker_thread(void *context)
 	(void)dispatch_assume_zero(r);
 	_dispatch_introspection_thread_add();
 
+#if HAVE_INTERNAL_PTHREAD_WORKQUEUE
+	bool overcommit = qc->dgq_wq_options & WORKQ_ADDTHREADS_OPTION_OVERCOMMIT;
+	bool manager = dq == &_dispatch_mgr_root_queue;
```
das: please add ()

das:
a few minor comments but overall the additions to the existing code look great to me!

src/queue.c
das: this is really the only one of my comments that requires fixing, the rest of my comments are more about fitting in better with the existing configuration/define scheme and some style nits that we can also address later if necessary

dgrove-oss:
Thanks!  I guarded the clearing of qc->dgq_pending, renamed to DISPATCH_USE_INTERNAL_WORKQUEUE, and added () at lines 5513 and 5514.   Squashed back into single commit.

I'd like to clean up  HAVE_PTHREAD_WORKQUEUES being defined when using the internal workqueues in a follow up PR.   I think it would make sense to look at this and switching to more of a QoS based priority definition (Pierre's comment about the priority levels being defined in workqueue_internal.h) as a single change.

das:
looks great, thanks!

das:
@swift-ci please test

MadCoder:
@swift-ci please test

dgrove-oss:
Looks like the build got into a confused state about whether or not libpwq was included or not.  I'm going to try pulling in https://github.com/dgrove-oss/swift-corelibs-libdispatch/commit/626f8a891986f937e0b62df7bc0f23c8541b8b55 which removes libpwq entirely and see if that is enough.  If it isn't, we might need to figure out how to force a clean workspace on the build machines.

dgrove-oss:
@swift-ci please test

MadCoder:
@swift-ci please test

dgrove-oss:
Yeah!  Build worked.  Do you want me to squish to 1 commit, or leave as two?

MadCoder:
please squash it all and I'll merge

dgrove-oss:
squashed; ready to merge.  Thanks!

das:
this causes build failures on Darwin due to numerous unsigned vs signed build warnings (which we have enabled as fatal) from the uint32_t -> int32_t change.
I am fixing as part of an upcoming source drop

PR #203, opened 2017-01-24
update libpwq and libkqueue submodule versions

dgrove-oss:

PR #207, opened 2017-01-30
SR-3771 and SR-3751: libdispatch should not depend on external libkqueue

dgrove-oss:
Revert back to libkqueue version that has the fix for
SR-2941 but does not have the changes to remove the
autotools-based build system.


dgrove-oss:
@swift-ci Please test

dgrove-oss:
@MadCoder would you mind merging this to master?  I screwed up in #203 and moved libkqueue too far forward.  I didn't notice (and neither did the CI testing) that because the latest libkqueue removed the autotools based build system it meant that we stopped building libkqueue from source when building libdispatch.  On both my machine and the Swift CI machines, libkqueue is installed, so the build just defaults back to using the system libkqueue which is not what we want.

Thanks & sorry for the churn.

PR #208, opened 2017-02-02
Fixes crash when DispatchData is created from an UnsafeBufferPointer<UInt8> with a nil address (Radar 29337927)

ktopley-apple:


ktopley-apple:
@swift-ci please test

PR #209, opened 2017-02-06
fix libdispatch.a to include Swift overlay symbols

weissi:
The static version (`libdispatch.a`) that can be compiled using
`./configure --enable static=yes` was missing all the Swift overlay symbols
(everything in `swift_overlay.o`). The reason for that is that the linker is
invoked through libtool which wants `.lo` files but the Swift overlay got passed
as a `.o` file. This first of all leads to this warning:

    *** Warning: Linking the shared library libdispatch.la against the non-libtool
    *** objects [...]/swift_overlay.o is not portable!

And the result is that libtool doesn't include `swift_overlay.o` when putting
together the static library `libdispatch.a`.

This patch fixes this problem is a pretty crude way. The real problem is that
libtool doesn't understand Swift. So it can't be used when compiling
`swift_overlay.o`. In order to still get a `.lo` file this patch tricks libtool
into generating one from the `swift_overlay.o` generated by `swiftc`. It's very
hacky but I'm not sure what else to do.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -148,11 +148,21 @@ SWIFT_SRC_FILES=\
 
 SWIFT_ABS_SRC_FILES = $(SWIFT_SRC_FILES:%=$(abs_srcdir)/%)
 SWIFT_OBJ_FILES = $(abs_builddir)/swift/swift_overlay.o
+SWIFT_LIBTOOL_OBJ_FILES = $(abs_builddir)/swift/swift_overlay.lo
 
 SWIFTC_FLAGS+= -Xcc -fmodule-map-file=$(abs_top_srcdir)/dispatch/module.modulemap -I$(abs_top_srcdir) -Xcc -fblocks
 if DISPATCH_ENABLE_OPTIMIZATION
 SWIFTC_FLAGS+=-O
 endif
+
+# this saves the object file, then tricks libtool into generating a .lo file and
+# then moves the object file back in the places libtool expects them to be for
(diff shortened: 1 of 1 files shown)


src/Makefile.am
```diff
@@ -148,12 +148,22 @@ SWIFT_SRC_FILES=\
...
 if DISPATCH_ENABLE_OPTIMIZATION
 SWIFTC_FLAGS+=-O
 endif
 
+# this saves the object file, then tricks libtool into generating a .lo file and
+# then moves the object file back in the places libtool expects them to be for
+# the PIC and non-PIC case.
+$(abs_builddir)/swift/swift_overlay.lo:	$(abs_builddir)/swift/swift_overlay.o
+	mv $(abs_builddir)/swift/swift_overlay.o $(abs_builddir)/swift/.libs/swift_overlay.o.save
+	$(LIBTOOL) --mode=compile --tag=CC true -o $< -c foo.c
```
dgrove-oss: I'd suggest using /dev/null instead of foo.c to make it clearer that your compilation command ('true') doesn't actually care about its input file you are giving it.

weissi: 👍 thanks, will fix right now

dgrove-oss:
looks plausible to me;  it's a hack, but it is better than what we were doing. @MadCoder any thoughts?

alblue:
@swift-ci please test

MadCoder:
this is so gross. but I guess we need to be unblocked until we get a better build system

PR #212, opened 2017-02-12
Check if in same queue to prevent a deadlock

amraboelela:


MadCoder:
1. this patch is bogus, different queues can have the same name
2. libdispatch already asserts if you try to deadlock in this obvious way helping you from staying deadlocked that way
3. you avoid this deadlock as a client by remembering which locks you already have, like you would do with a pthread mutex.

MadCoder:
The code doing this already is here:

https://github.com/apple/swift-corelibs-libdispatch/blob/master/src/queue.c#L3641-L3645

PR #216, opened 2017-02-22
Linux fixes for dispatch-806 merge

dgrove-oss:
Collection of small fixes to dispatch-806 merge to
allow code to compile/link/run on Linux.


tests/dispatch_test.c
```diff
@@ -60,6 +64,10 @@ dispatch_test_check_evfilt_read_for_fd(int fd)
 	int r = kevent(kq, &ke, 1, &ke, 1, &t);
 	close(kq);
 	return r > 0;
+#else
+	// TODO: Need to write a real check for epoll-backend here
+	return true;
+#endif
```
dgrove-oss: Not intending this as the final fix for this code; just an expedient hack to allow the merge to proceed.

src/event/event_epoll.c
```diff
@@ -194,7 +202,7 @@ _dispatch_unote_register(dispatch_unote_t du, dispatch_priority_t pri)
 	case DISPATCH_EVFILT_CUSTOM_ADD:
 	case DISPATCH_EVFILT_CUSTOM_OR:
 	case DISPATCH_EVFILT_CUSTOM_REPLACE:
-		du._du->du_wlh = DISPATCH_WLH_GLOBAL;
+		du._du->du_wlh = wlh;
```
MadCoder: this diff is wrong, it should always use DISPATCH_WLH_GLOBAL

MadCoder:
everything makes perfect sense to me but for the commented hunk.

src/event/event_epoll.c
dgrove-oss: ok;  I think I should then add an assert that the value of the wlh parameter is DISPATCH_WLH_GLOBAL.  I don't like ignoring the value the caller passed in even without asserting it has a known value.

MadCoder: no it needs to be GLOBAL here, event_kevent.c does the same. it is meant to be that.

dgrove-oss: Maybe we're looking at different pieces of code.  In event_kevent.c, I think the matching piece of code is:
`bool
_dispatch_unote_register(dispatch_unote_t du, dispatch_wlh_t wlh,
		dispatch_priority_t pri)
{
	dispatch_assert(!_dispatch_unote_registered(du));
	du._du->du_priority = pri;
	switch (du._du->du_filter) {
	case DISPATCH_EVFILT_CUSTOM_ADD:
	case DISPATCH_EVFILT_CUSTOM_OR:
	case DISPATCH_EVFILT_CUSTOM_REPLACE:
		du._du->du_wlh = wlh;
		return true;
	}
	if (!du._du->du_is_direct) {
		return _dispatch_unote_register_muxed(du, DISPATCH_WLH_GLOBAL);
	}
	return _dispatch_kq_unote_update(wlh, du, EV_ADD | EV_ENABLE);
}
`

src/event/event_config.h
```diff
@@ -133,7 +133,7 @@
 #	define EV_ONESHOT				0x0010
 #	define EV_CLEAR					0x0020
 #	define EV_DISPATCH				0x0080
-#	define EV_UDATA_SPECIFIC		0x0100
+#	define EV_UDATA_SPECIFIC			0x0000
```
das: should this really be defined at all ? I would have thought the `#ifdef EV_UDATA_SPECIFIC` below would have done the right thing if it is not

src/event/event_epoll.c
das: no, event_kevent.c also sets it to 'wlh'. I think these are arguably both wrong and should both be the GLOBAL constant (but it also doesn't really matter for this 'synthetic event' case)

src/event/event_config.h
dgrove-oss: My recollection is that @MadCoder preferred us to define it as 0x0000 instead of leaving it undefined.  I believe it is a matter of style; I think either leaving it undefined or defining it to be 0x0000 accomplish the same thing.

das: the conditional below defines it to 0x0000 already, that is what I am talking about

das: i.e.

```
#ifdef EV_UDATA_SPECIFIC
#	define DISPATCH_EV_DIRECT		(EV_UDATA_SPECIFIC|EV_DISPATCH)
#else
#	define DISPATCH_EV_DIRECT		0x0000
#	define EV_UDATA_SPECIFIC		0x0000
#	undef  EV_VANISHED
#	define EV_VANISHED				0x0000
#endif
```

that is there exactly for the cases where EV_UDATA_SPECIFIC isn't present (in both cases of backend, we also support the absence of this flag with kevent)

das: (and in particular doing it that way will also cause EV_VANISHED to be 0x0000 on Linux. currently it is 0x0200 via the #define at the top of the file)

dgrove-oss: ok.  makes sense.

I changed to not define EV_UDATA_SPECIFIC on Linux and squashed/rebased.

MadCoder:
Indeed, seems I misremembered, this is fine :)

src/event/event_config.h
das: cool, lgtm, thanks!

PR #221, opened 2017-02-22
remove unused dependency on linux/membarrier.h for intel to fix CI build

das:
add dependency on linux-libc-dev to INSTALL.md for other architectures

### DIFF
--- a/src/shims/lock.h
+++ b/src/shims/lock.h
@@ -89,7 +89,9 @@ _dispatch_lock_has_failed_trylock(dispatch_lock lock_value)
 
 #elif defined(__linux__)
 #include <linux/futex.h>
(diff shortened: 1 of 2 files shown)


src/shims/lock.h
```diff
@@ -89,7 +89,9 @@ _dispatch_lock_has_failed_trylock(dispatch_lock lock_value)
 
 #elif defined(__linux__)
 #include <linux/futex.h>
+#if !defined(__x86_64) && !defined(__i386__)
```
MadCoder: I think it should be `__x86_64__` not `__x86_64`

das: yup, typed too quickly

PR #223, opened 2017-02-23
queue_specific_queue is missing its do_push vtable entry

MadCoder:
Also reorder all vtable entries to be in the same order (push, invoke,
wakeup).

Reenable dispatch_context_for_key.

SR: SR-4039


MadCoder:
@swift-ci please test

src/init.c
```diff
@@ -406,8 +407,8 @@ DISPATCH_VTABLE_SUBCLASS_INSTANCE(queue_main, queue,
 	.do_type = DISPATCH_QUEUE_SERIAL_TYPE,
 	.do_kind = "main-queue",
 	.do_dispose = _dispatch_queue_dispose,
-	.do_invoke = _dispatch_queue_invoke,
 	.do_push = _dispatch_queue_push,
+	.do_invoke = _dispatch_queue_invoke,
```
das: FWIW push -> wakeup -> invoke would have seemed the natural order to me ;-)

MadCoder: I just kept the order that was dominating. For real the better order is:

wakeup -> push -> invoke. (cause it's in that order it happens)

MadCoder: actually no I'm tired, it's push on self, wakeup onto target, then invoke you're correct :P

PR #224, opened 2017-02-23
SR-4026: dispatch_test_check_evfilt_read_for_fd for epoll

dgrove-oss:
Implement dispatch_test_check_evfilt_read_for_fd test method
for the event_epoll backend.

### DIFF
--- a/tests/dispatch_test.c
+++ b/tests/dispatch_test.c
@@ -31,6 +31,8 @@
 #if __has_include(<sys/event.h>)
 #define HAS_SYS_EVENT_H 1
 #include <sys/event.h>
+#else
+#include <sys/poll.h>
 #endif
 #include <assert.h>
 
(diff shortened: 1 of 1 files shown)


MadCoder:
that seems overly complicated, that thing only checks whether a given fd is readable whcih you can do with a very simple poll() call, you don't need epoll at all.

```C
struct pollfd pfd = {
    .fd = fd,
    .events = EPOLLIN,
};
int rc;
do {
    rc = poll(&pfd, 1, 0);
} while (rc == -1 && errno == EINTR);
return rc == 1
```

should be it I think. we don't do that on Darwin because poll() actually is heavyweight and uses a short-lived kqueue.

MadCoder:
you may want to abort if poll returns -1 and the error isn't EINTR btw.

dgrove-oss:
Thanks; that is much more sensible.  I got distracted by trying to use epoll, but you are right there is no reason to do it that way.

PR #232, opened 2017-03-17
Makes the DispatchIO initializer that accepts a path failable, reflec…

ktopley-apple:
…ting the fact that a relative or non-existent path is invalid.

(Radar 31115994)


ktopley-apple:
@swift-ci please test

PR #233, opened 2017-03-31
internal: Move shims include so DISPATCH_INTERNAL_CRASH is defined

amboar:
Fixes the build on powerpc64le under Linux.

Signed-off-by: Andrew Jeffery <andrew@aj.id.au>

### DIFF
--- a/src/shims/lock.h
+++ b/src/shims/lock.h
@@ -547,7 +547,17 @@ _dispatch_once_xchg_done(dispatch_once_t *pred)
 	return os_atomic_xchg(pred, DLOCK_ONCE_DONE, release);
 #elif defined(__linux__)
 	if (unlikely(syscall(__NR_membarrier, MEMBARRIER_CMD_SHARED, 0) < 0)) {
-		DISPATCH_INTERNAL_CRASH(errno, "sys_membarrier not supported");
+		/*
+		 * sys_membarrier not supported
+		 *
+		 * Ideally we would call DISPATCH_INTERNAL_CRASH() here, but
+		 * due to ordering constraints in internal.h required by Darwin
+		 * the macro is undefined when this header is included.
+		 * Instead, open-code what would be a call to
+		 * _dispatch_hardware_crash() inside DISPATCH_INTERNAL_CRASH().
+		 */
+		__asm__("");
+		__builtin_trap();
 	}
 	return os_atomic_xchg(pred, DLOCK_ONCE_DONE, relaxed);
 #else


amboar:
@shahmishal Tagging you so you're aware for the PPC64LE Jenkins job.

das:
can you explain more clearly what the problem is  & for what use DISPATCH_INTERNAL_CRASH needs to be defined ?
the existing ordering here is intentional and your proposed re-ordering will not work as it breaks the build on Darwin

amboar:
@das yes, this is the build issue I'm seeing on powerpc64le under Linux:

```
...
/bin/bash ../libtool  --tag=CXX   --mode=compile /home/ubuntu/swift-dev/build/buildbot_incremental/llvm-linux-powerpc64le/bin/clang++ -DHAVE_CONFIG_H -I. -I../config  -I.. -I.. -I../private -DDISPATCH_USE_DTRACE=0 -I../libpwq/include -Wall -fvisibility=hidden -momit-leaf-frame-pointer  -isystem /usr/include/bsd -DLIBBSD_OVERLAY  -fblocks -I../src/BlocksRuntime -std=gnu++11 -fno-exceptions -O2 -c -o libdispatch_la-block.lo `test -f 'block.cpp' || echo './'`block.cpp
libtool: compile:  /home/ubuntu/swift-dev/build/buildbot_incremental/llvm-linux-powerpc64le/bin/clang++ -DHAVE_CONFIG_H -I. -I../config -I.. -I.. -I../private -DDISPATCH_USE_DTRACE=0 -I../libpwq/include -Wall -fvisibility=hidden -momit-leaf-frame-pointer -isystem /usr/include/bsd -DLIBBSD_OVERLAY -fblocks -I../src/BlocksRuntime -std=gnu++11 -fno-exce
ptions -O2 -c block.cpp  -fPIC -DPIC -o .libs/libdispatch_la-block.o
In file included from block.cpp:32:
In file included from ./internal.h:628:
In file included from ./shims.h:171:
./shims/lock.h:550:3: error: use of undeclared identifier 'DISPATCH_INTERNAL_CRASH'
                DISPATCH_INTERNAL_CRASH(errno, "sys_membarrier not supported");
                ^
1 error generated.
Makefile:701: recipe for target 'libdispatch_la-block.lo' failed
make[2]: *** [libdispatch_la-block.lo] Error 1
make[2]: Leaving directory '/home/ubuntu/swift-dev/swift-corelibs-libdispatch/src'
Makefile:541: recipe for target 'all' failed
make[1]: *** [all] Error 2
make[1]: Leaving directory '/home/ubuntu/swift-dev/swift-corelibs-libdispatch/src'
Makefile:457: recipe for target 'all-recursive' failed
make: *** [all-recursive] Error 1
```

In `src/shims/lock.h`:

```c
DISPATCH_ALWAYS_INLINE
static inline dispatch_once_t
{
#if defined(__i386__) || defined(__x86_64__)
        return os_atomic_xchg(pred, DLOCK_ONCE_DONE, release);
#elif defined(__linux__)
        if (unlikely(syscall(__NR_membarrier, MEMBARRIER_CMD_SHARED, 0) < 0)) {
                DISPATCH_INTERNAL_CRASH(errno, "sys_membarrier not supported");
        }
        return os_atomic_xchg(pred, DLOCK_ONCE_DONE, relaxed);
#else
#  error dispatch_once algorithm not available for this port
#endif
}
```

I guess an alternative is to move `#define _dispatch_hardware_crash() ...` up and call that directly  in `_dispatch_once_xchg_done()`, as the `#define` for `_dispatch_set_crash_log_cause_and_message()` appears to do nothing? Clearly it's unacceptable to move the includes down as I have currently if it breaks Darwin - unfortunately I don't have a Darwin system to test against.

MadCoder:
Just use __builtin_trap() here. And comment you can't use the define because of ordering.

amboar:
@MadCoder thanks, I'll rework the change

amboar:
I've reworked the commit and pushed - I wasn't sure whether to do that or apply a patch on top to fix (e.g. Rust's workflow). If I've taken the wrong approach please let me know.

ffried:
This also fixes the build on RasPi. 👍

PR #235, opened 2017-04-03
Fix compile error on s390x in shims/lock.h

vivkong:
s390x has a strong memory model and does not require membarrier.

This is the same as https://github.com/apple/swift-corelibs-libdispatch/pull/234 but for the 3.1 branch

### DIFF
--- a/src/shims/lock.h
+++ b/src/shims/lock.h
@@ -89,7 +89,7 @@ _dispatch_lock_has_failed_trylock(dispatch_lock lock_value)
 
 #elif defined(__linux__)
 #include <linux/futex.h>
-#if !defined(__x86_64__) && !defined(__i386__)
+#if !defined(__x86_64__) && !defined(__i386__) && !defined(__s390x__)
 #include <linux/membarrier.h>
 #endif
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


vivkong:
Yes it's comparable to Intel (see Table 5 in http://www.rdrop.com/~paulmck/scalability/paper/whymb.2009.04.05a.pdf).

src/shims/lock.h
```diff
@@ -542,8 +542,9 @@ DISPATCH_ALWAYS_INLINE
 static inline dispatch_once_t
 _dispatch_once_xchg_done(dispatch_once_t *pred)
 {
-#if defined(__i386__) || defined(__x86_64__)
+#if defined(__i386__) || defined(__x86_64__) || defined(__s390x__)
 	// On Intel, any load is a load-acquire, so we don't need to be fancy
+	// s390x has a strong memory model
```
MadCoder: I'd comment "same for s390x"

Do we want to throw in s390 too here? Or is that too obsolete as an architecture?

vivkong: ya I think we only have s390x in Swift so I'd like to stick with that.  Thanks for the review!

PR #236, opened 2017-04-08
libdispatch: Avoid submodules for headers that should not be included

bcardosolopes:
libdispatch has an umbrella header, dispatch.h, that should be used for
including all of the other headers.  This is enforced via
__DISPATCH_INDIRECT__.

Since it isn't legal to include one of the other headers on their own,
it isn't logical to have submodules for them.  Moreover, submodules
*should* have local visibility (be unaware of the context they're
included from); meaning that __DISPATCH_INDIRECT__ will never be
defined.

Logically, all of the headers are part of a single module/interface.

Patch by me & Duncan P. Exon Smith

### DIFF
--- a/dispatch/darwin/module.modulemap
+++ b/dispatch/darwin/module.modulemap
@@ -1,6 +1,5 @@
 module Dispatch [system] [extern_c] {
 	umbrella header "dispatch.h"
-	module * { export * }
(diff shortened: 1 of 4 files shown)


bcardosolopes:
@swift-ci please test

PR #239, opened 2017-04-12
use uintptr_t instead of long

compnerd:
The kind can sometimes hold a dispatch continuation, which is a pointer.
Ensure that the type of the parameter is able to hold a pointer
(consider LLP64 hosts which use 64-bit pointers, 32-bit longs and would
truncate).  This is an internal interface, and should be safe to change

### DIFF
--- a/src/source.c
+++ b/src/source.c
@@ -328,7 +328,7 @@ _dispatch_source_set_handler_slow(void *context)
 
 DISPATCH_NOINLINE
 static void
-_dispatch_source_set_handler(dispatch_source_t ds, long kind,
+_dispatch_source_set_handler(dispatch_source_t ds, uintptr_t kind,
 		dispatch_continuation_t dc)
 {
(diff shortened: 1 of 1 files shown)


MadCoder:
That doesn't make sense, this argument is used as an array index. I don't see any place where we masquerade a continuation as such a kind. (if so please enlighten me because that would be a bug)

This thing is supposed to be an index in this array:

```C
	struct dispatch_continuation_s *volatile ds_handler[3]
```

so long is sufficient enough

compnerd:
Ugh, I think I realize why I was confused.

348 `dc->dc_data = (void *)kind;`

On LLP64, that would be bad.  However, the value is only incidentally being cast to `void *`.  I guess that this should really just be cast to `uintptr_t`.

MadCoder:
dc_data is a void* we just need a place to stash it, which is a common pattern with dispatch. There's no truncation going on here, and when someone will do the windows port, which to my knowledge is the only popular LLP64 platform I know of, then we'll probably have way more issues than just that one, so I'm not keen on a one time fix here.

PR #240, opened 2017-04-13
Fix building on RasPi

ffried:
`shims.h` includes`shims/lock.h` which calls `DISPATCH_INTERNAL_CRASH`. However this is defined below the include of `shims.h` in `internal.h` and hence fails the build on the RasPi.
Moving the includes below the define of `DISPATCH_INTERNAL_CRASH` fixes the issue.

I don't know why this works on regular x86/x64 builds, though.


ffried:
Closing this in favor of #233 which is about the same issue but has a better solution.

PR #243, opened 2017-04-30
build: attempt to add swift support to cmake

compnerd:
Introduce support for building the swift SDK overlay into the CMake
based build system.  This was the one piece which was missing from the
original cmake port of the build system.


compnerd:
CC: @das @gottesmm @llvm-beanz

PR #244, opened 2017-05-03
Create a DISPATCH_PTR_SIZE macro and use it

compnerd:
This introduces a new macro, `DISPATCH_PTR_SIZE` which is defined much
like `LLVM_PTR_SIZE`.  We can gracefully fallback to alternate means of
checking the width of the pointer and use this rather than `__LP64__` to
determine if we are on a 64-bit host.

### DIFF
--- a/src/event/event_kevent.c
+++ b/src/event/event_kevent.c
@@ -1557,7 +1557,7 @@ _dispatch_mach_notify_port_init(void *context DISPATCH_UNUSED)
 	kern_return_t kr;
 #if HAVE_MACH_PORT_CONSTRUCT
 	mach_port_options_t opts = { .flags = MPO_CONTEXT_AS_GUARD | MPO_STRICT };
-#ifdef __LP64__
+#if DISPATCH_SIZEOF_PTR == 8
 	const mach_port_context_t guard = 0xfeed09071f1ca7edull;
 #else
 	const mach_port_context_t guard = 0xff1ca7edull;
--- a/tests/dispatch_queue_finalizer.c
+++ b/tests/dispatch_queue_finalizer.c
@@ -50,7 +50,7 @@ main(void)
 {
 	dispatch_test_start("Dispatch Queue Finalizer");
 
-#ifdef __LP64__
+#if DISPATCH_SIZEOF_PTR == 8
 	ctxt_magic = (void*)((uintptr_t)arc4random() << 32 | arc4random());
 #else
 	ctxt_magic = (void*)arc4random();
--- a/src/object_internal.h
+++ b/src/object_internal.h
@@ -225,7 +225,7 @@
 	.do_xref_cnt = DISPATCH_OBJECT_GLOBAL_REFCNT
 #endif
 
-#ifdef __LP64__
+#if DISPATCH_SIZEOF_PTR == 8
 // the bottom nibble must not be zero, the rest of the bits should be random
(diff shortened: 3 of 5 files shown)


compnerd:
CC: @MadCoder

compnerd:
CC: @das

dispatch/base.h
```diff
@@ -254,4 +254,16 @@
...
+#ifdef __SIZEOF_POINTER__
+#define DISPATCH_PTR_SIZE __SIZEOF_POINTER__
+#elif defined(_WIN64)
+#define DISPATCH_PTR_SIZE 8
+#elif defined(_WIN32)
+#define DISPATCH_PTR_SIZE 4
+#elif defined(_MSC_VER)
+#error "could not determine pointer size as a constant int for MSVC"
+#else
+#define DISPATCH_PTR_SIZE sizeof(void *)
```
MadCoder: you can't do that, it prevents doing comparisons with the preprocessor which is the whole point.

do
```
#elif defined(__LP64__) || defined(__LLP64__)
#define ... 8
#elif defined(__ILP32__)
#define ... 4
#else
#error ...
#endif // __SIZEOF_POINTER__
````

File: src/event/event_kevent.c [hunk 2]
```diff
@@ -1557,7 +1557,7 @@ _dispatch_mach_notify_port_init(void *context DISPATCH_UNUSED)
 	kern_return_t kr;
 #if HAVE_MACH_PORT_CONSTRUCT
 	mach_port_options_t opts = { .flags = MPO_CONTEXT_AS_GUARD | MPO_STRICT };
-#ifdef __LP64__
+#if DISPATCH_PTR_SIZE == 8
```
MadCoder: that doesn't work with `sizeof(void*)` typically

dispatch/base.h
```diff
@@ -254,4 +254,16 @@
 
 typedef void (*dispatch_function_t)(void *_Nullable);
 
+#ifdef __SIZEOF_POINTER__
+#define DISPATCH_PTR_SIZE __SIZEOF_POINTER__
```
MadCoder: `DISPATCH_SIZEOF_PTR` or SIZEOF_POINTER or a variant thereof is a better name.

compnerd:
Applied both of the suggestions.

dispatch/base.h
```diff
@@ -254,4 +254,20 @@
 
 typedef void (*dispatch_function_t)(void *_Nullable);
 
+#ifdef __SIZEOF_POINTER__
```
MadCoder: I didn't realize this was in base.h, sorry this is internal stuff and should be in "src/shims/hw_config.h"

for the test that uses it today, just use `__SIZEOF_POINTER__` it's less of an issue for tests.

compnerd:
Moved the definitions to `hw_config.h`.

tests/dispatch_queue_finalizer.c
```diff
@@ -50,7 +50,7 @@ main(void)
 {
 	dispatch_test_start("Dispatch Queue Finalizer");
 
-#ifdef __LP64__
+#if DISPATCH_SIZEOF_PTR == 8
```
MadCoder: that will not build will it?

MadCoder:
@swift-ci please test

MadCoder:
@swift-ci please test

compnerd:
At least on Linux, with CMake, all tests build and pass for me.

PR #245, opened 2017-05-07
Windows cleanups

compnerd:
Add some cl flags handling to the build system.  Improve macros for Windows builds to detect the target correctly.  This is not enough to perform a full Windows build yet but addresses some of the low hanging fruit.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,10 @@ check_include_files("unistd.h" HAVE_UNISTD_H)
 check_include_files("objc/objc-internal.h" HAVE_OBJC)
 
 check_library_exists(pthread sem_init "" USE_POSIX_SEM)
+if(CMAKE_SYSTEM_NAME STREQUAL Windows)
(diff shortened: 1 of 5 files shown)


MadCoder:
I thought that MSVC was supporting clang nowadays? I don't think we will ever be able to use MSVC as a compiler for the larger goal of supporting Swift because you will need blocks support and many other things.

am I missing something here?

compnerd:
Yeah, if you look at the diffs, the assumption is that you are using clang.  The support here is for the clang cl driver.

MadCoder:
oh I didn't realize that MSVC had a driver exposed through the same weird Windows command line parsing. I guess it makes sense, the same way clang is compatible with gcc CLI ;)

ok let me look at it again with that in mind then!

MadCoder:
@swift-ci please test

das:
it is invalid to put HAVE_* type checks in to API and SPI headers, there is no config.h file included by API clients on Darwin

PR #246, opened 2017-05-12
build: fix typo in build system

compnerd:
This fixes a typo in the build system which is important because it
fixes the dependency tracking and the output location for the swift doc
that is generated for the swift SDK overlay.


compnerd:
CC: @das @MadCoder @gottesmm @llvm-beanz

gottesmm:
moodule_directory => module_directory. I think this is fine to merge.

das:
thanks!

PR #248, opened 2017-05-19
clarify config of pthread_workqueue vs. internal_workqueue

dgrove-oss:
As suggested in review of PR-206, DISPATCH_USE_INTERNAL_WORKQUEUES
should not enable HAVE_PTHREAD_WORKQUEUES.  Instead, a new
DISPATCH_USE_WORKQUEUES is defined in queue.c to guard code that
should be enabled in either case.


dgrove-oss:
@MadCoder, I guess this needs to be tested by Swift CI and then it can be merged?

MadCoder:
Yes. I'm not suer why I forgot to fire it

@swift-ci please test

PR #250, opened 2017-05-23
Thread detach hook for Java JNI on Android

johnno1962:
This is a callback hook required when mixing multi threaded Swift and Java on Android. When Java JNI has been used on a thread the thread needs to be detached from JNI before the thread exits to free local variables or the application will crash e.g. https://github.com/SwiftJava/java_swift/blob/master/Sources/JavaJNI.swift#L20.
Introducing this hook gives any future Java integration code a chance to do this.

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -861,6 +861,18 @@ gettid(void)
 		if ((f) && tsd->k) ((void(*)(void*))(f))(tsd->k); \
 	} while (0)
 
+#ifdef __ANDROID__
+static void (*_dispatch_thread_detach_callback)(void);
+
+void
+_dispatch_install_thread_detach_callback(dispatch_function_t cb)
+{
+    if (os_atomic_xchg(&_dispatch_thread_detach_callback, cb, relaxed)) {
+        DISPATCH_CLIENT_CRASH(0, "Installing a thread detach callback twice");
+    }
+}
+#endif
+
 void
 _libdispatch_tsd_cleanup(void *ctx)
 {
@@ -885,6 +897,11 @@ _libdispatch_tsd_cleanup(void *ctx)
 	_tsd_call_cleanup(dispatch_voucher_key, _voucher_thread_cleanup);
 	_tsd_call_cleanup(dispatch_deferred_items_key,
 			_dispatch_deferred_items_cleanup);
+#ifdef __ANDROID__
+	if (_dispatch_thread_detach_callback) {
+		_dispatch_thread_detach_callback();
+	}
+#endif
 	tsd->tid = 0;
 }
 
--- a/src/swift/Queue.swift
+++ b/src/swift/Queue.swift
@@ -330,6 +330,15 @@ public extension DispatchQueue {
 		let p = Unmanaged.passRetained(v).toOpaque()
 		dispatch_queue_set_specific(self.__wrapped, k, p, _destructDispatchSpecificValue)
 	}
+
+	#if os(Android)
+	@_silgen_name("_dispatch_install_thread_detach_callback")
+	private static func _dispatch_install_thread_detach_callback(_ cb: @escaping @convention(c) () -> Void)
+
+	public static func setThreadDetachCallback(_ cb: @escaping @convention(c) () -> Void) {
+		_dispatch_install_thread_detach_callback(cb)
+	}
+	#endif
 }
 
 private func _destructDispatchSpecificValue(ptr: UnsafeMutableRawPointer?) {
--- a/private/queue_private.h
+++ b/private/queue_private.h
@@ -322,6 +322,20 @@ void
 dispatch_async_enforce_qos_class_f(dispatch_queue_t queue,
 	void *_Nullable context, dispatch_function_t work);
 
+#ifdef __ANDROID__
+/*!
+ * @function _dispatch_install_thread_detach_callback
+ *
+ * @param callback
+ * Function to be called before each worker thread exits to detach JVM.
+ *
+ * Hook to be able to detach threads from the Java JVM before they exit.
+ * If JNI has been used on a thread on Android it needs to have been
+ * "detached" before the thread exits or the application will crash.
+ */
+DISPATCH_EXPORT
+void _dispatch_install_thread_detach_callback(dispatch_function_t cb);
+#endif
 
 __END_DECLS
 


dispatch/queue.h
```diff
@@ -1346,6 +1346,16 @@ dispatch_assert_queue_not(dispatch_queue_t queue)
...
 
+/*!
+ * @handler dispatch_thread_detach_handler
+ *
+ * Hook to be able to detach threads before they exit from the Java JVM.
+ * If JNI has been used on a thread on Android it needs to have been
+ * "detached" before the thread exits or the application will crash.
+ */
+DISPATCH_EXPORT
+void (*dispatch_thread_detach_handler)();
```
MadCoder: 1) this should go to queue_internal.h
2) this should be `(void)`
3) this should only be enabled on android

src/queue.c
```diff
@@ -861,6 +861,8 @@ gettid(void)
 		if ((f) && tsd->k) ((void(*)(void*))(f))(tsd->k); \
 	} while (0)
 
+void (*dispatch_thread_detach_handler)();
```
MadCoder: Should be on android only

dispatch/queue.h
johnno1962: Changes made. queue_interal.h doesn’t seem to be visible from Swift so I have left it where it is and prefixed the symbol with an “_" for now.

dispatch/queue.h
```diff
@@ -1346,6 +1346,18 @@ dispatch_assert_queue_not(dispatch_queue_t queue)
...
+#ifdef __ANDROID__
+/*!
+ * @handler dispatch_thread_detach_callback
+ *
+ * Hook to be able to detach threads before they exit from the Java JVM.
+ * If JNI has been used on a thread on Android it needs to have been
+ * "detached" before the thread exits or the application will crash.
+ */
+DISPATCH_EXPORT
+void (*_dispatch_thread_detach_callback)(void);
```
MadCoder: for the second time this is **NOT** appropriate for an API header, this belongs to an internal one.
if the swift overlay is not built seeing the internal headers, then either try to fix that, or extern the prototype here again. it is an internal implementation detail between the overlay and libdispatch that should never be exposed anywhere.

this **HAS** to move to queue_internal.h, or maybe "private/private.h" this could be appropriate as well, and would likely be seen by the overlay.

src/queue.c
```diff
@@ -885,6 +889,10 @@ _libdispatch_tsd_cleanup(void *ctx)
 	_tsd_call_cleanup(dispatch_voucher_key, _voucher_thread_cleanup);
 	_tsd_call_cleanup(dispatch_deferred_items_key,
 			_dispatch_deferred_items_cleanup);
+#ifdef __ANDROID__
+	if (_dispatch_thread_detach_callback)
```
MadCoder: If it was like this before, sorry to have missed it, please add brackets.

dispatch/queue.h
johnno1962: There seems to be no way to expose the internal dispatch headers to the Swift side of things. There is a DispatchPrivate module but it won’t import during the build. So, I’ve used a @_silgen_name to pick up a setter function which isn’t ideal but at least it’s no affects the public api.

johnno1962: There seems to be no way to expose the internal dispatch headers to the Swift side of things. There is a DispatchPrivate module but it won’t import during the build. So, I’ve used a @_silgen_name to pick up a setter function which isn’t ideal but at least it’s no longer affects the public api.

src/queue.c
```diff
@@ -861,6 +861,14 @@ gettid(void)
 		if ((f) && tsd->k) ((void(*)(void*))(f))(tsd->k); \
 	} while (0)
 
+#ifdef __ANDROID__
+static void (*_dispatch_thread_detach_callback)(void);
```
MadCoder: if it's static then the overlay which is a dynamic library can't pick it up.

@das suggested to me the right place is to export the symbol is from `private/queue_private.h` or `private/private.h` and also to use silgen like you did.

johnno1962: Global symbol is now a function _dispatch_set_detach_callback as it doesn’t seem possible to express a global var in a silgen. I’ve moved it's declaration into queue_private.h.

src/queue.c
```diff
@@ -861,6 +861,15 @@ gettid(void)
...
+
+void
+*(*_dispatch_thread_detach_callback_ptr(void))(void) {
```
MadCoder: this is hard to read, also exposing a pointer to a function pointer is really weird. do you really need the getter? it feels like you only want to install it, I would have gone for:

```C
void
_dispatch_install_thread_detach_callback(dispatch_function_t cb)
{
    if (os_atomic_xchg(&_dispatch_thread_detach_callback, cb, relaxed)) {
        DISPATCH_CLIENT_CRASH(0, "Installing a thread detach callback twice");
    }
}
```

src/swift/Queue.swift
```diff
@@ -330,6 +330,20 @@ public extension DispatchQueue {
 		let p = Unmanaged.passRetained(v).toOpaque()
 		dispatch_queue_set_specific(self.__wrapped, k, p, _destructDispatchSpecificValue)
 	}
+
+	#if os(Android)
+	@_silgen_name("_dispatch_thread_detach_callback_ptr")
+	private static func _dispatch_thread_detach_callback_ptr() -> UnsafeMutablePointer<(@convention(c) () -> Void)?>
+
+	public static var threadDetachCallback: (@convention(c) () -> Void)? {
```
MadCoder: I think only a one-way setter should be exposed. having a getter makes the interface un-necessarily ugly.

johnno1962: Thanks for the potiners. It helps me zero in on something you’re happy with. Changed pushed.

johnno1962:
@MadCoder Any changes you’d like on the latest version?

johnno1962:
Hi @MadCoder, any chance we could arrive at something that is agreeable to you that we could merge?  I’d liek to tie the api down so I can release a Swift/Andoird precompiled toolchain. While it clutters the dispatch api a tad it is pretty much a requirement for apps mixing Java and Swift that use dispatch queues.

MadCoder:
Sorry I had forgotten about this pull request, the latest changes looks good, please squash it in a single commit so that I can merge.

johnno1962:
Can’t you do that as you merge for me? I can look at it but my git skills are very limited!

johnno1962:
Bah, tried a rebase and there are now 11 commits :( Changes are still correct. Can you squish as you merge please?

CodaFi:
Try

`git pull --rebase upstream master` followed by `git rebase -i` to interactively squash.

johnno1962:
Thanks but I’m afraid that hasn’t worked.. conflicts etc locally now. Can you not squish as you merge? Otherwise I’ll to start a new clone, remake the changes retest and resubmit.

johnno1962:
I’ve resubmitted this as PR https://github.com/apple/swift-corelibs-libdispatch/pull/259 after remaking the changes to a clean clone of up to date repo.

PR #251, opened 2017-05-24
Loosen test deadlines to reduce spurious failures in Swift CI

dgrove-oss:
Add build flag that defaults to true when building
with Swift support enabled (like when in Swift CI)
that relaxes the passing criteria for timing-sensitive
dispatch tests.  We continue to see occasional spurious
failures of dispatch tests in the Swift CI caused by
test machine load; this is an attempt to reduce the frequency
of load-induced failures without removing the test cases
entirely.


MadCoder:
@swift-ci please test

PR #253, opened 2017-05-29
Convert dispatch_workq from legacy priorities to qos

dgrove-oss:
Update dispatch_workq (DISPATCH_USE_INTERNAL_WORKQUEUE)
to use QoS-based constants instead of legacy priorities.

Enhance monitoring code to count runnable threads from
highest QoS to lowest and to suppress voluntary
oversubscription for lower QoS queues if the total count
of runnable worker threads is already over the desired
threshold.


src/event/workqueue.c
```diff
@@ -218,28 +218,27 @@ _dispatch_workq_monitor_pools(void *context DISPATCH_UNUSED)
...
 		}
 	}
 }
 #endif // HAVE_DISPATCH_WORKQ_MONITORING
 
-
-// temporary until we switch over to QoS based interface.
 static dispatch_queue_t
-get_root_queue_from_legacy_priority(int priority)
+get_root_queue_from_dispatch_qos(dispatch_qos_t qos)
```
das: this seems effectively the same as _dispatch_get_root_queue now ? (inline_internal.h)
can we not use that here ?

src/queue.c
```diff
@@ -154,7 +154,10 @@ struct dispatch_root_queue_context_s {
 			int volatile dgq_pending;
 #if DISPATCH_USE_WORKQUEUES
 			qos_class_t dgq_qos;
-			int dgq_wq_priority, dgq_wq_options;
+#if HAVE_PTHREAD_WORKQUEUES
```
das: I think this could be conditionalized more tightly with `HAVE_PTHREAD_WORKQUEUE_SETDISPATCH_NP || DISPATCH_USE_LEGACY_WORKQUEUE_FALLBACK` now (or add a define at the top for this combination to make it cleaner), those are the only remaining cases that need this field)

src/event/workqueue.c
```diff
@@ -69,7 +69,7 @@ typedef struct dispatch_workq_monitor_s {
 	int num_registered_tids;
 } dispatch_workq_monitor_s, *dispatch_workq_monitor_t;
 
-static dispatch_workq_monitor_s _dispatch_workq_monitors[WORKQ_NUM_PRIORITIES];
+static dispatch_workq_monitor_s _dispatch_workq_monitors[DISPATCH_QOS_MAX+1];
```
dgrove-oss: just to avoid index conversion, but I will change to match convention used elsewhere.

src/event/workqueue.c
dgrove-oss: thanks!  I thought there must be a routine that did this somewhere, but I missed finding it.

src/queue.c
dgrove-oss: will give it a try

dgrove-oss:
updated to address review comments.  

I'm not in a rush to get this merged, so fine with me if you need to put on the back burner until libdispatch-890 merge goes through.

das:
lgtm, thanks!

yes, lets please wait until libdispatch-890 can be merged (which will also create a small conflict here)

das:
@swift-ci please test and merge

das:
@swift-ci please test and merge

das:
@swift-ci please test

das:
@swift-ci please test

PR #254, opened 2017-06-01
Merge darwin/libdispatch-890 to master

das:
Merge tag darwin/libdispatch-890 on darwin/trunk

This includes the latest Apple source drop libdispatch-890 (up from libdispatch-806), on top of darwin/darwin-008 already merged in #252

PR #256, opened 2017-06-04
Re-merge darwin/libdispatch-890 to master

das:
Reverts apple/swift-corelibs-libdispatch#255 to remerge #254


das:
@swift-ci please test

PR #257, opened 2017-06-16
update CMake support

dgrove-oss:
remove libpwq support and add internal_workqueue.

also add dtrace rules to generate provider.h

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,45 +16,32 @@ set(THREADS_PREFER_PTHREAD_FLAG TRUE)
 find_package(Threads REQUIRED)
 
 include(GNUInstallDirs)
-include(ExternalProject)
 
 set(WITH_BLOCKS_RUNTIME "" CACHE PATH "Path to blocks runtime")
-set(WITH_PTHREAD_WORKQUEUES "" CACHE PATH "Path to pthread-workqueues")
 
 include(DispatchAppleOptions)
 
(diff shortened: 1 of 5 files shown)


dgrove-oss:
@MadCoder some updates to CMake support.  We'll need more than this before we can switch over, but this was a logical first step.  This is enough to at least get building libdispatch standalone and running the tests to work after the libpwq removal.

dgrove-oss:
Update with a number of additional fixes/tweaks so that CMake is now working nicely for me on Linux to build and test dispatch in combination with swift/utils/build-script.  @MadCoder I think this is ready to merge.   

The merge should have no actual effect (since it leaves the autoconf build in place).  Once this PR is merged, I will start submitting PRs to the other swift projects (swift, foundation, swiftpm) to incrementally switch them over to the new build system, once that is done we can finish the process by removing the autoconf-related files from libdispatch.

das:
oops, @dgrove-oss I think I just obsoleted/conflicted some of this by merging @compnerd's #260, should have checked for older open PRs first, sorry

das:
btw, before switching over other projects, it would be good to have the Darwin CMake build support in place so that we can test things together with other projects on Darwin also.
As mentioned in #260 I have mode some progress on that but it needs more work

dgrove-oss:
no worries; @compnerd is better with CMake than I am.   I'll take a look and see if there are parts of this are still needed and put them in a new PR

PR #258, opened 2017-06-16
Export objc_retainAutoreleasedReturnValue()

spevans:
- This allows it to be used by Foundation so that the duplicate
  function can be removed from CFRuntime.c in corelibs-foundation.

- This also makes the export match the symbol in libdispatch.a.

Since SwiftFoundation depends on libdispatch, there is no reason to have the function defined in both libraries. It shouldnt break the build as Foundation uses the linker argument `--allow-multiple-definition` but once this PR has been merged the duplicate function can be removed from Foundation and the linker argument eliminated.

At some point in the future the swift compiler may be fixed to not require this work around and then it can be removed from libdispatch as well.

### DIFF
--- a/src/swift/DispatchStubs.cc
+++ b/src/swift/DispatchStubs.cc
@@ -199,8 +199,19 @@ SOURCE(VNODE)
 #endif
 SOURCE(WRITE)
 
-// See comment in CFFuntime.c explaining why objc_retainAutoreleasedReturnValue is needed.
+#if !USE_OBJC
+
+// For CF functions with 'Get' semantics, the compiler currently assumes that
+// the result is autoreleased and must be retained. It does so on all platforms
+// by emitting a call to objc_retainAutoreleasedReturnValue. On Darwin, this is
+// implemented by the ObjC runtime. On non-ObjC platforms, there is no runtime,
+// and therefore we have to stub it out here ourselves. The compiler will
+// eventually call swift_release to balance the retain below. This is a
+// workaround until the compiler no longer emits this callout on non-ObjC
+// platforms.
 extern "C" void swift_retain(void *);
+
+SWIFT_CC(swift) DISPATCH_RUNTIME_STDLIB_INTERFACE
 extern "C" void * objc_retainAutoreleasedReturnValue(void *obj) {
     if (obj) {
         swift_retain(obj);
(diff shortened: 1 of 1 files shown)


parkera:
I'm not sure we want to propagate the lie of this function on Linux. It promises autorelease behavior but can't deliver it.

spevans:
@parkera Its really just a temporary fix, `libFoundation.so` is currently exporting it anyway, but this will allow it to be removed from libFoundation completely then once a better solution is implemented by the compiler (maybe just emiting the swift_retain() directly?) it can be removed from here. `libdispatch.a` currently exports the symbol so I dont think its anymore detrimental.

parkera:
dispatch seems like the wrong place for it though - it doesn't have any of the API which requires it.

dgrove-oss:
Seems like the logical temporary fix (if one is needed) would be to have a single symbol in the swift runtime that both dispatch and foundation could link against on linux.

parkera:
Yah I've long thought the runtime would be the place for both this stuff and the blocks as well. Putting it in dispatch is convenient but doesn't seem quite right.

spevans:
I can do a PR to put it in the runtime instead, it will just need to be merged at the same time it is removed completely from libdispatch to avoid duplicate symbol issues

spevans:
I updated the PR to remove the function and also created https://github.com/apple/swift/pull/10355 which adds the function to swift runtime.

This PR will need to be tested/ merged at the same time as the swift PR

spevans:
Is there any objection to this PR? I know it is a temporary work around but removing the duplicated symbol will be helpful in simplifying the linker commands used for SR-648. If this change is OK I will open a followup to remove the duplicate from `libFoundation`.

I opened https://bugs.swift.org/browse/SR-5271 for a correct fix for this problem which probably requires a compiler change but Im not sure how much work is involved there.

parkera:
Is this function actually used anywhere in dispatch? If not, maybe a better answer is to simply remove it and let Foundation use its own.

spevans:
@das I wrapped the function inside a `#if !USE_OBJC` and updated the comment changing Linux to non-ObjC platform.

@parkera It is used by swift_overlay.o

das:
looks good, thanks!

PR #259, opened 2017-06-21
Thread detach hook for Java JNI on Android

johnno1962:
This is a resubmission of a callback hook required when mixing multi threaded Swift and Java on Android. When Java JNI has been used on a thread the thread needs to be detached from JNI before the thread exits to free local variables or the application will crash e.g. https://github.com/SwiftJava/java_swift/blob/master/Sources/JavaJNI.swift#L25.
Introducing this hook gives any future Java integration code a chance to do this.

### DIFF
--- a/private/queue_private.h
+++ b/private/queue_private.h
@@ -323,6 +323,20 @@ void
 dispatch_async_enforce_qos_class_f(dispatch_queue_t queue,
 	void *_Nullable context, dispatch_function_t work);
 
+#ifdef __ANDROID__
+/*!
+ * @function _dispatch_install_thread_detach_callback
+ *
+ * @param callback
+ * Function to be called before each worker thread exits to detach JVM.
+ *
+ * Hook to be able to detach threads from the Java JVM before they exit.
+ * If JNI has been used on a thread on Android it needs to have been
+ * "detached" before the thread exits or the application will crash.
+ */
+DISPATCH_EXPORT
+void _dispatch_install_thread_detach_callback(dispatch_function_t cb);
+#endif
 
 __END_DECLS
 
--- a/src/queue.c
+++ b/src/queue.c
@@ -885,6 +885,18 @@ gettid(void)
 		if ((f) && tsd->k) ((void(*)(void*))(f))(tsd->k); \
 	} while (0)
 
+#ifdef __ANDROID__
+static void (*_dispatch_thread_detach_callback)(void);
+
+void
+_dispatch_install_thread_detach_callback(dispatch_function_t cb)
+{
+    if (os_atomic_xchg(&_dispatch_thread_detach_callback, cb, relaxed)) {
+        DISPATCH_CLIENT_CRASH(0, "Installing a thread detach callback twice");
+    }
+}
+#endif
+
 void
 _libdispatch_tsd_cleanup(void *ctx)
 {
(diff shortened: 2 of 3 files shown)


MadCoder:
@swift-ci please test

johnno1962:
Thanks @MadCoder, good to have this tied down. While you’re here there are in fact three problems building libdispatch on Android which you can probably provide an authoratative answer on.

The first is membarrier is not available:
```
diff --git a/src/shims/lock.h b/src/shims/lock.h
index 99c5563..34e1559 100644
--- a/src/shims/lock.h
+++ b/src/shims/lock.h
@@ -57,11 +57,11 @@ _dispatch_lock_owner(dispatch_lock lock_value)
 }
 
 #elif defined(__linux__)
 
 #include <linux/futex.h>
-#if !defined(__x86_64__) && !defined(__i386__) && !defined(__s390x__)
+#if !defined(__x86_64__) && !defined(__i386__) && !defined(__s390x__) && !defined(__ANDROID__)
 #include <linux/membarrier.h>
 #endif
 #include <unistd.h>
 #include <sys/syscall.h>   /* For SYS_xxx definitions */
 
@@ -471,11 +471,11 @@ _dispatch_once_gate_tryenter(dispatch_once_gate_t l)
 
 DISPATCH_ALWAYS_INLINE
 static inline dispatch_once_t
 {
-#if defined(__i386__) || defined(__x86_64__) || defined(__s390x__)
+#if defined(__i386__) || defined(__x86_64__) || defined(__s390x__) || defined(__ANDROID__)
        // same for s390x
        return os_atomic_xchg(pred, DLOCK_ONCE_DONE, release);
 #elif defined(__linux__)
        if (unlikely(syscall(__NR_membarrier, MEMBARRIER_CMD_SHARED, 0) < 0)) {
```
And the seems to be some problem with the MACRO “TRASHIT”
```
diff --git a/src/internal.h b/src/internal.h
index 0536db1..ecbfc8f 100644
--- a/src/internal.h
+++ b/src/internal.h
@@ -377,13 +377,15 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
 // sys/queue.h debugging
 #if defined(__linux__)
 #define QUEUE_MACRO_DEBUG 1
 #else
 #undef TRASHIT
-#define TRASHIT(x) do {(x) = (void *)-1;} while (0)
 #endif
 #endif // DISPATCH_DEBUG
+#ifndef TRASHIT
+#define TRASHIT(x) do {(x) = (void *)-1;} while (0)
+#endif
 #define _TAILQ_TRASH_ENTRY(elm, field) do { \
                        TRASHIT((elm)->field.tqe_next); \
                        TRASHIT((elm)->field.tqe_prev); \
                } while (0)
 #define _TAILQ_TRASH_HEAD(head) do { \
```
And finally libdispatch needs to be linked to libswiftCore.so while it is being built.

Do you want me to file a new PR for the solutions to the first two problems?

MadCoder:
This review is not the place to discuss it. however, the dispatch once thing is wrong on android, arm gives no such promise.

johnno1962:
Any suggestions as to what the proper solution might be?

MadCoder:
see #228 . If you aren't fluent with arm memory model, you should leave that patch to someone who does.

johnno1962:
Fair enough. WIll do. Thanks for your help

PR #260, opened 2017-06-28
cmake parity updates

compnerd:
This updates the CMake build to bring it to parity with the autotools build.  It also cleans up some of the custom list management for more modern approaches.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -53,6 +53,12 @@ add_library(dispatch
               shims/time.h
               shims/tsd.h
               shims/yield.h)
+if(DISPATCH_USE_INTERNAL_WORKQUEUE)
+  target_sources(dispatch
+                 PRIVATE
+                   event/workqueue.c
+                   event/workqueue_internal.h)
+endif()
 target_sources(dispatch
                PRIVATE
                  block.cpp)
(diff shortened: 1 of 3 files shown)


compnerd:
CC @das @MadCoder @llvm-beanz @gottesmm

das:
very cool, thanks!

BTW, I've made some progress on getting the CMake build to work on Darwin (against submodules of the dependent projects under github.com/apple/darwin*), I'll have to rebase on top of this and see if that solves some of the problems I was running into

compnerd:
@das awesome!  Really looking forward to seeing this become the one build system for libdispatch.  I am able to build with this change on linux and all tests pass.  If you have specific things that I can try to address in the build system, to help in getting everything switched over, let me know!

src/CMakeLists.txt
```diff
@@ -90,9 +52,57 @@ add_library(dispatch
...
+                      swift/Dispatch.swift
+                      swift/IO.swift
+                      swift/Private.swift
+                      swift/Queue.swift
+                      swift/Source.swift
+                      swift/Time.swift
+                      swift/Wrapper.swift
+                    CFLAGS
+                      -fblocks
+                      -fmodule-map-file=${CMAKE_SOURCE_DIR}/dispatch/module.modulemap
```
das: this is a pre-existing bug but this is the linux modulemap, if we ever want this part to work on Darwin this path will need to be conditionalized per platform

PR #261, opened 2017-06-28
build: switch the module map according to the target

compnerd:
Use a darwin specific modulemap on darwin, and the generic one
elsewhere.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,6 +188,23 @@ if(leaks_EXECUTABLE)
   set(HAVE_LEAKS TRUE)
 endif()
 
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @das 

Should help with the Darwin support.

das:
makes sense, don't we have this distinction on the install path somewhere already ? (the right modulemap has to be installed as well on a per-platform basis)
they could probably shared the same variable

compnerd:
Seems that the build use to rely on symbolic links.  Add support for doing that in the CMake build rather than the custom handling.

das:
ok, makes sense to replicate that if that is how the existing autoconf build system was doing it, and I guess that also makes the install of the modulemap work (by copying the target of the symlink)

PR #262, opened 2017-06-28
build: add option to use the gold linker

compnerd:
This used to be an option in the autotools build.  Address the TODO item
and add an option to control that.


compnerd:
CC: @das

PR #263, opened 2017-06-28
Fixes for C-level build of libdispatch on Linux

dgrove-oss:
Still needed pieces of PR#257 related to C-level build/test.
  1. rule to generate provider.h from provider.d using dtrace
  2. default to building shared libraries (needed for Swift)
  3. increase test timeout to 600 seconds (for noisy Swift CI)


dgrove-oss:
@das @compnerd   The C-level parts of #257 that I think we still need.

Will do the Swift-level fixes in second PR.

CMakeLists.txt
```diff
@@ -29,6 +29,8 @@ include(DispatchAppleOptions)
 option(ENABLE_DISPATCH_INIT_CONSTRUCTOR "enable libdispatch_init as a constructor" ON)
 set(USE_LIBDISPATCH_INIT_CONSTRUCTOR ${ENABLE_DISPATCH_INIT_CONSTRUCTOR})
 
+option(BUILD_SHARED_LIBS "build shared libraries" ON)
```
compnerd: Do we want to default to the shared libraries first?  We can, I went with the defaults of auto tools, which was static by default.

src/CMakeLists.txt
```diff
@@ -1,7 +1,18 @@
 
 include(SwiftSupport)
 
+set(dispatch_DTRACE_HEADERS)
+if(dtrace_EXECUTABLE)
+  list(APPEND dispatch_DTRACE_HEADERS ${CMAKE_CURRENT_BINARY_DIR}/provider.h)
+  ADD_CUSTOM_COMMAND(
```
compnerd: Lower case please.  Can you please create a target for this?  This will otherwise run constantly.  I can quickly write a patch to prevent that if you like.

src/CMakeLists.txt
```diff
@@ -1,7 +1,18 @@
...
+    OUTPUT  ${CMAKE_CURRENT_BINARY_DIR}/provider.h
+    COMMAND ${dtrace_EXECUTABLE} -h -s ${CMAKE_CURRENT_SOURCE_DIR}/provider.d -o ${header} ${CMAKE_CURRENT_BINARY_DIR}/provider.h
+    DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/provider.d
+  )
+endif()
+
 add_library(dispatch
+              ${dispatch_DTRACE_HEADERS}
```
compnerd: Use `target_sources` instead please.

tests/CMakeLists.txt
```diff
@@ -90,7 +90,7 @@ function(add_unit_test name)
            COMMAND bsdtestharness $<TARGET_FILE:${name}>)
   set_tests_properties(${name}
                        PROPERTIES
-                         TIMEOUT 30
+                         TIMEOUT 600
```
compnerd: Ill leave this to @das to figure out.  The old timeout used to be 30s.

compnerd:
@dgrove-oss what swift level changes are needed?

tests/CMakeLists.txt
dgrove-oss: right.  30 is not long enough.  Swift CI machines get very loaded.   I'm seeing dispatch_select routinely take 40 or 50 seconds.

CMakeLists.txt
dgrove-oss: we need the shared libraries when building to include in Swift.  For me at least, autotools/libtool was defaulting to shared libs before.

src/CMakeLists.txt
dgrove-oss: not sure what you mean;  the dependency seems to be working for me.

dgrove-oss:
For Swift, we need to add dependencies on a couple of Swift runtime libraries when linking the test exectutables.

dgrove-oss:
Mostly redundant with #264. Closing.  Will try again with the rest of my fixes once you merge the pending PRs from @compnerd

das:
sounds good, thanks. I merged the pending ones

PR #264, opened 2017-06-28
build: add dtrace USDT probe support

compnerd:
Add the missing support for dtrace header script support.

### DIFF
--- a/cmake/modules/DTrace.cmake
+++ b/cmake/modules/DTrace.cmake
@@ -1,0 +1,26 @@
+
+function(dtrace_usdt_probe script)
+  set(options)
+  set(single_parameter_options TARGET_NAME OUTPUT_SOURCES)
+  set(multiple_parameter_options)
+
+  cmake_parse_arguments("" "${options}" "${single_parameter_options}" "${multiple_parameter_options}" ${ARGN})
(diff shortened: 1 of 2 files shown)


compnerd:
@dgrove-oss does this work for you?  (I dont have dtrace installed, so if you could test this, it would be wonderful)

dgrove-oss:
Almost, you need to add the directory in which provider.h is generated to the includes.  With the additional patch below it works for me.

diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index ca6ef75..bdc9162 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -115,6 +115,7 @@ endif()
 target_include_directories(dispatch
                            PRIVATE
                              ${CMAKE_BINARY_DIR}
+                             ${CMAKE_CURRENT_BINARY_DIR}
                              ${CMAKE_SOURCE_DIR}
                              ${CMAKE_CURRENT_SOURCE_DIR}
                              ${CMAKE_SOURCE_DIR}/private)

compnerd:
@dgrove-oss updated

das:
awesome, thanks!

PR #265, opened 2017-06-28
tests: fix test build

compnerd:
The DISPATCH_SIZEOF_PTR caused a test to go down the wrong path.  This
fixes the test and accounts for the Windows 64 bit case as well (which
is not a LP64 environment).


compnerd:
CC: @das

das:
FWIW we have a lot of reliance on __LP64__ throughout to detect a 64bit platform, if we want to make this portable to _WIN64 I suspect we'll need many more changes than this

PR #266, opened 2017-06-28
build: remove dead code

compnerd:
Remove the supporting code for WITH_PTHREAD_WORKQUEUES which was removed
previously.  NFC.


compnerd:
CC: @das

PR #267, opened 2017-06-29
CMake compatibility with Swift build assumptions

dgrove-oss:
All of the swift projects that depend on finding libdispatch.so
during their builds (swift, foundation, swiftpm) all expect it
to be in src/.libs/libdispatch.so (libtool convention). Temporarily
add a rule to copy the built library to where libtool would have
placed it to decouple using CMake for libdispatch from updating
all of the other dependent projects' build expectations.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -188,6 +188,15 @@ if(USE_GOLD_LINKER)
                  -fuse-ld=gold)
 endif()
 
+# Temporary staging; the various swift projects that depend on libdispatch
+# all expect libdispatch.so to be in src/.libs/libdispatch.so
+# So for now, make a copy so we don't have to do a coordinated commit across
+# all the swift projects to change this assumption.
+add_custom_command(TARGET dispatch POST_BUILD
+                   COMMAND cmake -E make_directory .libs
+                   COMMAND cmake -E copy $<TARGET_FILE:dispatch> .libs
+                   COMMENT "Copying libdispatch to .libs")
+
 install(TARGETS
           dispatch
         DESTINATION


dgrove-oss:
@das @compnerd  -- staging for invoking libdispatch CMake from swift/utils/build-script.     The changes to the other projects to avoid needing this are simple, but I don't want to try to coordinate across all of the repos.  So we put in this crutch with the plan of updating the other projects soon after we cut-over to CMake for libdispatch and then getting rid of this.

compnerd:
@dgrove-oss yeah, I have a pr for that in the swift repo.  Don't see a problem with trying to stage it with some hacks in the build system.

src/CMakeLists.txt
```diff
@@ -188,6 +188,15 @@ if(USE_GOLD_LINKER)
...
 endif()
 
+# Temporary staging; the various swift projects that depend on libdispatch
+# all expect libdispatch.so to be in src/.libs/libdispatch.so
+# So for now, make a copy so we don't have to do a coordinated commit across
+# all the swift projects to change this assumption.
+add_custom_command(TARGET dispatch POST_BUILD
+                   COMMAND cmake -E make_directory .libs
+                   COMMAND cmake -E copy $<TARGET_FILE:dispatch> .libs
+                   COMMENT "Copying libdispatch to .libs")
```
compnerd: Doesn't this always run?  Perhaps have a output to prevent that.

dgrove-oss: for me it runs exactly when dispatch is built, which I think is what we want.

dgrove-oss: for the record, can't specify both a TARGET and an OUTPUT in a custom_command.

CMake Error at src/CMakeLists.txt:195 (add_custom_command):
  add_custom_command Wrong syntax.  A TARGET and OUTPUT can not both be
  specified.

das:
lgtm, thanks!

src/CMakeLists.txt
compnerd: Ah, I guess I was mixing up some other behaviour then.  If it is only run once, then it is what we want.

PR #269, opened 2017-06-30
Set CMake defaults to match autoconf Linux behavior.

dgrove-oss:
There may be a more elegant way to do this, but we really want
to build shared libraries by default on Linux and we want to
build the test cases and be able to run them.


dgrove-oss:
cc @das @compnerd

das:
makes sense

PR #270, opened 2017-06-30
link swift libraries when building test executables

dgrove-oss:
Add hook to enable passing in the location of the swift
runtime libraries so we can include them when building
test executables.  If we don't do this and we built
dispatch as a shared library and enabled swift support
the link step fails.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -39,6 +39,13 @@ if(BSD_OVERLAY_FOUND)
                           ${BSD_OVERLAY_LDFLAGS})
 endif()
 
+if(CMAKE_SWIFT_RUNTIME_LIBDIR)
+  target_link_libraries(bsdtestharness
+                        PRIVATE
+                        -L${CMAKE_SWIFT_RUNTIME_LIBDIR} -lswiftCore -lswiftSwiftOnoneSupport
+                        -Wl,-rpath -Wl,${CMAKE_SWIFT_RUNTIME_LIBDIR})
+endif()
+
 function(add_unit_test name)
   set(options DISABLED_TEST)
   set(single_value_args)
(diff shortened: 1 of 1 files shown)


dgrove-oss:
@das @compnerd   This matches what the autotools build currently does.  If there is a better way to accomplish this in CMake, I'm happy to change the patch.  But as far as I can tell, we do need to get these link options when (a) building shared libraries and (b) building dispatch including the swift overlay.

@compnerd, in my version of the patch for swift/utils/build-script-impl, I also define CMAKE_SWIFT_RUNTIME_LIBDIR .   I think you may want to pick up some fragments of https://github.com/dgrove-oss/swift/commit/9b5a1491f56ddd711b397afaf7161113408f103c to include in https://github.com/apple/swift/pull/10670

das:
hmm, it would be good to have/preserve the ability link against the libdispatch shared library without linking in the swift runtime (which the C unit test binaries shouldn't otherwise need right?)

but it sounds like that is more of a pre-exisisting issue that the autoconf buildsystem already had on Linux ? i.e. we can deal with this separately later on

FWIW we will definitely not be able to have this dependency on Darwin if we want to be able to build a replacement of the libdispatch.dylib that ships with the OS (like we used to be able to with the autoconf buildsystem and is something I would strongly like make work again with CMake)

dgrove-oss:
It should work ok on Darwin.  The linking/inclusion of the Swift overlay and Swift libraries are conditional on two vartiables (CMAKE_SWIFT_RUNTIME_LIBDIR and CMAKE_SWIFT_COMPILER) being defined when CMake is invoked (done by swift/utils/build-script).  if they aren't defined, then it just builds a straight C library with no Swift dependencies.

tests/CMakeLists.txt
```diff
@@ -39,6 +39,13 @@ if(BSD_OVERLAY_FOUND)
                           ${BSD_OVERLAY_LDFLAGS})
 endif()
 
+if(CMAKE_SWIFT_RUNTIME_LIBDIR)
+  target_link_libraries(bsdtestharness
+                        PRIVATE
+                        -L${CMAKE_SWIFT_RUNTIME_LIBDIR} -lswiftCore -lswiftSwiftOnoneSupport
+                        -Wl,-rpath -Wl,${CMAKE_SWIFT_RUNTIME_LIBDIR})
```
compnerd: Adding rpaths is discouraged.  It can be used as a security issue to interposition other libraries.  I think think that we should be trying to do this differently.  Furthermore, the `-l` should not be added when doing the `target_link_libraries` as it will normally put the right prefix.  Finally, `-L` should be handled via `link_directories`.  Ill put up a PR to address some of this.

PR #271, opened 2017-07-02
build: avoid rpath, library search path handling

compnerd:
Create an imported library target for the runtime libraries rather than
add the explicit library search path handling.  This avoids the need for
the rpath in the tests, and allows cmake to properly model dependencies.
It also makes the build system more amenable to
cross-compiling/cross-linking by avoiding ELF linker specific options.


compnerd:
CC: @das @dgrove-oss

das:
very nice, thanks!

PR #274, opened 2017-07-06
test suite fixes to prep for enabling additional warnings

dgrove-oss:
fixes to allow clean compilation of test suite with addtional
clang warnings enabled.


das:
@dgrove-oss this broke the build and was reverted with #278, my bad for not running it through the swift-ci before merging

you'll want to make a new PR for this branch

PR #279, opened 2017-07-14
Fix improper double-fire of signal sources on Linux

MadCoder:
(the issue was reported to me privately)


MadCoder:
@swift-ci please test

PR #280, opened 2017-07-14
Actually update epoll events if needed when unregistering

MadCoder:
_dispatch_epoll_update() was being called without dmn->dmn_events being
updated beforehand.

Signed-off-by: Pierre Habouzit <phabouzit@apple.com>


MadCoder:
@swift-ci please test

PR #282, opened 2017-07-18
test suite fixes to prep for enabling additional warnings

dgrove-oss:
fixes to allow clean compilation of test suite with additional
clang warnings enabled.

Updated version of #274 that leaves out the problematic change to dispatch_c99.c that broke the build.  There is a divergence between autotools and cmake compilation flags for dispatch_c99.c (autotools compiles this test without ${BSD_OVERLAY_CFLAGS} while CMake compiles all tests with ${BSD_OVERLAY_CFLAGS}).  We can fix that later.

### DIFF
--- a/tests/bsdtestharness.c
+++ b/tests/bsdtestharness.c
@@ -93,7 +93,7 @@ main(int argc, char *argv[])
 
 #endif
 	int i;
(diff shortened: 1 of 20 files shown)


das:
@swift-ci please test

PR #283, opened 2017-07-25
Enable additional compiler warnings for CMake build

dgrove-oss:
1. add a CMake module to define additional warning flags to enable
   when compiling dispatch.

2. enable the additional warning flags by default for CMake

3. match autotools behavior of not including the BSD_OVERLAY
   when compiling the dispatch_c99 test case. This avoids
   a warning about __printflike being redefined when compiling
   this test case with the expanded set of warning flags.


dgrove-oss:
@compnerd please comment if there are better ways to do this in CMake.  Thanks

cmake/modules/DispatchCompilerWarnings.cmake
```diff
@@ -0,0 +1,85 @@
+
+macro(dispatch_common_warnings)
```
compnerd: This definitely is problematic.  You break cross-compiling for Windows this way.  You should check if the compiler flag is supported and if you are using a GNU frontend before adding these options.  This is also a very large set of flags that you are adding, maintaining this seems like a headache.  How were these handled in autotools?

das: these are the flags we have enabled in the darwin build internally (in Xcode), so shared code should be built with these enabled to avoid breaking the darwin build (and I want the darwin CMake build to definitely have these so that it matches the Xcode build as closely as possible).
We maintain these flags internally by hand by adding or removing new ones as clang adds support for them (based on -Weverything).

If we need to opt out non-clang platforms like Windows that we don't build & support on a continual basis right now, that is a much lesser concern IMO

tests/CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ function(add_unit_test name)
   if(WITH_BLOCKS_RUNTIME)
     target_link_libraries(${name} PRIVATE BlocksRuntime)
   endif()
-  if(BSD_OVERLAY_FOUND)
+  if(BSD_OVERLAY_FOUND AND (NOT ${name} MATCHES "dispatch_c99"))
```
compnerd: These exclusions are ugly.  How about adding a `NO_BSD_OVERLAY` keyword option to the `add_unit_test` so that it is generic.

dgrove-oss:
updated based on review comments.

cmake/modules/DispatchCompilerWarnings.cmake
```diff
@@ -0,0 +1,91 @@
+
+# TODO: really should be checking for clang/gcc compatibile compiler 
+if(CMAKE_SYSTEM_NAME STREQUAL Windows)
```
compnerd: if("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")

tests/CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ if(ENABLE_SWIFT)
                           swiftSwiftOnoneSupport)
 endif()
 
-function(add_unit_test name)
+function(add_unit_test name uses_bsd_overlay)
```
compnerd: Just make this a keyword argument (add `NO_BSD_OVERLAY` to the `options` list below), you can reference it as `AUT_NO_BSD_OVERLAY` then.

dgrove-oss:
updated again based on @compnerd feedback

tests/CMakeLists.txt
```diff
@@ -60,6 +60,7 @@ endif()
 
 function(add_unit_test name)
   set(options DISABLED_TEST)
+  set(options NO_BSD_OVERLAY)
```
compnerd: This drops the `DISABLED_TEST` since you overwrite the variable.  Just append to it.

    set(options DISABLED_TEST;NO_BSD_OVERLAY)

tests/CMakeLists.txt
```diff
@@ -138,6 +138,8 @@ set(DISPATCH_C_TESTS
     data
     io_net
     select)
+set(DISPATCH_C99_TESTS
```
compnerd: I wouldn't bother with a list for a single test.  Just inline it below.

dgrove-oss:
updated.

There are probably going to be merge conflicts between this and #284.  I'd suggest merging this one first and I will update #284 to resolve the conflicts after this is merged to master.

das:
@dgrove-oss sounds good, I've merged #284

das:
argh, and now I re-read what you said and I did it the wrong way around... sorry!

dgrove-oss:
no worries; I can make it work either way.  working on it now

dgrove-oss:
ok.  conflicts resolved.  should be ready to go.

das:
thanks!

PR #284, opened 2017-07-26
CMake: carry over some nuances from autotools test config

dgrove-oss:
Port over test-specific compilation flags and support for
lenient deadlines for running in the Swift CI from Makefile.am

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -74,6 +74,11 @@ function(add_unit_test name)
                                ${CMAKE_CURRENT_BINARY_DIR}
                                ${CMAKE_CURRENT_SOURCE_DIR}
                                ${CMAKE_SOURCE_DIR})
+  if(CMAKE_SWIFT_COMPILER)
+	# For testing in swift.org CI system; make deadlines lenient by default
+	# to reduce probability of test failures due to machine load.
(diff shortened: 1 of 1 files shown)


tests/CMakeLists.txt
```diff
@@ -154,3 +157,9 @@ endforeach()
 target_link_libraries(dispatch_group PRIVATE m)
 target_link_libraries(dispatch_timer_short PRIVATE m)
 
+# test-specific compiler flags
+target_compile_options(dispatch_c99 PRIVATE -std=c99)
+
+# test-specific compiler flags (for currently disabled tests)
+#add_unit_test(dispatch_priority2 SOURCES dispatch_priority.c)
+#target_compile_options(dispatch_priority2, PRIVATE -DUSE_SET_TARGET_QUEUE=1)
```
compnerd: Drop the `,`.  Why is this test disabled?  We do have the `DISABLED_TEST` keyword, so we can add the test but not have it be built/run.

    add_unit_test(dispatch_priority2
                  DISABLED_TEST
                  SOURCES
                    dispatch_priority.c)

I think that we should add a dummy target so we can actually add the additional flags.

dgrove-oss:
updated and expanded (noticed some additional useful bits that hadn't been ported over).

das:
lgtm, thanks!

PR #285, opened 2017-07-28
tests: fix undefined behaviour

compnerd:
Ensure that we do not attempt a divide-by-zero.  Adjust the loop count
accordingly.


compnerd:
CC: @das

das:
heh, fun one, thanks!

PR #287, opened 2017-07-29
build: add support to build with sanitization

compnerd:
Add support to configure the build to use sanitizers to aid in finding
issues.  The `DISPATCH_USE_SANIZITERS` is a cmake option that defaults
to none, but can be one of `Address`, `Memory`, `MemoryWithOrigins`,
`Undefined`, `Leaks`, `Address;Undefined`.  This allows for easily building
libdispatch and the tests with the sanitizers.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ include(GNUInstallDirs)
 set(WITH_BLOCKS_RUNTIME "" CACHE PATH "Path to blocks runtime")
 
 include(DispatchAppleOptions)
+include(DispatchSanitization)
 
 include(DispatchCompilerWarnings)
 dispatch_common_warnings()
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @das @dgrove-oss

compnerd:
Since I put this up on a weekend, a gentle ping during the work week!

das:
out of interest, do the sanitizers actually work with libdispatch on Linux ? 

on Darwin, they depend on layers of the OS layered above libdispatch (or entangled with it) so we cannot use them for libdispatch at present

probably means you want to conditionalize  this for linux

compnerd:
Yeah, I was able to build with them on Linux.  I can certainly add a limitation to only allow them on Linux.

das:
but do they run? (the problem on Darwin is not at build time FWIW)

compnerd:
Yes, it did run as well for me :-).

das:
awesome!

PR #288, opened 2017-07-31
test: avoid -Wunused-variable on NoAsserts build

compnerd:
Ensure that the variable is referenced to avoid an unused variable
warning from GCC/clang when building the tests in a Release+NoAsserts
configuration.  This is done as part of the swift build on Linux for
SourceKit.


compnerd:
CC: @das @dgrove-oss

MadCoder:
why is it needed? dispatch makes sure to reference anything that is used in assertions even when assertions are off

MadCoder:
wait I should have read the patch first...

PR #286, opened 2017-07-29
linux: update header used for `major` macro

compnerd:
Newer versions of glibc indicate that they intend to move the `major`
macro from `sys/types.h` to `sys/sysmacros.h`.  Add a check for the
header and include that earlier to ensure that the macro is provided by
the newer header when available/possible.  This avoids an unnecessary
warning from the system headers.

PR #292, opened 2017-08-02
Revert "linux: update header used for `major` macro"

atrick:
Reverts apple/swift-corelibs-libdispatch#286

Sorry to revert, but this breaks Swift CI incremental bots and PR testing (so it's blocking all other work)

swift-corelibs-libdispatch/os/linux_base.h:16:10: fatal error: 'config/config_ac.h' file not found
#include <config/config_ac.h>
^

https://ci.swift.org/job/oss-swift-incremental-RA-linux-ubuntu-16_04/4182/console


compnerd:
Ugh, Ive been building with the CMake build, not autotools, which is why I didnt notice that.  The original version of the patch that @das recommended to remove the conditional include would've worked -- since autoconf would generate config.h.

das:
No I don’t think it would have, see my comments in the original PR

PR #294, opened 2017-08-02
linux: update header used for `major` macro

compnerd:
Newer versions of glibc indicate that they intend to move the major
macro from sys/types.h to sys/sysmacros.h. Add a check for the header
and include that earlier to ensure that the macro is provided by the
newer header when available/possible. This avoids an unnecessary warning
from the system headers.

Because `config_ac.h` is not available at the right location, we cannot
include the header to check whether the header is available.  Rely on
the compiler provided `__has_include` instead of the configure check.
Adjust the inclusion of `sys/cdefs.h` accordingly.


private/private.h
```diff
@@ -39,7 +39,7 @@
 #if HAVE_UNISTD_H
 #include <unistd.h>
 #endif
-#if HAVE_SYS_CDEFS_H
+#if __has_include(<sys/cdefs.h>)
```
das: one concern with this approach here we've always had is that __has_include is not supported on all compilers we support at least in the Darwin SDK (that is why dispatch/base.h defines them). 

I would suggest replicating the approach at the top of dispatch/dispatch.h here (replacing lines 30-32)

```
#ifdef __APPLE__
#include <Availability.h>
#include <os/availability.h>
#include <os/base.h>
#elif defined(__linux__)
#include <os/linux_base.h>
#endif
```

and then actually dropping the `#include <sys/cdefs.h>` completely here (as it will already be covered by the base*.h includes)

compnerd:
Updated.

private/private.h
```diff
@@ -28,7 +28,12 @@
 #define __DISPATCH_PRIVATE__
 
 #ifdef __APPLE__
+#include <Availability.h>
+#include <os/availability.h>
 #include <TargetConditionals.h>
+#include <os/base.h>
+#else
```
das: shouldn't this be `#elif defined(__linux__)` like in dispatch/dispatch.h, these headers are included by other platforms as well...

das:
@swift-ci please test

private/private.h
compnerd: Ah, good idea; I would've hit this the next time I did a windows build!

das:
CI is happy: https://ci.swift.org/job/swift-corelibs-libdispatch-PR-Linux/140/

merging

PR #295, opened 2017-08-26
Track armed events in muxnotes closely

MadCoder:
This fixes issues around resuming or re-registering file-descriptors with
a previously triggered EPOLLONESHOT event.

Signed-off-by: Pierre Habouzit <phabouzit@apple.com>
Fixes: SR-5759

### DIFF
--- a/src/event/event_epoll.c
+++ b/src/event/event_epoll.c
@@ -47,12 +47,13 @@ typedef struct dispatch_muxnote_s {
 	TAILQ_ENTRY(dispatch_muxnote_s) dmn_list;
 	TAILQ_HEAD(, dispatch_unote_linkage_s) dmn_readers_head;
 	TAILQ_HEAD(, dispatch_unote_linkage_s) dmn_writers_head;
-	int     dmn_fd;
-	uint32_t dmn_ident;
-	uint32_t dmn_events;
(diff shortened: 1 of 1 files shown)


MadCoder:
@swift-ci please test

MadCoder:
I'd appreciate a review from someone, @weissi (who is on vacation atm) has indepdently checked that it fixes SR-5759 for him.

the problem with epoll is that what is two different filters in the kevent world (EVFILT_READ/EVFILT_WRITE) is really a single one in epoll.

the multiplexing code which is a simplified version of the kevent one assumes that the first source that is resumed, resumes the multiplexing note, and that delivery disables it fully. However this is not quite how it works, we need to be able to do partial resumes/disarm of EPOLLIN and EPOLLOUT independently as all write sources of a given fd could be suspended while read sources are active and vice versa.

MadCoder:
Note: this is slated for swift4 too

dgrove-oss:
was out on vacation last week; will take a look at this tomorrow (unburying from other stuff today).  Sorry for the delay.

MadCoder:
@dgrove-oss no worries, this isn't even a day old ;) I'll wait for you to have a look then.

PR #296, opened 2017-08-27
Fix building on armv6

zwaldowski:
Applies same change from #155.


MadCoder:
@swift-ci please test

MadCoder:
you may want to check if it builds with cmake too as I understand cmake is the future and autoconf is on life support.

zwaldowski:
My (admittedly rather thin) understanding of CMake is that neither problem exists there. When built within the Swift tree, it'll either pass or inherit the correct arch-canonicalized install paths into Dispatch's cmake step like it does to produce other overlays.

PR #297, opened 2017-08-27
[4.0] Fix building on armv6

zwaldowski:
See #296. I don't know if this is kosher as part of the release schedule, but since it's such a trivial change I figure it'd be worth a shot.


MadCoder:
it's fine, I'll probably not merge this but force swift-4.0-branch to be master as they haven't diverged yet.

MadCoder:
(was merged by pushing master to switf4)

PR #300, opened 2017-09-07
SR-4677: avoid SIGSEGV if _dispatch_muxnote_create returns NULL

dgrove-oss:
It is possible for _dispatch_muxnote_create to return NULL, this error
condition was not being handled in dispatch_unote_register leading to
an immediate SIGSEGV when it happened.


MadCoder:
hah, good one.

MadCoder:
(I also synced swift 4 with master since they haven't diverged yet)

PR #301, opened 2017-09-15
Fix overflow traps in DispatchTime/DispatchWallTime/DispatchTimeInterval

ktopley-apple:
This is the swift-corelibs-libdispatch version of the PR at https://github.com/apple/swift/pull/11927.

Fixes overflows in DispatchTime/DispatchWallTime/DispatchTimeInterval.

Examples:

DispatchTime.now() + Date.distantFuture.timeIntervalSinceNow // traps
DispatchTime.now() + Date.distantPast.timeIntervalSinceNow. // traps
let t = DispatchTimeInterval.seconds(Int.max)
t == t // Traps due to conversion to nanoseconds.

rdar://problem/34440944

### DIFF
--- a/src/swift/Time.swift
+++ b/src/swift/Time.swift
@@ -119,6 +119,40 @@ public func ==(a: DispatchWallTime, b: DispatchWallTime) -> Bool {
 	return a.rawValue == b.rawValue
 }
 
+// Returns m1 * m2, clamped to the range [Int64.min, Int64.max].
+// Because of the way this function is used, we can always assume
+// that m2 > 0.
+private func clampedInt64Product(_ m1: Int64, _ m2: Int64) -> Int64 {
+	assert(m2 > 0, "multiplier must be positive")
+	let (result, overflow) = m1.multipliedReportingOverflow(by: m2)
+	if overflow {
+		return m1 > 0 ? Int64.max : Int64.min
+	}
+	return result
+}
+
+// Returns its argument clamped to the range [Int64.min, Int64.max].
+private func toInt64Clamped(_ value: Double) -> Int64 {
+	if value.isNaN { return Int64.max }
+	if value >= Double(Int64.max) { return Int64.max }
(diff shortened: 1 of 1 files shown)


ktopley-apple:
@swift-ci please test

src/swift/Time.swift
```diff
@@ -119,6 +119,40 @@ public func ==(a: DispatchWallTime, b: DispatchWallTime) -> Bool {
 	return a.rawValue == b.rawValue
 }
 
+// Returns m1 * m2, clamped to the range [Int64.min, Int64.max].
+// Because of the way this function is used, we can always assume
+// that m2 > 0.
+private func clampedInt64Product(_ m1: Int64, _ m2: Int64) -> Int64 {
+	assert(m2 > 0, "multiplier must be positive")
```
moiseev: `assert` is a noop in release builds. Should perhaps use `_precondition`?

src/swift/Time.swift
```diff
@@ -119,6 +119,40 @@ public func ==(a: DispatchWallTime, b: DispatchWallTime) -> Bool {
...
+	let (result, overflow) = m1.multipliedReportingOverflow(by: m2)
+	guard overflow == false else {
```
moiseev: This reads a little backwards to me. What do you think about the following instead?

```swift
if overflow {
  return m1 > 0 ? Int64,max : Int64.min
}
return result
```

src/swift/Time.swift
ktopley-apple: Not having the check in release builds is exactly what I want. The only calls to this method are internal to this file and the intent is only to document the assumption and to catch (admittedly unlikely) future changes that violate it.

src/swift/Time.swift
ktopley-apple: Fixed.

ktopley-apple:
@swift-ci please test

ktopley-apple:
@swift-ci please test

PR #303, opened 2017-09-17
build: fix symlink creation

compnerd:
This adjusts the symlink creation so that it always occurs.  This also
allows us to ensure that we wire up the dependency for the swift module
to the symlink creation.  The CMake based build would fail without this
when trying to build dispatch for SourceKit on Linux.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -201,21 +201,19 @@ if(leaks_EXECUTABLE)
 endif()
 
 if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
-  add_custom_command(OUTPUT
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @das

compnerd:
CC @dgrove-oss so, the swift side of the change is ready to go after this change.  We are two commits away from all of the non-Darwin targets being ready to switch over to the CMake build!

compnerd:
Friendly ping @das (for a work day).  This is the last patch holding up the switch over to CMake for the integration with swift.

MadCoder:
@das is on vacation ;)

compnerd:
@MadCoder oh, haha!  Thanks for the assist!

PR #304, opened 2017-09-19
docs: update documentation to show how to use CMake

compnerd:
Update the documentation to reference the CMake build infrastructure
rather than autotools for the non-Darwin targets.  Once the Darwin build
is switched over, the entire build process can be described in a single
block.


compnerd:
CC: @MadCoder @dgrove-oss

dgrove-oss:
LGTM.  Tested Linux instructions.

compnerd:
@MadCoder any thing else to add to this?

PR #305, opened 2017-09-21
Properly pick up new events when reusing a muxnote

jblache:
Changes for SR-5759 introduced a bug where disarmed events got dropped when
a muxnote is reused, due to erroneously setting dmn->dmn_events to the
combination of new events and currently armed events.

Combining existing events with the combination of new events and armed events
implements the intended behavior.

### DIFF
--- a/src/event/event_epoll.c
+++ b/src/event/event_epoll.c
@@ -270,7 +270,7 @@ _dispatch_unote_register(dispatch_unote_t du,
 			if (_dispatch_epoll_update(dmn, events, EPOLL_CTL_MOD) < 0) {
 				dmn = NULL;
 			} else {
-				dmn->dmn_events = events;
+				dmn->dmn_events |= events;
 				dmn->dmn_disarmed_events &= ~events;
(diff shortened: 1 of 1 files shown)


src/event/event_epoll.c
```diff
@@ -270,7 +270,7 @@ _dispatch_unote_register(dispatch_unote_t du,
 			if (_dispatch_epoll_update(dmn, events, EPOLL_CTL_MOD) < 0) {
 				dmn = NULL;
 			} else {
-				dmn->dmn_events = events;
+				dmn->dmn_events |= events;
```
MadCoder: 🤦‍♂️

MadCoder:
@swift-ci please test

src/event/event_epoll.c
MadCoder: you gotta love 1-character fixes

MadCoder:
Seems the buildbot is broken, however the fix is obviously correct, so let's merge.

PR #306, opened 2017-09-21
Attempt to be better with signed conversions

compnerd:
Most targets use unsigned chars.  Explicitly change the static tables to
`signed char`.  An alternative  would be to force signed chars during
the compilation of libdispatch.  When building for Linux AArch64, the
use of `-1` causes signed conversion warnings when initializing the
table.


compnerd:
CC: @MadCoder

compnerd:
CC: @das

gottesmm:
@swift-ci smoke test

gottesmm:
@swift-ci test

PR #308, opened 2017-10-10
build: add a dependency on the swift compiler

compnerd:
Add a dependency on the swift compiler to ensure that a change to the
compiler will rebuild the swift portions.


compnerd:
CC @gottesmm @das @MadCoder

jrose-apple:
This definitely makes sense to me.

gottesmm:
This is just a cmake thing and should be safe to take. It is just adding a dependency on the swift compiler changing, so it shouldn't affect anyones build. Just for safety I will test.

gottesmm:
@swift-ci test

PR #311, opened 2017-10-11
Fixes overflow trap when creating DispatchTime objects with large upt…

ktopley-apple:
…imeNanoseconds values.

(rdar://problem/34922683)

[related to change https://github.com/apple/swift/pull/12387]

The change is in three parts:

The value UInt64.max passed to the DispatchTime initializer means DISPATCH_TIME_FOREVER and should not be scaled using the Mach timebase.
The scaling calculations need to detect overflow and clamp to UInt64.max if this occurs, making the value equal to DISPATCH_TIME_FOREVER
The getter for the uptimeNanoseconds also needs to be protected against overflow and return UInt64.max if it occurs.

### DIFF
--- a/src/swift/Time.swift
+++ b/src/swift/Time.swift
@@ -59,9 +59,13 @@ public struct DispatchTime : Comparable {
 	public init(uptimeNanoseconds: UInt64) {
 		var rawValue = uptimeNanoseconds
 #if HAVE_MACH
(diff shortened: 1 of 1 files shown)


ktopley-apple:
@swift-ci please test

ktopley-apple:
@swift-ci please test

ktopley-apple:
Squashed commits

PR #314, opened 2017-10-18
Fix the install path for the toolchain

aciidb0mb3r:
https://bugs.swift.org/browse/SR-6084
<rdar://problem/35040697> [SR-6084]: libdispatch missing from Trunk Development snapshots

### DIFF
--- a/dispatch/CMakeLists.txt
+++ b/dispatch/CMakeLists.txt
@@ -14,12 +14,12 @@ install(FILES
           source.h
           time.h
         DESTINATION
-          ${CMAKE_INSTALL_FULL_INCLUDEDIR}/dispatch)
+          "${CMAKE_INSTALL_LIBDIR}/swift/dispatch")
 if(ENABLE_SWIFT)
   get_filename_component(MODULE_MAP module.modulemap REALPATH)
   install(FILES
             ${MODULE_MAP}
           DESTINATION
-            ${CMAKE_INSTALL_FULL_INCLUDEDIR}/dispatch)
+            "${CMAKE_INSTALL_LIBDIR}/swift/dispatch")
 endif()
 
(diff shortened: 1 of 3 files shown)


shahmishal:
//cc @compnerd @dgrove-oss

shahmishal:
https://github.com/apple/swift/pull/12502

@swift-ci test

aciidb0mb3r:
@compnerd updated!

shahmishal:
apple/swift#12502

@swift-ci test

CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ enable_testing()
 set(CMAKE_C_VISIBILITY_PRESET hidden)
 set(CMAKE_CXX_STANDARD 11)
 
+set(CMAKE_INSTALL_LIBDIR lib CACHE PATH "Installation directory for libraries")
```
aciidb0mb3r: @compnerd Please comment on the PR and not the commit as force pushing removes them.

This is needed because its value is `lib/x86-gnu-linux/` (or similar), where as we need to install the libraries at `lib/`

compnerd: `GNUInstallDirs` will set the value to `lib` or `lib64` depending on the system that you are building for.  The only place that the `CMAKE_INSTALL_LIBDIR` uses the form that you specify is when you are on Debian as that is the style that Debian uses.  You should be able to override that from the command line.  I think that is better handled in `build-script` rather than hardcoding it here.

See https://cmake.org/cmake/help/v3.4/module/GNUInstallDirs.html

aciidb0mb3r: I see. I don't know why it sets that path for Ubuntu then. Sure, I can override it in the build-script if that is preferable.

aciidb0mb3r: Alright, done!

aciidb0mb3r:
apple/swift#12502

@swift-ci test

aciidb0mb3r:
apple/swift#12502

@swift-ci test

PR #320, opened 2017-11-12
Base approximate time on a monotonic clock, like absolute time

jblache:
Caught this today; _dispatch_approximate_time() uses CLOCK_REALTIME on Linux for some reason, while _dispatch_absolute_time() uses CLOCK_MONOTONIC. This seems wrong and I can't think of a reason why this would be desirable in the first place.

Darwin code uses the same clock family for both, so use CLOCK_MONOTONIC_COARSE on Linux as the approximate time source, and test for it during configure.

### DIFF
--- a/src/shims/time.h
+++ b/src/shims/time.h
@@ -121,6 +121,19 @@ _dispatch_get_nanoseconds(void)
 #endif
 }
 
+/* On the use of clock sources in the CLOCK_MONOTONIC family
+ *
+ * The code below requires monotonic clock sources that only tick
+ * while the machine is running.
+ *
+ * Per POSIX, the CLOCK_MONOTONIC family is supposed to tick during
+ * machine sleep; this is not the case on Linux, and that behavior
+ * became part of the Linux ABI.
+ *
+ * Using the CLOCK_MONOTONIC family on POSIX-compliant platforms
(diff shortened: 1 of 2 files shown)


src/shims/time.h
```diff
@@ -152,9 +152,9 @@ _dispatch_approximate_time(void)
 	struct timespec ts;
 	dispatch_assume_zero(clock_gettime(CLOCK_UPTIME_FAST, &ts));
 	return _dispatch_timespec_to_nano(ts);
-#elif defined(__linux__)
+#elif HAVE_DECL_CLOCK_MONOTONIC_COARSE
```
MadCoder: please keep the `if defined(__linux__)` because CLOCK_MONOTONIC is supposed to tick during machine sleeps if you look at POSIX, and it happens to not be the case on linux which eventually became ABI on this platform.

However if a platform has (1) MONOTONIC_COARSE and not UPTIME_FAST, and (2) is POSIX compliant, then this would cause a bug for the said platform.

However you're right that MONOTONIC_COARSE is what we want here.

jblache:
Amended; do we want to do the same in _dispatch_absolute_time() with a comment on top?

MadCoder:
we probably should now that you point this out.

jblache:
How does that look? Ship it?

PR #325, opened 2017-12-04
Fully initialize ev.data.

Lukasa:
On Linux systems, ev.data is an 8-byte field, but this code currently initializes only 4 bytes of the field, leaving the trailing 4 bytes unininitialized. This causes memory safety tools like Valgrind to complain, as noted in #321.

This patch no longer explicitly initializes `.data`, ensuring that C structure initialization will force the `.data` field to be zero-initialized. Manual checks of the compiled binary show that the compiler correctly initializes the 4 bytes for `.u32` only once, so this change incurs minimal overhead (specifically, setting the next 4 bytes to 0).

### DIFF
--- a/src/event/event_epoll.c
+++ b/src/event/event_epoll.c
@@ -391,13 +391,14 @@ _dispatch_timeout_program(uint32_t tidx, uint64_t target,
 	dispatch_epoll_timeout_t timer = &_dispatch_epoll_timeout[clock];
 	struct epoll_event ev = {
 		.events = EPOLLONESHOT | EPOLLIN,
-		.data = { .u32 = timer->det_ident },
+
 	};
 	int op;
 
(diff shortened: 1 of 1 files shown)


MadCoder:
@swift-ci please test

alblue:
@swift-ci please test

Lukasa:
@MadCoder Sorry, I completely missed your feedback until just now! Would you prefer that I update the dispatch codebase to use `u64` instead? I agree that it'll clean things up, but it'll be a much more far-reaching patch.

MadCoder:
no that's fine, the patch you did is fine too

MadCoder:
what is happening to the bots now...

@swift-ci please test

MadCoder:
@swift-ci please test

shahmishal:
@swift-ci test

PR #326, opened 2017-12-14
Add special check for Fedora flavor of Linux to disable dtrace support

tachoknight:
`dtrace` support is not officially available on Fedora Linux, unlike Ubuntu; this prevents cmake from completing successfully. This pull requests adds a separate module with a function to get the **ID** field from /etc/os-release, which is available as part of any Linux distribution that uses systemd. The ID field returns the short name of the Linux distribution (e.g. "fedora", "ubuntu").

The modification to CMakeLists.txt checks that, if on a Linux-based system, check whether the distribution is Fedora and, if it is, set `ENABLE_DTRACE_DEFAULT` to OFF, otherwise ON.


MadCoder:
Can't we do it by detecting if the proper dtrace binaries & headers are present, that would be more generic...

tachoknight:
That seems like a more reasonable way to go; I'll rework it and submit a new pull request.

PR #327, opened 2017-12-14
Perform test for dtrace program when compiling under Linux

tachoknight:
The Fedora flavor of Linux does not include dtrace, while Ubuntu does. This pull request checks if we're building under Linux and, if we are, test for the presence of the dtrace program. If found it will set the `ENABLE_DTRACE_DEFAULT` variable to ON, otherwise OFF.


tachoknight:
Going to resubmit on a cleaner copy of the project

PR #328, opened 2017-12-27
Check for dtrace and only enable its use if found

tachoknight:
The Fedora flavor of Linux does not include dtrace, while Ubuntu does. This pull request checks for the presence of the dtrace program. If found it will set the ENABLE_DTRACE_DEFAULT variable to ON, otherwise OFF.


tachoknight:
I changed the title of the pull request to reflect the fact that we should check for dtrace regardless of whether we're on Linux or not.

PR #312, opened 2017-10-11
build: improve cross-compilation for android

compnerd:
The sysroot headers for android require some definitions to be altered.
Ensure that we do not define `_GNU_SOURCE` for Android.

PR #332, opened 2018-01-18
Fix build for android

ephemer:
As mentioned in #312, this fixes the build for Android and works on both Mac and Linux with the following CMake invocation:

```
cmake -G Ninja \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_SYSTEM_NAME=Android \
    -DCMAKE_SYSROOT=$ANDROID_STANDALONE_TOOLCHAIN/sysroot \
    -DENABLE_SWIFT=YES \
    -DBUILD_SHARED_LIBS=YES \
    -DCMAKE_SWIFT_COMPILER=$SWIFT_INSTALLATION_PATH/usr/Darwin/swiftc \
    -DENABLE_TESTING=OFF \
    -DCMAKE_ANDROID_ARCH_ABI=armeabi-v7a \
    -DCMAKE_ANDROID_NDK_TOOLCHAIN_VERSION=clang \
    $SWIFT_SOURCE_PATH/swift-corelibs-libdispatch
```


dispatch/dispatch.h
```diff
@@ -30,6 +30,8 @@
 #include <os/linux_base.h>
 #endif
 
+#include <string.h>
```
ephemer: This should probably be wrapped with an #if ANDROID

cmake/modules/SwiftSupport.cmake
```diff
@@ -38,6 +38,10 @@ function(add_swift_library library)
     endforeach()
   endif()
 
+  if(CMAKE_SYSTEM_NAME STREQUAL Android)
+    list(APPEND flags -sdk;${CMAKE_SYSROOT})
```
ephemer: It probably makes sense to always add the sysroot?

cmake/modules/SwiftSupport.cmake
```diff
@@ -13,7 +13,7 @@ function(add_swift_library library)
   list(APPEND flags -emit-library)
 
   if(ASL_TARGET)
-    list(APPEND FLAGS -target;${ASL_TARGET})
+    list(APPEND flags -target;${ASL_TARGET})
```
ephemer: apparently CMake variable names are case sensitive. This is a typo introduced in an earlier PR

MadCoder:
@compnerd you did most of the cmake porting, can you review the change please?

dispatch/dispatch.h
```diff
@@ -30,6 +30,10 @@
 #include <os/linux_base.h>
 #endif
 
+#ifdef __ANDROID__
+#include <string.h>
```
compnerd: I think that this can be done unconditionally.

cmake/modules/SwiftSupport.cmake
compnerd: Yeah, I think that we should always add the sysroot here.

compnerd:
Can you please squash the changes into a single commit?

dispatch/dispatch.h
```diff
@@ -30,6 +30,8 @@
 #include <os/linux_base.h>
 #endif
 
+
```
MadCoder: please remove the empty line

cmake/modules/SwiftSupport.cmake
```diff
@@ -93,7 +95,7 @@ function(get_swift_host_arch result_var_name)
     set("${result_var_name}" "s390x" PARENT_SCOPE)
   elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "armv6l")
     set("${result_var_name}" "armv6" PARENT_SCOPE)
-  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "armv7l")
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "^armv7")
```
MadCoder: do we want the same for armv6l above?

src/shims/android_stubs.h
```diff
@@ -20,4 +20,8 @@
 #define _Static_assert(...)
 #endif
 
+#ifndef TRASHIT
```
MadCoder: I'm fairly sure this has been moved to `internal.h` already in a previous PR, please remove.

MadCoder:
Also please squash in a single commit.

Fischiii:
@MadCoder can you elaborate? We are successfully working with this fork, since we need the android support. Would be nice to have this merged.

compnerd:
I'm going to close this PR for now.  It no longer cleanly applies.  I believe that libdispatch builds for android currently, but, if it does not, that is certainly that should be addressed.

PR #338, opened 2018-02-20
shims: improve Windows CPU configuration scanning

compnerd:
Use the newer Windows APIs to detect the windows CPU state.  This allows
us to collect all of the CPU configuration information to properly
affinitise work to the preferred CPU.


compnerd:
CC: @MadCoder @das @parkera

compnerd:
Okay, I have a much larger patch that this is a part of now, and does a bunch of windows specific stuff, I'll go ahead and roll it into a single patch.

PR #340, opened 2018-02-21
build: require C11

compnerd:
The atomics support used in libdispatch requires C11.  Ensure that the
compiler supports it and that the support is enabled.


compnerd:
@MadCoder @das the source seem to indicate that C11 is required (primarily for the atomics support).  This change just ensures that the compiler supports it and is used.

compnerd:
@swift-ci please test

MadCoder:
yup

PR #341, opened 2018-02-21
dispatch: adjust for LLP64 environments

compnerd:
LLP64 environments use a 32-bit long value.  This would result in
truncation as the `long` is used to transfer a pointer value here.
Adjust the types accordingly.  Since the width on LP64 environments is
equivalent in this change, this should only impact LLP64 environments
and be ABI stable otherwise.

### DIFF
--- a/dispatch/once.h
+++ b/dispatch/once.h
@@ -38,7 +38,7 @@ __BEGIN_DECLS
  * Note: static and global variables default to zero.
  */
 DISPATCH_SWIFT3_UNAVAILABLE("Use lazily initialized globals instead")
-typedef long dispatch_once_t;
+typedef intptr_t dispatch_once_t;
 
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @das @MadCoder

compnerd:
@swift-ci please test

dispatch/once.h
```diff
@@ -38,7 +38,7 @@ __BEGIN_DECLS
  * Note: static and global variables default to zero.
  */
 DISPATCH_SWIFT3_UNAVAILABLE("Use lazily initialized globals instead")
-typedef long dispatch_once_t;
+typedef intptr_t dispatch_once_t;
```
MadCoder: should really be `uintptr_t` I don't like signed integers for these.

MadCoder: oh right and since it's API it may cause build fallouts... ok convinced let's keep intptr_t ugh

PR #342, opened 2018-02-21
BlocksRuntime: adjust implementation for Windows x64

compnerd:
This adjust the BlocksRuntime to work on Windows x64 which is an LLP64
environment.  This primarily comes down to using `size_t` for the size
parameter for the allocation routines.  A secondary change which is
needed is the type cast for the `InterlockedCompareExchange` operation
on Windows.  A small cleanup is included which removes the definition of
an unused function.

### DIFF
--- a/src/BlocksRuntime/runtime.c
+++ b/src/BlocksRuntime/runtime.c
@@ -28,25 +28,22 @@
 #define os_assert(_x) assert(_x)
 #endif
 
-#if TARGET_OS_WIN32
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @das @MadCoder @parkera

compnerd:
@swift-ci please test

compnerd:
A similar change is in apple/swift-corelibs-foundation for this, but, since both are used, we should still get this merged.

PR #343, opened 2018-02-22
workaround `-Wassign-enum` warnings

compnerd:
There are a number of cases where we use `0` or macro definitions for
values to enumerated types.  When building for Windows, these cause a
warning due to the value not being part of the enumeration.  This change
silences the warnings and identifies the various places where this
occurs.  Ideally, the values would be part of the enumerations.

### DIFF
--- a/src/apply.c
+++ b/src/apply.c
@@ -171,7 +171,10 @@ _dispatch_apply_f2(dispatch_queue_t dq, dispatch_apply_t da,
 		dispatch_continuation_t next = _dispatch_continuation_alloc();
 		uintptr_t dc_flags = DISPATCH_OBJ_CONSUME_BIT;
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wassign-enum"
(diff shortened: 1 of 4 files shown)


compnerd:
@das @MadCoder I don't know of a better way to deal with these.  If you have a suggestion on how to handle this better, I'd love to hear about it.

MadCoder:
I dislike this patch strongly, enums have been used for "flags" for times imemorial, clang has to provide a way to make it clear that's what we're doing. I'd rather disable `-Wassign-enum` that clutter the code with this

compnerd:
The problem isn't that the enum is being used for flags, but rather than there are values being used that are not in that enumeration.  It just wants the rest of the flags to be part of the enumeration.

MadCoder:
we should disable the warning, I don't understand what value it adds

compnerd:
Sure, I don't have a complaint against that.  I'm not sure that this would add much value.  I'll upload a different change.

PR #344, opened 2018-02-22
revive the Windows port

compnerd:
The `DISPATCH_EXPORT` macro added an extraneous `extern` on the
declaration.  This was pointed out by a newer clang.  Adjust the
declaration accordingly.

Add build rules to improve Windows builds.  These additions help
cross-compile to Windows with both clang-cl as well as the GNU clang
driver.

Use the newer Windows APIs to detect the windows CPU state.  This allows
us to collect all of the CPU configuration information to properly
affinitise work to the preferred CPU.

Use the FLS API to create thread local data storage for libdispatch's
queues.


compnerd:
CC: @das @MadCoder @jrose-apple @parkera

compnerd:
This is a partial patch, it still needs more work.  But, it sets up a lot of the work that is needed.  It also fixes cross-compilation on Linux to Windows.  The assumption here is that you are using `clang-cl` or `clang` to build libdispatch.  I figure that putting this up here earlier is better as it allows others to take a look as well as possibly solve some of the remaining issues.

src/shims/tsd.h
```diff
@@ -99,6 +99,40 @@ _dispatch_thread_key_create(const unsigned long *k, void (*d)(void *))
...
+	void *dispatch_bcounter_key;
+#endif
+	void *dispatch_priority_key;
+	void *dispatch_r2k_key;
+	void *dispatch_wlh_key;
+	void *dispatch_voucher_key;
+	void *dispatch_deferred_items_key;
+};
+
+extern __declspec(thread) struct dispatch_tsd __dispatch_tsd;
```
MadCoder: can't you try to reuse `DISPATCH_USE_THREAD_LOCAL_STORAGE` it's supposed to be the same trick, no? instead of duplicating the whole struct

MadCoder:
I have to look at the patch more, and it's more than I can do in a timely fashion, but I've seen nothing shocking in there overall.

src/shims/tsd.h
compnerd: Sure, its actually the same exact structure too ... just didn't move things around to make it easier.  I can do that.

os/generic_win_base.h
```diff
@@ -0,0 +1,129 @@
...
+#ifndef __OS_GENERIC_WIN_BASE__
+#define __OS_GENERIC_WIN_BASE__
+
+// Unices provide `roundup` via sys/param.h
+#define roundup(x, y) ((((x) + ((y) - 1)) / (y)) * (y))
+// Unices provide `MAX` via sys/param.h
+#define MAX(a,b) (((a)>(b))?(a):(b))
+// Unicies provide `MIN` via sys/param.h
+#define MIN(a,b) (((a)<(b))?(a):(b))
+// Unicies provide `howmany` via sys/param.h
```
jrose-apple: Nitpick: inconsistent between `Unices` and `Unicies`.

src/block.cpp
```diff
@@ -69,7 +69,8 @@ struct dispatch_block_private_data_s {
 	{
 		// copy constructor, create copy with retained references
 		if (dbpd_voucher) voucher_retain(dbpd_voucher);
-		if (o.dbpd_block) dbpd_block = _dispatch_Block_copy(o.dbpd_block);
+		if (o.dbpd_block) dbpd_block =
```
MadCoder: please use braces

private/private.h
```diff
@@ -32,7 +32,9 @@
 #include <os/availability.h>
 #include <TargetConditionals.h>
 #include <os/base.h>
-#elif defined(__linux__) || defined(__FreeBSD__)
+#elif defined(_WIN32)
```
MadCoder: this should use `TARGET_OS_WIN32` to be consistent with the rest of the tests I think

src/introspection.c
```diff
@@ -137,7 +137,7 @@ _dispatch_introspection_thread_add(void)
 	_dispatch_unfair_lock_unlock(&_dispatch_introspection.threads_lock);
 }
 
-static void
+static __stdcall void
```
MadCoder: is `__stdcall` defined by clang on darwin or linux?

jrose-apple: I get

```
<stdin>:1:6: warning: calling convention '__stdcall' ignored for this target
      [-Wignored-attributes]
void __stdcall test();
     ^
```

So yeah, it might be something you want to guard.

compnerd: Sure, I can control this by a macro

compnerd:
Updated to address the comments.  This is further along now, with a lot more of the port formalized.

src/internal.h
```diff
@@ -283,13 +283,19 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
...
 #include <string.h>
 #if defined(__unix__) || (defined(__APPLE__) && defined(__MACH__))
 #include <unistd.h>
 #endif
+#if TARGET_OS_WIN32
+#include <io.h>
+#endif
 
-#if __GNUC__
+#if __GNUC__ || defined(__clang__)
```
MadCoder: `#if defined(__GNUC__) || defined(__clang__)` would look better ;)

compnerd:
There is a small amount of porting left: src/io.c (this previously was completely disabled).

MadCoder:
if TARGET_OS_WIN32 comes from the build system you can't have it in private.h or any header, you should use _WIN32 there, and then it is a question whether we need TARGET_OS_WIN32 at all, it seems confusing. TARGET_OS* comes from <TargetConditionals.h> which is a thing that Apple has when we build things on Windows platforms that's why it's this way. I don't think it should stay this way.

have you ported all of event/event.c too? ;) because that part is the one that needs basically full reimplementation, unless windows supports kevent :P (not that it should block anything, but between having dispatch sources working and io.c I know what should be done first :P)

MadCoder:
so please get rid of TARGET_OS_WIN32 consistently, we use `__linux__` and `__freebsd__` and such for other platforms, there's no reason for windows to be in between IMO. Once you've done that I'll merge.

compnerd:
Hah, `kevent.c` is huge (and as you pointed out, will require a full re-implementaiton), and I think that doing that as a separate change is best -- this is a pretty big change as is.  While I agree that `kevent.c` is crucial, having this part means that we can at least ensure that build regressions don't occur while that is ported over.  Replacing `TARGET_OS_WIN32` with `defined(_WIN32)` sounds completely fine to me.

MadCoder:
as far as kevent.c is, it's huge, but you should instead model it after `event_epoll.c` which is way smaller, I have abstracted platform specific code this way way more than it used to be.

compnerd:
So, the build should be complete at this point, the top patch is the only thing which I don't know how to formalize.  But, the entire build now completes successfully :-)

src/object_internal.h
```diff
@@ -340,7 +340,7 @@ DISPATCH_ENUM(dispatch_invoke_flags, uint32_t,
 #define _DISPATCH_INVOKE_AUTORELEASE_MASK	  0x00300000u
 );
 
-enum {
+enum : unsigned long {
```
MadCoder: this syntax is not supported by C mode on clang I think, can you use DISPATCH_ENUM() instead?

compnerd: Done

MadCoder:
@swift-ci please test

MadCoder:
@swift-ci please test

PR #345, opened 2018-03-09
Add support for 32-bit Ubuntu

ionel71089:
Add support for 32-bit ubuntu by comparing the architecture in `SwiftSupport.cmake` and adding the `__mulodi4` function in `src/shims.c` to fix "undefined reference to __mulodi4" linker error.


src/shims.c
```diff
@@ -33,3 +33,46 @@ size_t strlcpy(char *dst, const char *src, size_t size)
 	return res;
 }
 #endif
+
+#if __i386__ && !defined(__APPLE__)
```
MadCoder: why? the compiler should provide this, this seems the wrong place to put this.

ionel71089: I am getting `undefined reference to '__mulodi4` in the linker step when buidling on 32-bit Ubuntu. 
This is apparently caused by a bug in llvm (see https://llvm.org/bugs/show_bug.cgi?id=28629 and https://gist.github.com/uraimo/33715886cef21c37149b6c712fed14a8). I've similarly solved this for the swift standard library by adding the function in `stdlib/public/stubs/Stubs.cpp` (see https://github.com/apple/swift/blob/9f0d19f69948f2d99e8523724768785c1d37204f/stdlib/public/stubs/Stubs.cpp) where `__muloti4` is also copied (both functions are from compiler-rt).

`shims.c` seems the appropriate place to put the function in this case. Should I move it to another file or does it require a different solution?

MadCoder: wow I forgot about this request and didn't see the reply sorry.

why isn't llvm fixing their bug then?

ionel71089: I honestly don't know. 
Sadly I haven't done any work related to this in the last couple of months... 
(Was trying to make swift run on a i586, got it running on Ubuntu 32-bit)
So I think I'll just have to cancel this pull request, I guess...

Kaiede: FYI, I've gotten a PR in Swift itself that exposes __mulodi4 to 386 like it does for arm3: https://github.com/apple/swift/pull/19607/commits/eddecbfcd43d72e33fa9788b484424ec2eaa5036

You really don't need to add it here. But the root problem is that the llvm "bug" is that on Linux, Swift doesn't require the use of compiler-rt, which is normally where this symbol would come from, and there's no other source for it from glibc or similar on these 32-bit platforms. The work around is that Swift itself is providing the symbol for the time being to avoid the dependency on compiler-rt. :(

Kaiede:
I've also been trying to stand up 32-bit Linux as a host platform. https://github.com/apple/swift-corelibs-libdispatch/pull/345/commits/bf9b1165101f09de12af5f76a98d0e6114d7aaa3 from this PR would really help me (the other commit isn't needed). But I can open up a PR with just the CMake changes if you aren't actively looking here @ionel71089.

src/shims.c
ionel71089: Hello, @Kaiede. Please create the CMake PR you mentioned in the other comment. Also, could you create a post on the forums about your progress on 32-bit Linux support? I've seen a few other people interested in this topic. Thanks.

@MadCoder Let's close this PR, since it probably won't be needed anymore.

Kaiede: I could probably split it out from the thread where I have been discussing the blocking issues on Armv7 (and i686) trying to get 4.2 and master building.

PR #346, opened 2018-03-10
[gardening] Use `Optional.map` over `flatMap` where appropriate

ikesyo:
Ref: https://github.com/apple/swift/pull/15142

---

The initializers used here are not optional ones, so there is no need to use `flatMap`.

- https://github.com/apple/swift-corelibs-libdispatch/blob/9295346d20e52c9ff4d82906fd85287d8636c542/src/swift/Data.swift#L100-L102
- https://github.com/apple/swift-corelibs-libdispatch/blob/9295346d20e52c9ff4d82906fd85287d8636c542/src/swift/Queue.swift#L22-L25


ikesyo:
@swift-ci please test

PR #348, opened 2018-03-15
SR-2100 - Add support for multi-lib systems

capent:
[SR-2100 - Add support for multi-lib systems](https://bugs.swift.org/browse/SR-2100)

Use the new SWIFT_LIBDIR cmake variable defined by swift's build-script (pending pull request), so
Linux distributions using lib64 instead of lib for 64-bit libraries could install
swift without the need for patches, like Fedora.


capent:
@MadCoder The intermediate merges are messing with git rebase, i will do a new pull request from a clean branch, better that way. Closing this one.

PR #349, opened 2018-03-20
[gardening] Use https for swift.org links

ikesyo:


ikesyo:
@swift-ci Please test

PR #351, opened 2018-03-29
build: add the default search paths for Windows

compnerd:
When building for Windows on Linux, we need to explicitly setup the
linker search paths.  These are derivable from the UCRT version and the
WinSDK base directories which are expected environment variables.  This
is also how the swift build system finds the directories.  Enable this
to avoid having the user explicitly setup the `LIB` environment variable
for the linker to find the system import libraries.


compnerd:
CC: @das @MadCoder

PR #353, opened 2018-04-13
SR-2100 - Add support for multi-lib systems

capent:
Supersedes #348. 

[SR-2100 - Add support for multi-lib systems](https://bugs.swift.org/browse/SR-2100)

Use the new SWIFT_LIBDIR cmake variable defined by swift's build-script (pending pull request), so
Linux distributions using lib64 instead of lib for 64-bit libraries could install
swift without the need for patches, like Fedora.

@MadCoder - Now a new clean branch from #348.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,8 @@ include(CheckSymbolExists)
 include(GNUInstallDirs)
 include(SwiftSupport)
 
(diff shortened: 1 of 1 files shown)


MadCoder:
@swift-ci please test

shahmishal:
@swift-ci please test

PR #358, opened 2018-06-05
Use _Thread_local for thread-local storage

adierking:
When building for Windows, shims/tsd.h declares __dispatch_tsd with
__declspec(thread). init.c needs to match this or else it will not compile with
newer versions of Clang.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -112,7 +112,7 @@ void (*_dispatch_end_NSAutoReleasePool)(void *);
 #endif
 
 #if DISPATCH_USE_THREAD_LOCAL_STORAGE
-__thread struct dispatch_tsd __dispatch_tsd;
+_Thread_local struct dispatch_tsd __dispatch_tsd;
 #if defined(_WIN32)
 DWORD __dispatch_tsd_key;
 #else
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,7 +102,7 @@ option(USE_GOLD_LINKER "use the gold linker" ${USE_GOLD_LINKER_DEFAULT})
 
 option(USE_LLD_LINKER "use the lld linker" OFF)
 
-option(ENABLE_THREAD_LOCAL_STORAGE "enable usage of thread local storage via __thread" ON)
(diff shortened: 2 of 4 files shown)


adierking:
cc @compnerd

src/init.c
```diff
@@ -112,10 +112,11 @@ void (*_dispatch_end_NSAutoReleasePool)(void *);
 #endif
 
 #if DISPATCH_USE_THREAD_LOCAL_STORAGE
-__thread struct dispatch_tsd __dispatch_tsd;
 #if defined(_WIN32)
+__declspec(thread) struct dispatch_tsd __dispatch_tsd;
```
MadCoder: would using `_Thread_local` work instead? I would assume it maps to the right thing on both platforms, and libdispatch otherwise already requires C11

adierking: MSVC actually doesn't support C11 threads, so using `__declspec(thread)` is the canonical method of declaring thread-local variables in Windows code. I agree that using `_Thread_local` everywhere would be best if we don't care about building with MSVC (I don't).

MadCoder: by MSVC you mean the MSFT C compiler as opposed to clang, correct? Not MSFT Visual Studio?

libdispatch needs clang (which is I'm told supported on Windows nowadays?) for blocks support anyway (and more). So yeah I don't think we care about MSVC in this instance, if `_Thread_local` does the job, we should use it, `__thread` really is a gcc-ism and C11 has the right thing.

adierking: >by MSVC you mean the MSFT C compiler as opposed to clang, correct?

Right.

I updated this to use `_Thread_local` now. Thanks for the guidance.

compnerd:
I think that using `_Thread_local` is fine.  We require C11, we require clang/clang-cl.  It's better to use the modern, standards approach.  LGTM

MadCoder:
@swift-ci please test

compnerd:
@swift-ci please teset

compnerd:
@swift-ci please test

PR #360, opened 2018-06-12
build: fix clang-cl exception flags

adierking:
When building with clang-cl, we currently pass the `/EHsc-` flag in an attempt to disable exceptions, but this is equivalent to `/EHs` which still enables C++ exceptions and trips the `#error` in block.cpp. clang-cl actually disables exceptions by default, so we don't need to pass anything.

See https://msdn.microsoft.com/en-us/library/1deeycx5.aspx.


adierking:
cc @compnerd

compnerd:
CC @madcoder

compnerd:
@swift-could please test

compnerd:
@swift-ci please test

PR #361, opened 2018-06-18
build: support system installed BlocksRuntime

compnerd:
This uses the standard CMake mechanism for finding the blocks runtime.
If one is not found, one will be provided for you.


compnerd:
CC: @MadCoder @millenomi

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

MadCoder:
What is the motivation here? I'm worried about random blocks runtimes, we've made fixes to the copy we have.

compnerd:
The swift instructions tell you to install and build against the blocks runtime (via apt-get).  This enables that.  It also allows you to switch between the version in s-f-l or dispatch or if you have a copy checked out (like libclosure) or from compiler-rt.

MadCoder:
mmk

PR #363, opened 2018-06-18
build: fix the clang-cl executable link command

adierking:
ClangClCompileRules.cmake overrides CMAKE_C_LINK_EXECUTABLE to invoke the C
compiler, but clang-cl does not accept linker flags, so tests will fail to link.
The default is fine, so delete this line.


adierking:
cc @MadCoder @compnerd

compnerd:
I guess that I had must've left it behind accidentally.

PR #364, opened 2018-06-19
Fixed handling of the AMD64 processor architecture on Windows

dplanitzer:
Invoking cmake on a system where MSVC (2017) sets the PROCESSOR_ARCHITECTURE environment variable to AMD64 would previously fail with this error:

`CMake Error at cmake/modules/DispatchWindowsSupport.cmake:12 (message):
  do not know MSVC spelling for ARCH: `AMD64`
Call Stack (most recent call first):
  CMakeLists.txt:331 (dispatch_windows_arch_spelling)`

We now treat AMD64 like x86_64.


dplanitzer:
cc @MadCoder @compnerd

compnerd:
@swift-ci please test

PR #365, opened 2018-06-19
build: correctly enable blocks for tests when using clang-cl

adierking:
clang-cl does not support -fblocks, so prefix it with -Xclang to pass it
directly to the frontend like src/CMakeLists.txt does.


adierking:
cc @MadCoder @compnerd

compnerd:
@swift-ci please test

PR #366, opened 2018-06-19
CMake: create_symlink isn't supported on Windows. Use copy instead

dplanitzer:
We previously tried to symlink the module maps with the create_symlink command. But this command isn't supported on Windows and thus fails. We now copy the files instead.


dplanitzer:
cc @MadCoder @compnerd

MadCoder:
@swift-ci please test

PR #367, opened 2018-06-19
implemented getprogname() for Windows

dplanitzer:
We use GetModuleFileName() to get the filename of the process.  InitOnceExecuteOnce() is used to do this once per app run.

### DIFF
--- a/src/shims/getprogname.c
+++ b/src/shims/getprogname.c
@@ -1,0 +1,72 @@
+/*
+ * Copyright (c) 2009-2010 Mark Heily <mark@heily.com>
+ * All rights reserved.
+ *
+ * @APPLE_APACHE_LICENSE_HEADER_START@
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * @APPLE_APACHE_LICENSE_HEADER_END@
+ */
+
+#include "getprogname.h"
+
+#if !HAVE_GETPROGNAME
+
+#if defined(_WIN32)
+#define WIN32_LEAN_AND_MEAN
+#ifndef _WIN32_WINNT
+#define _WIN32_WINNT 0x0600
+#endif /* _WIN32_WINNT */
+#include <windows.h>
+#include <stdlib.h>
+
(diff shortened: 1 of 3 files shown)


dplanitzer:
cc @MadCoder @compnerd

compnerd:
I thought that libbsd has an implementation of this function.  Can we not just use that?

adierking:
>I thought that libbsd has an implementation of this function.

Not for Windows :(

dplanitzer:
The lack of an actually functioning getprogname() implementation In libbsd on Windows is one problem that I see with the libbsd solution. But there are additional problems:

I've identified 3 major road blocks to making libdispatch work on Windows:

a) missing getprogname() implementation

b) missing TAILQ macros

c) missing event backend implementation

libbsd doesn't help with (a) and (c) because it doesn't come with a working getprogname(), kevent nor epoll implementation for Windows. It would only help to solve issue (b). Except that issue (b) is trivial enough to solve in the context of libdispatch itself.

The general problem that I see with adding 3rd party dependencies to Swift in the context of Windows is that:

1) Windows doesn't come with a package manager, making installation of 3rd party libs less convenient and more time consuming compared to Linux.

2) A 3rd party dependency may have a compiling, working and tested implementation for POSIX but not necessarily for Windows, forcing someone to sit down and make things work on Windows.

3) Windows users are not used to the idea that they would have to use a package manager to satisfy dependencies for an application before they will be able to run the app. They are used to downloading an installer that takes care of everything.

4) Every 3rd party dependency that Swift requires increases the minimum download size of an app that was built with Swift. But the goal should be to keep this minimum runtime size as small as possible.

So I would vastly prefer that libdispatch would just provide its own implementation of the necessary getprogname() and TAILQ macros instead of dragging libbsd in just for that.

dplanitzer:
@MadCoder updated patch

src/shims/getprogname.c
```diff
@@ -0,0 +1,72 @@
...
+	PVOID *lpContext)
+{
+	TCHAR path[MAX_PATH];
+	DWORD length = GetModuleFileName(NULL, path, sizeof(path));
+	
+	if (length < 0) {
+		progname[0] = '\0';
+		return TRUE;
+	} else {
+		const char * filename;
```
MadCoder: remove the space after the star

src/shims/getprogname.c
```diff
@@ -0,0 +1,72 @@
...
+		strcpy_s(progname, sizeof(progname), filename);
+		return TRUE;
+	}
+}
+
+const char *
+getprogname(void)
+{
+	(void) InitOnceExecuteOnce(&getprogname_init_once,
+				getprogname_init_once_handler,
```
MadCoder: these should be 2 indents from the base line, and remove the space after the cast.

src/shims/getprogname.h
```diff
@@ -30,6 +30,10 @@
 extern const char *__progname;
 #endif /* __ANDROID */
 
+#if defined(_WIN32)
+const char* getprogname(void);
```
MadCoder: `const char *progname` (mind the spaces)

dplanitzer:
cc @MadCoder fixed spacing

MadCoder:
@swift-ci please test

PR #368, opened 2018-06-19
added a sys/queue.h implementation for Windows

dplanitzer:
Added a copy of the Apple / Darwin sys/queue.h and hooked things up such that we use that one when building on Windows because the Windows SDK doesn't come with a sys/queue.h header.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -61,6 +61,7 @@ if(UNIX)
 elseif(WIN32)
   target_sources(dispatch
                  PRIVATE
+                   shims/generic_sys_queue.h
                    shims/generic_win_stubs.c
                    shims/generic_win_stubs.h)
 endif()
--- a/src/internal.h
+++ b/src/internal.h
@@ -242,11 +242,11 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
 #endif
 
 #include <sys/stat.h>
-#include <sys/queue.h>
 
 #if defined(_WIN32)
 #include <time.h>
 #else
+#include <sys/queue.h>
 #include <sys/mount.h>
 #ifdef __ANDROID__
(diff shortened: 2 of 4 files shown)


dplanitzer:
cc @MadCoder @compnerd

compnerd:
Why can't we use libbsd?  I've been building against libbsd for Windows.

adierking:
I've been working on adding Windows support to libbsd: https://gitlab.freedesktop.org/libbsd/libbsd/merge_requests/1

tkremenek:
What precisely is the dependency here on libbsd?  Is it just to get a header, or a link dependency?  The license issue just needs to be reviewed (I can consult Apple's lawyers), but not necessarily a blocker.

dplanitzer:
See my comment [here](https://github.com/apple/swift-corelibs-libdispatch/pull/367l). I can create an independent implementation of needed queue macros instead if that makes things simpler.

tkremenek:
@dplanitzer I get a 404 with that link

dplanitzer:
@MadCoder @tkremenek I've updated the PR with an independent implementation of the relevant TAILQ macros. Libdispatch only uses a handful of them anyway. I've tested the implementation locally and verified that it works beautifully :)

MadCoder:
@swift-ci please test

tkremenek:
@dplanitzer was the relevant implementation of `TAILQ` cribbed from the libbsd header?  If so, technically that's copyrighted and can't just be copy-and-pasted into another place.

dplanitzer:
@tkremenek I did a new implementation from scratch based on the documentation in the man 3 queue page. It's a simple standard doubly-linked list implementation while the sys/queue.h thing does something more fancy to avoid a few if-then checks on list manipulation.

tkremenek:
@dplanitzer OK, perfect.

tkremenek:
I think we can merge this.

PR #369, opened 2018-06-20
Report crash messages on Windows to the CRT

adierking:
Provide a Windows implementation of the `_dispatch_set_crash_log_*` macros which
pass the crash message to the CRT using _RPTF0. If libdispatch is linked against
a debug version of msvcrt, this will display a dialog with detailed information
about the crash and provide an option to attach a debugger. On release versions
of msvcrt, this reporting is a no-op.

Example of what this dialog looks like (from an actual issue I encountered):

Screenshot

### DIFF
--- a/src/internal.h
+++ b/src/internal.h
@@ -293,6 +293,7 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
 #endif
 #if defined(_WIN32)
 #include <io.h>
+#include <crtdbg.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


adierking:
cc @MadCoder @compnerd

compnerd:
I like this.  It definitely would improve the developer experience.  I am in favor of this, but will let @parkera or @MadCoder make the final call on this.

MadCoder:
this is exactly how it's meant to be used, excellent!

MadCoder:
@swift-ci please test

PR #371, opened 2018-06-22
BlocksRuntime: match the ABI specification for LLP64

compnerd:
The Blocks ABI v1 defined the `size` and `reserved` fields as `unsigned
long int` rather than `uintptr_t`.  This breaks on LLP64 hosts where
`uintptr_t` is larger than `unsigned long int`.  Make this conform to
the specification.


compnerd:
CC: @MadCoder

compnerd:
@swift-ci please test

PR #372, opened 2018-06-22
Windows doesn't support symlinks so use copy instead (tests/CMakeLists.txt)

dplanitzer:
Same fix that we did for CMakeLists.txt but this time for the cmake file in tests.


dplanitzer:
cc @MadCoder @compnerd

MadCoder:
@swift-ci please test

PR #373, opened 2018-06-22
Use snprintf instead of _snprintf on Windows

adierking:
_snprintf does not guarantee null termination if the size argument is too small.
Starting with VS2015, the UCRT provides a C99-compliant snprintf. Use it, and
also fix a related issue where _dispatch_client_assert_fail was passing a
too-small size.

### DIFF
--- a/src/internal.h
+++ b/src/internal.h
@@ -928,7 +928,7 @@ _dispatch_ktrace_impl(uint32_t code, uint64_t a, uint64_t b,
 		dispatch_assert(_length != -1); \
 		_msg = (char *)malloc((unsigned)_length + 1); \
 		dispatch_assert(_msg); \
-		_snprintf(_msg, (unsigned)_length, "%s" fmt, DISPATCH_ASSERTION_FAILED_MESSAGE, ##__VA_ARGS__); \
+		snprintf(_msg, (unsigned)_length + 1, "%s" fmt, DISPATCH_ASSERTION_FAILED_MESSAGE, ##__VA_ARGS__); \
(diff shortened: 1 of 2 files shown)


adierking:
cc @MadCoder @compnerd

src/internal.h
```diff
@@ -928,7 +928,7 @@ _dispatch_ktrace_impl(uint32_t code, uint64_t a, uint64_t b,
 		dispatch_assert(_length != -1); \
 		_msg = (char *)malloc((unsigned)_length + 1); \
 		dispatch_assert(_msg); \
-		_snprintf(_msg, (unsigned)_length, "%s" fmt, DISPATCH_ASSERTION_FAILED_MESSAGE, ##__VA_ARGS__); \
+		snprintf(_msg, (unsigned)_length + 1, "%s" fmt, DISPATCH_ASSERTION_FAILED_MESSAGE, ##__VA_ARGS__); \
```
compnerd: Bleh; too bad that we still have the `_scprintf` here.  Can we use `snprintf` with a 0 size instead to get the desired length?  Then we could just use this version globally rather than having a windows only path.  The difference between the `asprintf` version and this would be the extra calls, but it is on the abort path anyways, so I don't think that it matters too much.

adierking: Yeah, I agree this is kinda nasty - is it fine if I make that change as a follow-up?

asprintf is also used in several places in test code :(

MadCoder:
@swift-ci please test

adierking:
ping @MadCoder

PR #374, opened 2018-06-22
Stub out unimplemented Windows functions

adierking:
Provide stubs for unimplemented Windows functions so that dispatch.dll can link
and we can start building the tests. If one of these stubs is called, it will be
reported to the CRT.

### DIFF
--- a/src/shims/generic_win_stubs.c
+++ b/src/shims/generic_win_stubs.c
@@ -1,3 +1,23 @@
+#include "internal.h"
+
+/*
+ * This file contains stubbed out functions we are using during
+ * the initial Windows port.  When the port is complete, this file
+ * should be empty (and thus removed).
+ */
+
+void
+_dispatch_runloop_queue_dispose(dispatch_queue_t dq DISPATCH_UNUSED,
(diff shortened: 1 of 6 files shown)


adierking:
cc @MadCoder @compnerd

src/shims/generic_win_stubs.c
```diff
@@ -1,3 +1,94 @@
...
+void
+_dispatch_event_loop_wake_owner(dispatch_sync_context_t dsc DISPATCH_UNUSED,
+		dispatch_wlh_t wlh DISPATCH_UNUSED, uint64_t old_state DISPATCH_UNUSED,
+		uint64_t new_state DISPATCH_UNUSED)
+{
+	WIN_PORT_ERROR();
+}
+
+void
+_dispatch_event_loop_wait_for_ownership(
```
MadCoder: for this family of functions, you should instead start a: src/event/event_windows.c and copy the functions from event_epoll.c that do nothing and or assert.

they will likely not have an implementation on windows for a while in the first place.

adierking:
Updated to put event functions in an event_windows.c.

PR #376, opened 2018-06-26
Added MSVC definitions for the DISPATCH_* macros in dispatch/base.h

dplanitzer:
Added MSVC specific definitions of the DISPATCH_* macros to dispatch/base.h. This is one necessary step to make it possible to compile the Swift APIs for ibdispatch. E.g. we need a working definition of DISPATCH_NORETURN to ensure that we can successfully compile his function:

`public func dispatchMain() -> Never {
	CDispatch.dispatch_main()
}`

because its return type is `Never`.

DISPATCH_* macros which are empty are empty because MSVC has no equivalent to the corresponding clang / gcc attribute.

### DIFF
--- a/dispatch/base.h
+++ b/dispatch/base.h
@@ -65,6 +65,29 @@
 #define DISPATCH_ALWAYS_INLINE __attribute__((__always_inline__))
 #define DISPATCH_UNAVAILABLE __attribute__((__unavailable__))
 #define DISPATCH_UNAVAILABLE_MSG(msg) __attribute__((__unavailable__(msg)))
+#elif defined(_MSC_VER)
+#define DISPATCH_NORETURN __declspec(noreturn)
+#define DISPATCH_NOTHROW __declspec(nothrow)
+#define DISPATCH_NONNULL1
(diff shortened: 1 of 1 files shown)


MadCoder:
@swift-ci please test

compnerd:
This is generally not going to end very well.  MSVC doesn't support the swift calling convention nor does it support Blocks.  How do you intend to deal with those?

MadCoder:
I think this is more about allowing people to use libdispatch on windows with MSVC, which is IMO sth we should do.

building libdispatch with MSVC makes no sense, but we should support using it in a project.

MadCoder:
(as in maybe it's not why @dplanitzer was doing it, but we should still do it anyway for the reason I just gave)

dplanitzer:
MadCoder stated one reason.

The other reason is that we need proper definitions for macros like DISPATCH_NORETURN for building libdispatch for Swift with the Swift compiler / clang-cl.

Previously compiling the Swift overlay (specifically the Swift function dispatchMain) would fail because the prototype for the `dispatch_main` function was missing the noreturn annotation.

PR #377, opened 2018-06-26
Ignore Windows FLS destructors during process exit

adierking:
There's a quirk on Windows where exiting the process will still run fiber-local
storage destructors for the thread that called ExitProcess() (but only for that
thread). pthreads-based platforms don't call any key destructors on exit, and
_libdispatch_tsd_cleanup() assumes this. So, on Windows, calling exit() from a
queue thread will trigger the fatal error in _dispatch_frame_cleanup().

Implement a reliable method of detecting whether the process is exiting, and if
it is, ignore calls to _libdispatch_tsd_cleanup().

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -961,9 +961,57 @@ _dispatch_install_thread_detach_callback(dispatch_function_t cb)
 }
 #endif
 
+#if defined(_WIN32)
+static bool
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

PR #379, opened 2018-06-27
Fixed some Swift overlay compilation problems on Windows

dplanitzer:
This fixes some of the problems that prevent us from successfully compiling the Swift overlay on Windows. I've separated this out from a larger fix because this should be pretty much the uncontroversial part. Anyway, we simply treat Windows the same as Linux and Android which makes this part of the code compile.

There are additional (and potentially controversial) fixes needed to get the whole overlay to compile on Windows. Please see the Swift forum topic [here](https://forums.swift.org/t/libdispatch-api-how-to-support-llp64-platforms/14012/4) for more info.

Specifically, @MadCoder : what would be the preferred solution to make LLP64 and LP64 platforms happy? E.g. one proposed solution is to replace uses of `long`, etc with `size_t`. E.g. change

`dispatch_semaphore_t  dispatch_semaphore_create(long value)`

to

`dispatch_semaphore_t  dispatch_semaphore_create(ssize_t value)`

Also please see [this PR](https://github.com/apple/swift/pull/17533) over on the compiler side which fixes another problem with the Swift overlay.

### DIFF
--- a/src/swift/Source.swift
+++ b/src/swift/Source.swift
@@ -113,7 +113,7 @@ public extension DispatchSource {
 	}
 #endif
 
-#if !os(Linux) && !os(Android)
+#if !os(Linux) && !os(Android) && !os(Windows)
 	public struct ProcessEvent : OptionSet, RawRepresentable {
 		public let rawValue: UInt
 		public init(rawValue: UInt) { self.rawValue = rawValue }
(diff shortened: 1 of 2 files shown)


dplanitzer:
cc @MadCoder @compnerd

compnerd:
@dplanitzer seems that there are some conflicts, would you mind rebasing?

compnerd:
(Closing since it is out of date, #413 has an updated patch)

PR #380, opened 2018-06-29
[Do not merge] Made the semaphore API integrate better with LLP64 (Windows) systems

dplanitzer:
This PR is an attempt to solve the problem [discussed here](https://forums.swift.org/t/libdispatch-api-how-to-support-llp64-platforms/14012).

The short summary is that the libdispatch Swift API uses `Int` types in its API which are 64bit wide on 64bit Windows. But the underlying libdispatch C API uses `long` types which are 32bit wide on 64bit Windows. Consequently we get type conflicts when we try to build the Swift API for libdispatch.

An important goal here is to come up with a solution that ensures that we do not end up forcing users of the Swift API to sprinkle `#if os(Windows)` checks throughout their code to support both non-Windows and Windows platforms.

What does the PR do?

It replaces uses of `long` with `dispatch_long_t` and uses of `unsigned long` with `dispatch_ulong_t`. The code compiles as is on Windows and Linux and it has been verified on Linux that the unit tests for semaphores still work.

The PR is limited to semaphores because the main question at this point is whether the approach as such would be acceptable or whether a different approach would be preferred by the libdispatch maintainers.

### DIFF
--- a/dispatch/base.h
+++ b/dispatch/base.h
@@ -288,6 +288,34 @@
 #define DISPATCH_TRANSPARENT_UNION
 #endif
 
+#if defined(_WIN64)
+// LLP64
+typedef long long dispatch_long_t;
+typedef unsigned long long dispatch_ulong_t;
+
+#define DISPATCH_LONG_MIN LLONG_MIN
+#define DISPATCH_LONG MAX LLONG_MAX
+
+#define DISPATCH_ULONG_MIN ULLONG_MIN
+#define DISPATCH_ULONG_MAX ULLONG_MAX
+
+#define PRIDISPATCHLONG "%lld"
+#define PRIDISPATCHULONG "%llu"
+#else
+// LP64 and ILP32
+typedef long dispatch_long_t;
+typedef unsigned long dispatch_ulong_t;
+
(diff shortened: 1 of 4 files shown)


dplanitzer:
cc @MadCoder @troughton

MadCoder:
I don't like using size_t which is not the right semantic. when using long we really mean natural register/pointer-sized. I'd rather have a dispatch_long / dispatch_ulong typedef instead.

internally we'll likely need DISPATCH_{,U}LONG_MAX and similar.

dplanitzer:
I've updated the PR with your feedback. Please let me know whether that's what you had in mind.

If this approach is acceptable then I could either:

a) sunset this PR and create a new PR which applies this kind of change to all of libdispatch in one go

or

b) we merge this PR and I create additional and separate ones for queue, source and groups

I guess that (b) might make reviewing easier. Let me know what you'd prefer.

compnerd:
I'm going to close this PR for now.  I believe that the issue here has been resolved by changing most of the `long` to `intptr_t` which gives the desired semantics on all the platforms.  Additionally, this commit no longer cleanly applies.

PR #381, opened 2018-07-05
Drop @_downgrade_exhaustivity_check attribute

CodaFi:
This hack was only needed for Swift 3 mode in a narrow case.  Now that
we're dropping Swift 3 from master, the source break this was working
around is no longer relevant.


CodaFi:
@swift-ci test

CodaFi:
⛵️

PR #382, opened 2018-07-05
build: support embedding in other CMake projects

adierking:
Allow libdispatch to be embedded in other CMake projects using
add_subdirectory(). This essentially means that `CMAKE_SOURCE_DIR` and
`CMAKE_BINARY_DIR` cannot be used directly. Use the versions of these variables
which are scoped to the current project or directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 
 cmake_minimum_required(VERSION 3.4.3)
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
 
 project(dispatch
         VERSION 1.3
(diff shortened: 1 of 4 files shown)


adierking:
cc @MadCoder @compnerd

compnerd:
This really is functionally equivalent.  No real strong feelings on this.  Up to @MadCoder or @millenomi

tkremenek:
@swift-ci test

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 
 cmake_minimum_required(VERSION 3.4.3)
 
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake/modules")
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules")
```
ktopley-apple: Why is this CMAKE_CURRENT_SOURCE_DIR and not PROJECT_SOURCE_DIR like everywhere else?

adierking: The PROJECT_* variables are set by the project() command, which is called below this, so they aren't available yet. I assume that the module path is intentionally set before calling project() because it could allow for defining custom platform/toolchain modules that get picked up by the command.

PR #383, opened 2018-07-11
build: link against librt if it's available

adierking:
CMakeLists.txt checks for librt's presence but the result is never used. This
causes a link failure on systems where clock_gettime() is provided by librt.


adierking:
cc @MadCoder @compnerd

PR #384, opened 2018-07-12
Remove generic_unix_stubs

adierking:
generic_unix_stubs.c seems to no longer be necessary. generic_unix_stubs.h still
had some required definitions in it, so move those into shims.h and a new
shims/mach.h. This also lets us remove some duplicated definitions in the
Windows shims.


adierking:
cc @MadCoder @compnerd

compnerd:
Seems reasonable

PR #385, opened 2018-07-16
Change the signature for DispatchData.enumerateBytes() to match that …

ktopley-apple:
…of Data.enumerateBytes() for Swift 4.2 and obsolete the old version.

rdar://problem/40775762

### DIFF
--- a/src/swift/Data.swift
+++ b/src/swift/Data.swift
@@ -117,8 +117,22 @@ public struct DispatchData : RandomAccessCollection {
 		return try body(contentPtr)
 	}
 
+	@available(swift 4.2)
+	public func enumerateBytes(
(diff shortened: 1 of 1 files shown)


ktopley-apple:
@swift-ci: please smoke test

parkera:
I'm not sure you really want to emulate this particular method; we need to get rid of it in favor of using the `UnsafeRaw` family.

ktopley-apple:
> we need to get rid of it in favor of using the UnsafeRaw family.
I agree that we should do something better. This is a short term fix.

parkera:
Ok

ktopley-apple:
@swift-ci: please test

ktopley-apple:
@swift-ci please test

PR #389, opened 2018-07-20
build: link against swiftOnoneSupport in debug configuration

compnerd:
When building debug, swift requires a supplementary library.  The swift driver
is aware of this and will make the appropriate arrangements.  However, since the
swift component is compacted into the C library, the linker driver being used
does not understand this requirement.  Explicitly link against the support
library.  This allows building the debug configuration of libdispatch with a
release configuration of the swift standard library.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -84,7 +84,7 @@ if(HAVE_OBJC)
 endif()
 if(ENABLE_SWIFT)
   set(swift_optimization_flags)
-  if(CMAKE_BUILD_TYPE MATCHES Release)
+  if(NOT CMAKE_BUILD_TYPE MATCHES Debug)
     set(swift_optimization_flags -O)
   endif()
   add_swift_library(swiftDispatch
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @MadCoder

compnerd:
@swift-ci please test

compnerd:
@MadCoder ping

aschwaighofer:
@compnerd This commit caused failures on the LSAN bot. https://ci.swift.org/job/oss-swift-incremental-RA-lsan-linux-ubuntu-16_10/4863

The failures look like false positives to me: Memory leaks in SIL instructions should have been caught earlier during building the Swift dylib and overlays.

aschwaighofer:
@compnerd Actually, I was wrong. We have known leaks in the sanitizer. This change just caused libdispatch to be build optimized instead of in a debug configuration and now we see those leaks.
We should probably use the standard library build type instead of the swift tools build type for building libdispatch.

aschwaighofer:
That should have read: "We have known leaks in the optimizer."

aschwaighofer:
https://github.com/apple/swift/pull/18898

PR #391, opened 2018-08-23
Remove version number used in 'available' attribute for non-specific platform '*'.

dingobye:
This patch aims to make swift-corelibs-libdispatch adapted to [this PR](https://github.com/apple/swift/pull/18895), which warns for version number in 'available' attribute for non-specific platform '*'.

### DIFF
--- a/src/swift/Queue.swift
+++ b/src/swift/Queue.swift
@@ -51,19 +51,19 @@ public extension DispatchQueue {
 
 	public enum GlobalQueuePriority {
 		@available(macOS, deprecated: 10.10, message: "Use qos attributes instead")
-		@available(*, deprecated: 8.0, message: "Use qos attributes instead")
+		@available(iOS, deprecated: 8.0, message: "Use qos attributes instead")
 		case high
 
 		@available(macOS, deprecated: 10.10, message: "Use qos attributes instead")
-		@available(*, deprecated: 8.0, message: "Use qos attributes instead")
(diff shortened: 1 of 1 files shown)


jrose-apple:
Do we care about watchOS and tvOS too? tvOS version numbers match iOS's, but watchOS's don't.

ktopley-apple:
I agree with Jordan - the tvOS and watchOS deprecations should also be included.

jrose-apple:
I don't think I or Ding knows for sure what those should be! It's also a little funny that these are *just* in corelibs-dispatch, which is never used on iOS/tvOS/watchOS. Or did we miss some in the overlays too?

dingobye:
Thank you for the feedbacks, Jordan and Kim!

Can we just remove those lines of `@available(*, deprecated: 8.0, ...)` here? Just take a look at [the relevant part in master](https://github.com/apple/swift/blob/master/stdlib/public/SDK/Dispatch/Queue.swift#L51). There are attributes for four platforms macOS/iOS/tvOS/watchOS, among which I think keeping only macOS is sufficient for _swift-corelibs-libdispatch_ project.

jrose-apple:
I suspect that SCD is trying to match the contents of the Darwin overlay exactly for most of these files, and that they should therefore be kept in sync.

dingobye:
Alright, let's keep it in sync. This is a conservative solution anyway.

jrose-apple:
@swift-ci Please test Linux

PR #392, opened 2018-09-28
Non-Darwin SDKs product results in per-arch directories.

drodriguez:
Depends on apple/swift#19432

Unix (other than Darwin) and Windows do not support fat binaries.
However, the build system was placing the build results in the
platform/OS folder, instead of the per-architecture folder. Having two
architectures side by side was impossible.

After applying the above patch in the Swift compiler, non-Darwin
platforms will look into the per-architecture folders, so the sibling
projects need to leave the products in the right place.


compnerd:
@drodriguez can we stage the installation so that we put a copy as a workaround in the old location as well so that we can merge this without the rest of the changes?  Having fewer dependent changes makes things easier.

drodriguez:
I suppose that it might work. We can only put one architecture in the old location, though, but Jenkins only builds one architecture, so it should work.

PR #393, opened 2018-10-01
Build both shared and static libraries when ENABLE_SWIFT is ON

kevints:
Fixes: https://bugs.swift.org/browse/SR-7085
       https://bugs.swift.org/browse/SR-7039
       https://bugs.swift.org/browse/SR-8854

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,6 +72,7 @@ if(ENABLE_SWIFT)
                            ${SWIFT_RUNTIME_LIBDIR}/${CMAKE_SHARED_LIBRARY_PREFIX}swiftSwiftOnoneSupport${CMAKE_SHARED_LIBRARY_SUFFIX})
 
   set(INSTALL_TARGET_DIR "${INSTALL_LIBDIR}/swift/${SWIFT_OS}" CACHE PATH "Path where the libraries will be installed")
+  set(INSTALL_STATIC_TARGET_DIR "${INSTALL_LIBDIR}/swift_static/${SWIFT_OS}" CACHE PATH "Path where the static libraries will be installed")
   set(INSTALL_DISPATCH_HEADERS_DIR "${INSTALL_LIBDIR}/swift/dispatch" CACHE PATH "Path where the headers will be installed for libdispatch")
   set(INSTALL_BLOCK_HEADERS_DIR "${INSTALL_LIBDIR}/swift/Block" CACHE PATH "Path where the headers will be installed for the blocks runtime")
   set(INSTALL_OS_HEADERS_DIR "${INSTALL_LIBDIR}/swift/os" CACHE PATH "Path where the os/ headers will be installed")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -2,7 +2,74 @@
 include(SwiftSupport)
 include(DTrace)
 
-add_library(dispatch
+if(ENABLE_SWIFT)
+    set(dispatch_targets dispatch dispatch_static dispatch_objects)
+else()
+   set(dispatch_targets dispatch)
+endif()
+
+function(dispatch_add_library)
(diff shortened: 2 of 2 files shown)


weissi:
this is great! CC'ing @tkremenek and @das as the Swift overall and Dispatch release managers for 4.2. I'd like to get this into 4.2.1 because it's clearly a regression from 4.1 that should be fixed ASAP.

ktopley-apple:
@swift-ci please test

ktopley-apple:
So then we should wait for  apple/swift#19640 <https://github.com/apple/swift/pull/19640> to merge before proceeding with this change, correct?

CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ if(ENABLE_SWIFT)
                            ${SWIFT_RUNTIME_LIBDIR}/${CMAKE_SHARED_LIBRARY_PREFIX}swiftSwiftOnoneSupport${CMAKE_SHARED_LIBRARY_SUFFIX})
 
   set(INSTALL_TARGET_DIR "${INSTALL_LIBDIR}/swift/${SWIFT_OS}" CACHE PATH "Path where the libraries will be installed")
+  set(INSTALL_STATIC_TARGET_DIR "${INSTALL_LIBDIR}/swift_static/${SWIFT_OS}" CACHE PATH "Path where the static libraries will be installed")
```
compnerd: To be honest, I'm not sure I understand the point of these two cached variables.  They are part of the resource dir naming scheme for the compiler, so why not always compose them when used?

src/CMakeLists.txt
```diff
@@ -2,7 +2,74 @@
 include(SwiftSupport)
 include(DTrace)
 
-add_library(dispatch
+if(ENABLE_SWIFT)
+    set(dispatch_targets dispatch dispatch_static dispatch_objects)
+else()
+   set(dispatch_targets dispatch)
+endif()
```
compnerd: 2-space indents please

src/CMakeLists.txt
```diff
@@ -2,7 +2,74 @@
...
+    # Swift builds both shared and static libraries
+    if(ENABLE_SWIFT)
+        add_library(dispatch_objects OBJECT ${ARGV})
+        set_property(TARGET dispatch_objects PROPERTY POSITION_INDEPENDENT_CODE ON)
+        add_library(dispatch SHARED $<TARGET_OBJECTS:dispatch_objects>)
+        set_target_properties(dispatch PROPERTIES
+            LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/swift/${SWIFT_OS}
+            ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib/swift/${SWIFT_OS}
+        )
+        add_library(dispatch_static STATIC $<TARGET_OBJECTS:dispatch_objects>)
```
compnerd: This breaks for windows.  Please do a combined build for Windows.  We need to build the objects twice, using the same objects will not work.

CMakeLists.txt
kevints: I am happy to drop them

src/CMakeLists.txt
compnerd: No, object libraries are fine, just as long as they are built for the final type (e.g. built with knowledge that they will be used for a shared library or not).  The best way to accomplish this is to perform two builds with the right CMake flags so that you build the objects the right way.

compnerd:
https://github.com/apple/swift-corelibs-libdispatch/pull/395 actually will unconditionally build the shared and static versions of BlocksRuntime now.  That should be sufficient for your needs right?

kevints:
Closing in favor of a new approach where swift's build-script creates 2 libdispatch build directories, one for shared and one for static after talking this approach through with @compnerd and @gottesmm offline.

PR #395, opened 2018-10-02
build: perform parallel builds of BlocksRuntime

compnerd:
Build both the static and shared variants of the BlocksRuntime.  This is needed
for proper shared linkage to the BlocksRuntime (e.g. when building SourceKit).
Build both variants and install to the appropriate location.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -74,14 +74,12 @@ if(ENABLE_SWIFT)
 
   set(INSTALL_TARGET_DIR "${INSTALL_LIBDIR}/swift/${SWIFT_OS}" CACHE PATH "Path where the libraries will be installed")
   set(INSTALL_DISPATCH_HEADERS_DIR "${INSTALL_LIBDIR}/swift/dispatch" CACHE PATH "Path where the headers will be installed for libdispatch")
-  set(INSTALL_BLOCK_HEADERS_DIR "${INSTALL_LIBDIR}/swift/Block" CACHE PATH "Path where the headers will be installed for the blocks runtime")
   set(INSTALL_OS_HEADERS_DIR "${INSTALL_LIBDIR}/swift/os" CACHE PATH "Path where the os/ headers will be installed")
 endif()
 
 if(NOT ENABLE_SWIFT)
   set(INSTALL_TARGET_DIR "${INSTALL_LIBDIR}" CACHE PATH "Path where the libraries will be installed")
   set(INSTALL_DISPATCH_HEADERS_DIR "include/dispatch" CACHE PATH "Path where the headers will be installed")
-  set(INSTALL_BLOCK_HEADERS_DIR "include" CACHE PATH "Path where the headers will be installed for the blocks runtime")
   set(INSTALL_OS_HEADERS_DIR "include/os" CACHE PATH "Path where the headers will be installed")
 endif()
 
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @gottesmm @kevints

CMakeLists.txt
```diff
@@ -132,34 +130,82 @@ endif()
...
-                            INTERFACE_LINK_LIBRARIES ${CMAKE_DL_LIBS})
-  endif()
-
-  add_library(BlocksRuntime::BlocksRuntime ALIAS BlocksRuntime)
+                          INTERFACE_INCLUDE_DIRECTORIES
+                            ${PROJECT_SOURCE_DIR}/src/BlocksRuntime
+                          IMPORTED_LOCATION
+                            ${binary_dir}/${CMAKE_SHARED_LIBRARY_PREFIX}BlocksRuntime${CMAKE_SHARED_LIBRARY_SUFFIX}
+                          IMPORTED_IMPLIB
+                            ${binary_dir}/${CMAKE_IMPORT_LIBRARY_PREFIX}BlocksRuntime${CMAKE_IMPORT_LIBRARY_SUFFIX})
```
kevints: What does this do on platforms that don't use import libraries?

CMakeLists.txt
```diff
@@ -132,34 +130,82 @@ endif()
...
+                        -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
+                        -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+                        -DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}
+                        -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}
+                        -DBUILD_SHARED_LIBS=YES
+                      BUILD_BYPRODUCTS
+                        <BINARY_DIR>/${CMAKE_SHARED_LIBRARY_PREFIX}BlocksRuntime${CMAKE_SHARED_LIBRARY_SUFFIX}
+                        <BINARY_DIR>/${CMAKE_IMPORT_LIBRARY_PREFIX}BlocksRuntime${CMAKE_IMPORT_LIBRARY_SUFFIX}
+                      BUILD_ALWAYS
+                        1)
```
kevints: My experience using `ExternalProject` hasn't been great compared to creating library targets that are directly addressable in the build system and I strongly suggest we don't use it in dispatch.

1. We need to propagate all possible interesting CMake variables from the outer build system to the inner ExternalProject build system via `CMAKE_ARGS`. For example here you don't propagate `CMAKE_C_FLAGS`, `CMAKE_AR` or `CMAKE_LIBRARY_PATH` which could cause the downstream libraries to be built differently. There are too many influential variables to enumerate here.
1. ExternalProjects don't get good dependency tracking compared to targets that cmake understands natively. You have to declare BUILD_BYPRODUCTS manually and it's easy to under or over-depend. ExternalProjects also don't have great IDE integration and they appear opaque.
1. Setting up a library target as `IMPORTED` prevents dependency tracking and prevents it being exported from the build system (via `install(EXPORT)` for example).
1. Ninja is not designed to be invoked recursively - you can get into a situation where the downstream ninja is performing tasks equal to the number of CPUs, so you can have (N^2-1) tasks running in parallel (where `N` is the number of cores your build machine has). Console output is also broken.

Having said all that, have you considered making `BlocksRuntime` and `BlocksRuntime_static` targets using `add_library` instead?

src/BlocksRuntime/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+
+set(CMAKE_C_STANDARD 11)
+set(CMAKE_C_STANDARD_REQUIRED YES)
+
+set(CMAKE_C_VISIBILITY_PRESET hidden)
+set(CMAKE_C_VISIBILITY_INLINES_HIDDEN YES)
+
+check_include_files("objc/objc-internal.h" HAVE_OBJC)
+
+option(INSTALL_PRIVATE_HEADERS "install private headers in the same location as the public ones" NO)
```
kevints: By placing this inside an ExternalProject there's no way to trigger this option anymore.

compnerd: Oh, but there is!  There is the still the option in libdispatch, which just passes it through :-)

CMakeLists.txt
compnerd: It does nothing meaningful on those platforms.  The install rules are what matter, which need to be guarded.

CMakeLists.txt
compnerd: Well, we already are using ExternalProject for libdispatch in swift, and it has been working quite well :-).  That said, there is interest in moving to the LLVM wrapper.  Doing so will automatically propagate those other variables implicitly.

The dependency tracking issue that you portray is not entirely correct.  There is tracking of the dependencies in the sense that the external project's byproducts are linked into the imported library target.  So if the sub-library changes, CMake/ninja realize that and rebuild the dependent targets.

Yes, I did consider that.  I suppose that if we want to just duplicate the rules, thats certainly an option.  I really am strongly opposed to creating the custom wrappers for both variants.  I suppose that we could always extract the BlocksRuntime rules into a separate CMakeList and include it in this one to make this easier to work with.

compnerd:
Discussed with @kevints offline.  We have a better approach that we are going to be taking and this change will be revised with a new version.

PR #396, opened 2018-10-03
build: honour `BUILD_SHARED_LIBS`

compnerd:
This is needed for proper shared linkage to the BlocksRuntime (e.g. when
building SourceKit).  We now build BlocksRuntime shared or static as requested.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,6 +87,9 @@ endif()
 
 option(ENABLE_DTRACE "enable dtrace support" "")
 
+# NOTE(abdulras) this is the CMake supported way to control whether we generate
+# shared or static libraries.  This impacts the behaviour of `add_library` in
+# what type of library it generates.
 option(BUILD_SHARED_LIBS "build shared libraries" ON)
 
 option(ENABLE_TESTING "build libdispatch tests" ON)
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

CMakeLists.txt
```diff
@@ -74,14 +74,12 @@ if(ENABLE_SWIFT)
 
   set(INSTALL_TARGET_DIR "${INSTALL_LIBDIR}/swift/${SWIFT_OS}" CACHE PATH "Path where the libraries will be installed")
   set(INSTALL_DISPATCH_HEADERS_DIR "${INSTALL_LIBDIR}/swift/dispatch" CACHE PATH "Path where the headers will be installed for libdispatch")
-  set(INSTALL_BLOCK_HEADERS_DIR "${INSTALL_LIBDIR}/swift/Block" CACHE PATH "Path where the headers will be installed for the blocks runtime")
```
kevints: These directories should be `swift` or `swift_static` depending on the value of `BUILD_SHARED_LIBS`

compnerd: Sure, though this issue has existed for some time now and is not made any worse by this change.  I'll put up a separate PR to address that.

CMakeLists.txt
```diff
@@ -132,35 +130,43 @@ endif()
...
+              ${PROJECT_SOURCE_DIR}/src/BlocksRuntime/data.c
+              ${PROJECT_SOURCE_DIR}/src/BlocksRuntime/runtime.c)
   set_target_properties(BlocksRuntime
                         PROPERTIES
                           POSITION_INDEPENDENT_CODE TRUE)
   if(HAVE_OBJC AND CMAKE_DL_LIBS)
-    set_target_properties(BlocksRuntime
-                          PROPERTIES
-                            INTERFACE_LINK_LIBRARIES ${CMAKE_DL_LIBS})
+    target_link_libraries(BlocksRuntime
```
kevints: This is a functional linkage change that should go into another PR (though I suspect this branch is never executed). If this is required then we need to change our approach to the module map as well (as we'd now need to link dl and objc).

CMakeLists.txt
```diff
@@ -132,35 +130,43 @@ endif()
...
+    set(LIBDISPATCH_BLOCKSRUNTIME_HEADER_DIR ${SWIFT_LIBDIR}/swift/Block)
+  else()
+    set(LIBDISPATCH_BLOCKSRUNTIME_HEADER_DIR ${CMAKE_INSTALL_INCLUDEDIR})
+  endif()
+
   install(FILES
             ${PROJECT_SOURCE_DIR}/src/BlocksRuntime/Block.h
           DESTINATION
-            "${INSTALL_BLOCK_HEADERS_DIR}")
+            ${LIBDISPATCH_BLOCKSRUNTIME_HEADER_DIR})
```
kevints: Header dir vs headers dir, let's keep this consistent. Also do we need to drop this cache var in this PR? Also I thought project-prefixed UPPER_SNAKE_CASED variables were conventionally used for cache vars rather than locals?

compnerd: Yeah, lowercase variables are preferred.

CMakeLists.txt
compnerd: `INTERFACE` doesn't work with shared libraries, so this should be public (since `INTERFACE` is equal to `PUBLIC` with shared).

compnerd:
apple/swift#19674

@swift-ci please test

compnerd:
apple/swift#19674

@swift-ci please test

drodriguez:
@compnerd: I might be wrong, but it seems that `libBlocksRuntime.so` ends up in the “root” build directory, not anywhere its sibling `libdispatch.so`. If I understand correctly, any Swift code that imports `Dispatch` will need `libBlocksRuntime.so` available (from the changes in the `modulemap`).

One of those is XCTest. While compiling it uses https://github.com/apple/swift-corelibs-xctest/blob/master/CMakeLists.txt#L48 to point to the `libdispatch.so` library, but it points the search path to `…/src` (where no `libBlocksRuntime.so` can be found).

Another is LLBuild. From https://github.com/apple/swift-llbuild/blob/master/products/llbuildSwift/CMakeLists.txt#L19-L24 it points the library search path to `…/src/.libs`.

I think those might need to be fixed, but I will wait until I can talk with you to confirm.

PR #397, opened 2018-10-04
build: support static builds for Swift

compnerd:
Adjust the paths for static runtime builds of Swift.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,38 +44,32 @@ dispatch_common_warnings()
 option(ENABLE_DISPATCH_INIT_CONSTRUCTOR "enable libdispatch_init as a constructor" ON)
 set(USE_LIBDISPATCH_INIT_CONSTRUCTOR ${ENABLE_DISPATCH_INIT_CONSTRUCTOR})
 
+# NOTE(abdulras) this is the CMake supported way to control whether we generate
+# shared or static libraries.  This impacts the behaviour of `add_library` in
+# what type of library it generates.
+option(BUILD_SHARED_LIBS "build shared libraries" ON)
+
 option(ENABLE_SWIFT "enable libdispatch swift overlay" OFF)
 if(ENABLE_SWIFT)
   if(NOT CMAKE_SWIFT_COMPILER)
     message(FATAL_ERROR "CMAKE_SWIFT_COMPILER must be defined to enable swift")
   endif()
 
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @kevints

kevints:
@compnerd looking at the build the test programs are failing due to undefined symbols in ICU. This is due to the once-private link dependencies of swiftCore being needed (as there's no portable way for unix archives to contain library dependencies). The dispatch buildsystem currently creates an IMPORTED target for the swiftCore library without regard to its dependencies. This happens to work in the shared case but doesn't in the static case, but to make it work robustly is equivalent to duplicating the `-static-stdlib` feature of `swiftc` in the dispatch buildsystem, which is the feature we're trying to fix. A more robust approach would be to use swiftc (with the -static-stdlib flag when BUILD_SHARED_LIBS is NO) to link these test programs. We could also use an exported swiftCore target from the swift build system but I suspect that's not currently tractable.

gottesmm:
@kevints We support creating exported targets from the parent swift cmake.

compnerd:
@swift-ci please test

compnerd:
https://github.com/apple/swift/pull/19834

@swift-ci please test

CMakeLists.txt
```diff
@@ -44,38 +44,32 @@ dispatch_common_warnings()
 option(ENABLE_DISPATCH_INIT_CONSTRUCTOR "enable libdispatch_init as a constructor" ON)
 set(USE_LIBDISPATCH_INIT_CONSTRUCTOR ${ENABLE_DISPATCH_INIT_CONSTRUCTOR})
 
+# NOTE(abdulras) this is the CMake supported way to control whether we generate
```
kevints: Consider linking straight to the reference: https://cmake.org/cmake/help/v3.4/variable/BUILD_SHARED_LIBS.html

compnerd: I do like that better; I think Im going to do that in a follow up (to get this merged first).  This was just moving the option around.

compnerd:
Also note that this fixes the builds for me with tests after @DougGregor's change (layout differences between Darwin and Linux bites us again).  My follow up change will actually use the swift compiler to link the Swift SDK overlay and split it out, so that actually makes all this significantly more robust and easier to work with.

gottesmm:
This LGTM from the swift side of things. I think the cmake will work fine as well given that we are essentially going through cmake config and performing an identity transformation (since swift's config provides that same information). @MadCoder hows this look to you?

compnerd:
@MadCoder - ping?

PR #398, opened 2018-10-09
[CMake] Make sure to link swiftrt.o when building the Swift Dispatch overlay.

DougGregor:
Because we are linking libdispatch through clang rather than swift, we need to
explicitly add swiftrt.o, which is needed by ELF and COFF to register metadata
sections. Fixes rdar://problem/44941707.


DougGregor:
@swift-ci please test

DougGregor:
@swift-ci please test

PR #399, opened 2018-10-11
swift: cast to the appropriate type

compnerd:
This fixes an error from the swift compiler:
	error: cannot convert value of type 'UInt32' to expected argument type 'UInt'

PR #400, opened 2018-10-11
queue: clean up a covered switch

compnerd:
clang objects to the `default` case in a covered switch:

	src/queue.c:1127:2: error: default label in switch which covers all enumeration values [-Werror,-Wcovered-switch-default]

Remove the label, relying on the covered switch to catch the issue.


compnerd:
Please test with the following PRs:
apple/swift-corelibs-libdispatch#399

@swift-ci please test

PR #403, opened 2018-10-15
Fix the signature of _dispatch_install_thread_detach_callback()

adierking:
This function is declared as accepting a `dispatch_function_t` callback, which
is a function pointer with a `void *` argument. However, the implementation and
Swift overlay declare the callback without arguments, causing a conflict which
Clang warns about. Change the function signature to accept the correct type.


compnerd:
CC: @MadCoder

compnerd:
@swift-ci please test

compnerd:
This seems trivially correct.  Any objections to merging this @MadCoder?

PR #404, opened 2018-10-16
io: use posix_fadvise() instead of readahead()

adierking:
readahead() is a Linux-specific system call, and many C libraries (e.g.
Android's Bionic) do not declare it by default. A more-portable alternative is
to use posix_fadvise() with `POSIX_FADV_WILLNEED`. This is functionally
equivalent to readahead() on Linux.

Additionally, on FreeBSD, Dispatch is currently using fcntl() with `F_RDAHEAD`.
This is not exactly equivalent to Darwin's `F_RDADVISE`. Using posix_fadvise()
should work better here too.


compnerd:
@swift-ci please test

compnerd:
I think that this is the right approach to make this more portable.  Any opinions @MadCoder or @parkera?

ktopley-apple:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@MadCoder  - okay to merge?

compnerd:
@ktopley-apple - okay to merge this as well?

PR #406, opened 2018-10-16
dispatch_c99: remove unnecessary __printflike() definition

adierking:
Some C libraries (e.g. Bionic on Android) provide a `__printflike()` macro but
this test always redefines it on Linux-based systems. bsdtests.h already has
better logic to ensure it is available.


adierking:
cc @MadCoder @compnerd

compnerd:
@swift-ci please test

PR #407, opened 2018-10-17
tests: read from a temporary file instead of vi on non-Apple platforms

adierking:
Some tests hardcode `/usr/bin/vi` as a path to a file to read from. This is
definitely not going to exist on Android and Windows. Instead of hardcoding this
path in each test, create a `dispatch_test_get_test_file()` function to retrieve
the path. On Linux and Windows, return the path to the test executable, because
this should be significantly more reliable.

### DIFF
--- a/tests/dispatch_io_net.c
+++ b/tests/dispatch_io_net.c
@@ -58,16 +58,14 @@ int
 main(int argc, char** argv)
 {
 	struct hostent *he;
-	int sockfd, clientfd;
+	int sockfd = -1, clientfd = -1;
+	int read_fd = -1, fd = -1;
 	struct sockaddr_in addr1, addr2, server;
 	socklen_t addr2len;
 	socklen_t addr1len;
 	pid_t clientid;
 
-	const char *path = "/usr/bin/vi";
-	int read_fd, fd;
-
-	if (argc == 2) {
+	if (argc == 3) {
 		// Client
 		dispatch_test_start(NULL);
 
@@ -94,11 +92,19 @@ main(int argc, char** argv)
 
 		// Read from the socket and compare the contents are what we expect
 
+		const char *path = argv[2];
 		fd = open(path, O_RDONLY);
 		if (fd == -1) {
 			test_errno("client-open", errno, 0);
 			test_stop();
 		}
+
+		// The reference file path given to us by the server was produced by
+		// dispatch_test_get_large_file(). It may point to a temporary file, and
+		// we are responsible for cleaning it up because we are the last to
+		// access it.
+		dispatch_test_release_large_file(path);
+
 #ifdef F_NOCACHE
 		if (fcntl(fd, F_NOCACHE, 1)) {
 			test_errno("client-fcntl F_NOCACHE", errno, 0);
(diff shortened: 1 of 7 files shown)


tests/dispatch_test.c
```diff
@@ -82,6 +84,35 @@ dispatch_test_check_evfilt_read_for_fd(int fd)
 #endif
 }
 
+const char *
+dispatch_test_get_test_file(void)
+{
+#if defined(_WIN32)
+	static char path[MAX_PATH];
```
compnerd: I think that you want `MAX_PATH + 1` to account for the trailing null-terminator.

tests/dispatch_test.c
```diff
@@ -82,6 +84,35 @@ dispatch_test_check_evfilt_read_for_fd(int fd)
...
+	DWORD len = GetModuleFileNameA(NULL, path, sizeof(path));
+	if (len == 0 || len == sizeof(path)) {
+		fprintf(stderr, "GetModuleFileNameA() failed: %u\n", GetLastError());
+		exit(EXIT_FAILURE);
+	}
+	return path;
```
compnerd: Uh, this is returning a buffer to a stack allocated buffer.  This is a classic use-after-free.

tests/dispatch_test.c
```diff
@@ -82,6 +84,35 @@ dispatch_test_check_evfilt_read_for_fd(int fd)
...
+	if (len == -1) {
+		perror("readlink");
+		exit(EXIT_FAILURE);
+	}
+	if (len == (ssize_t)sizeof(path)) {
+		fprintf(stderr, "readlink: buffer too small\n");
+		exit(EXIT_FAILURE);
+	}
+	path[len] = '\0';
+	return path;
```
compnerd: Same.

MadCoder:
the test binary will not be large enough for this test to be valuable.

tests/dispatch_test.c
adierking: `MAX_PATH` by definition includes room for a null terminator.

tests/dispatch_test.c
adierking: The buffer is static.

tests/dispatch_test.c
adierking: The buffer is static.

adierking:
>the test binary will not be large enough for this test to be valuable.

That's fair. I still want to avoid hardcoding a path though because my experience is that this is extremely unreliable on Android. Does it sound reasonable to write out a ~2MB temporary file and then use that in the tests?

compnerd:
@swift-ci please test

tests/dispatch_io_net.c
```diff
@@ -94,11 +92,13 @@ main(int argc, char** argv)
 
 		// Read from the socket and compare the contents are what we expect
 
+		const char *path = argv[2];
 		fd = open(path, O_RDONLY);
 		if (fd == -1) {
 			test_errno("client-open", errno, 0);
 			test_stop();
 		}
+		dispatch_test_release_large_file(path);
```
ktopley-apple: It looks odd that you release a file here without having created it. I had to read the rest of the source file to find out why this is correct. Some comments here about how this file is created would be useful.

compnerd:
@swift-ci please test

ktopley-apple:
Looks good. I will merge when the build is complete.

PR #408, opened 2018-10-17
tests: fall back to fork() if posix_spawnp() is not available

adierking:
Android only started supporting posix_spawnp() in API level 28 (Android 9), so
it is not going to be available on most devices. If the system doesn't provide
it, then fall back to using fork() and execve().


adierking:
cc @MadCoder @compnerd

ktopley-apple:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@MadCoder  - okay to merge?

adierking:
I updated this to also handle the posix_spawnp() call in bsdtestharness and to error-check execve(). Please re-test.

cc @MadCoder @compnerd @ktopley-apple

compnerd:
@swift-ci please test

tests/bsdtestharness.c
```diff
@@ -104,9 +104,12 @@ main(int argc, char *argv[])
 	struct timeval tv_start;
 	gettimeofday(&tv_start, NULL);
 
+#ifdef HAVE_POSIX_SPAWNP
+#ifdef __APPLE__
```
ktopley-apple: It's not clear why are using this flag - why not just always use posix_spawn() to both fork and exec, just as you do in the other change (in dispatch_io_net.c)?

adierking: I assume you're referring to the `__APPLE__` check here - it's used to handle the Apple-only codepath where `POSIX_SPAWN_SETEXEC` can be set. The entire call isn't Apple-only. (Previously the code above defined `POSIX_SPAWN_SETEXEC` to 0 on non-Apple platforms, but I think that's confusing.)

ktopley-apple: My question really is, why is there an Apple-only code path. Doesn't the non-Apple code path work for all platforms that have posix_spawn()? And if not, why is the code in dispatch_io_net.c different?

adierking: The Apple-only codepath is used to start the process in a suspended state and set up monitoring before resuming it. The flags used to do this (`POSIX_SPAWN_SETEXEC` and `POSIX_SPAWN_START_SUSPENDED`) are Apple extensions. This codepath isn't necessary in dispatch_io_net because the server doesn't need to monitor the client process, and the non-Apple codepath below (`#if SIMPLE_TEST_HARNESS`) does much simpler process monitoring.

adierking:
Rebased

ktopley-apple:
@swift-ci please test

ktopley-apple:
Thanks for the explanation. I'll merge this when the build has completed.

PR #409, opened 2018-10-21
dispatch: make public interfaces LLP64 friendly

compnerd:
This attempts to replace the public uses of `unsigned long` with
`uintptr_t` and `long` with `intptr_t`.  The use of `long` and `unsigned
long` here was as a type suitable for matching pointer width.  However,
on LLP64 targets, this does not hold.  Replace them with `intptr_t`
which gives a proper pointer sized type.  This is done only for the
argument types and not for named types to ensure that the ABI is not
modified for C++ (where a template may be specialized on the named
type).

### DIFF
--- a/dispatch/block.h
+++ b/dispatch/block.h
@@ -322,7 +322,7 @@ dispatch_block_perform(dispatch_block_flags_t flags,
  */
 API_AVAILABLE(macos(10.10), ios(8.0))
 DISPATCH_EXPORT DISPATCH_NONNULL1 DISPATCH_NOTHROW
-long
+intptr_t
 dispatch_block_wait(dispatch_block_t block, dispatch_time_t timeout);
 
 /*!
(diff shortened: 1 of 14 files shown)


compnerd:
CC: @parkera @das @MadCoder

compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple

compnerd:
@swift-ci please test

compnerd:
Okay, so, this *almost* works for everything.  The problem becomes `dispatch_block_flags_t` which is a named type and a typed enumeration.  The current enumeration is `typedef`ed to an `unsigned long` which is not machine word sized on LLP64.  Changing the underlying type to `uintptr_t` would resolve the issue.  However, doing so would change the underlying type, which is a problem for C++ code.  This would matter specifically where `dispatch_block_flags_t` is used for a C++ function definition or used in a template specialization, where the change to the underlying type would result in a decoration change and thus break ABI.  This obviously would not change the interfaces for libdispatch itself as it is a pure C library.  I would love some hints on how to best address that particular type (Is it acceptable to break users in such a way?  Or do we want to change the underlying type only on certain targets?)

PR #410, opened 2018-10-23
build: remove "/usr/include" for swift

compnerd:
Remove the explicit /usr/include search path for the swift library build.  This
allows us to build for Windows as well.


compnerd:
@swift-ci please test

PR #413, opened 2018-10-29
SDK: guard BSD specific paths against windows

compnerd:
Treat windows like Linux and android and remove the BSD interfaces on
these targets.  This allows us to mostly build the SDK overlay for
Windows.


compnerd:
@swift-ci please test

compnerd:
CC: @MadCoder @ktopley-apple

PR #414, opened 2018-10-29
WIP/DNM: LLP64 support for SDK Overlay

compnerd:
If we cannot break the ABI for C++ users of libdispatch, this would
allow us to build the SDK overlay for LLP64 targets.

### DIFF
--- a/src/swift/Block.swift
+++ b/src/swift/Block.swift
@@ -40,14 +40,24 @@ public class DispatchWorkItem {
 	internal var _block: _DispatchBlock
 
 	public init(qos: DispatchQoS = .unspecified, flags: DispatchWorkItemFlags = [], block: @escaping @convention(block) () -> ()) {
-		_block =  dispatch_block_create_with_qos_class(dispatch_block_flags_t(UInt(flags.rawValue)),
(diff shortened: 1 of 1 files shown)


ktopley-apple:
@compnerd - is this an alternative to PR #409?

compnerd:
@ktopley-apple, no, not an alternative, it is a supplement to.  The problem is that I did not know whether changing the underlying type for `dispatch_flags_t` is a reasonable option or not.  If it is not possible to change the underlying type of the typedef, then this will be needed (which IMO is unfortunate).  It also means that using libdispatch on an LLP64 host is difficult.  One other option may be to change the ABI for LLP64 hosts, in which case, this patch would not be needed.  This is related to the comment https://github.com/apple/swift-corelibs-libdispatch/pull/409#issuecomment-433219177

ktopley-apple:
@compnerd #409 was just merged. My reading of your reply is that this change should no longer be needed. If that's not the case, please let me know and I'll merge it.

compnerd:
@ktopley-apple this change is still needed (as the flag sizes are different across Windows x64 and the other targets).

PR #415, opened 2018-11-02
DispatchStubs: make more Windows friendly

compnerd:
Provide an alias for the import symbol as the ObjC runtime is normally
in a separate DLL and the compiler will annotate the function as being
DLLImport.  This allows us to resolve the symbols when building the SDK
overlay.


gparker42:
Should `objc_retainAutoreleasedReturnValue` be defined as `__declspec(dllexport)`? I suspect that calls to it are emitted inside code other than libdispatch itself.

compnerd:
Yeah, but I expect that to be handled by `DISPATCH_RUNTIME_STDLIB_INTERFACE` no?  Those are interfaces that should be exposed right?

PR #421, opened 2018-11-05
build: honour the USE_*_LINKER flags for swift

compnerd:
Ensure that we honour the linker setting when building the swift SDK
overlay.  This is particularly important for android.


compnerd:
CC: @ktopley-apple @MadCoder

compnerd:
@swift-ci please test

PR #422, opened 2018-11-05
build: add an explicit link to BlocksRuntime

compnerd:
The SDK overlay requires the BlocksRuntime but did not explicitly link
to it.  This works on most platforms, however, Windows requires that all
symbols are fully resolved (effectively `-z defs` on Linux).  Add the
missing dependency.


compnerd:
CC: @MadCoder @ktopley-apple

compnerd:
@swift-ci please test

PR #427, opened 2018-11-15
enable cross-linking/synchronise with Foundation

compnerd:
Synchronise some of the CMake support logic with Foundation/XCTest and enable cross-linking.


compnerd:
CC: @ktopley-apple @MadCoder

compnerd:
@swift-ci please test

PR #429, opened 2018-11-17
build improvements

compnerd:
This improves the build for Windows, making the installed image usable for building on Windows target from a distributed toolchain.  It also cleans up some of the build for static library construction and re-synchronises with XCTest (resource bundling was not supported, but this is mostly for synchrony and not because it is used).


compnerd:
CC: @MadCoder @ktopley-apple

compnerd:
@swift-ci please test

compnerd:
CC: @spevans - this should have all the changes to properly build and install the static and shared versions of libdispatch.  Would you mind taking on the changes to build-script-impl to get that resolved?

PR #430, opened 2018-11-24
dispatch: include `time.h` on windows

compnerd:
time.h is needed on Windows to ensure that we have a definition of
`timespec`.  This is needed when building the swift SDK overlay.  It
seems that when cross-compiling, we were getting lucky with the
`timespec` definition being provided by some other header being
included.  This repairs the build of libdispatch's swift SDK overlay on
Windows targeting Windows.

### DIFF
--- a/dispatch/dispatch.h
+++ b/dispatch/dispatch.h
@@ -42,6 +42,9 @@
 #include <unistd.h>
 #endif
 #include <fcntl.h>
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @das @MadCoder @ktopley-apple

compnerd:
@swift-ci please test

PR #432, opened 2018-11-27
libdispatch: clean up some Win64 warnings

compnerd:
This cleans up the few warnings that remain when building libdispatch.
It now builds clean of warnings with clang 8.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -2568,11 +2568,11 @@ static size_t
 _dispatch_io_debug_attr(dispatch_io_t channel, char* buf, size_t bufsiz)
 {
 	dispatch_queue_t target = channel->do_targetq;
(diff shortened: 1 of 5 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple @MadCoder

src/io.c
```diff
@@ -19,6 +19,7 @@
  */
 
 #include "internal.h"
+#include <inttypes.h>
```
ktopley-apple: This should be added to internal.h instead.

src/semaphore.c
```diff
@@ -19,6 +19,7 @@
  */
 
 #include "internal.h"
+#include <inttypes.h>
```
ktopley-apple: This should be added to internal.h instead.

compnerd:
@swift-ci please test

PR #433, opened 2018-11-27
build: make the static build work well

compnerd:
This is needed to make the shared+static build for build-script work.  The
primary issue is the dependency on the module-maps doesn't work due to the lack
of add_custom_command, so the target dependency cannot be hooked up.
Additionally, undo the hack for the AST_NAME as we need that to be proper for
the copy to succeed.  Finally, make sure that the we set the location for the
archive output correctly.


compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple @MadCoder @kevints

PR #434, opened 2018-12-01
build: build libdispatch as position independent

compnerd:
When building libdispatch as a static library ensure that it is built as
a PIC library.  This is implicit when building a shared library.  If the
library is not built shared, we cannot statically link it on Unix
targets.

This partially resolves SR-9384!


compnerd:
CC: @gottesmm @spevans @ktopley-apple @MadCoder

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #438, opened 2019-01-18
build: remove unnecessary `find_package`

compnerd:
Now that we link using the swift driver, we no longer need to explicitly
add in the path to the swift registrar.  Remove the unncessary
`find_package` to make it easier to build libdispatch.


compnerd:
@swift-ci please test

PR #440, opened 2019-01-30
tests: port the test harness to Windows

adierking:
This provides a functional Windows port of bsdtestharness and its dependencies
so that we can start running the test suite on Windows. Some tests already build
and run correctly, while others either fail at runtime or don't compile. I'll
be following up with fixes for individual tests later.


compnerd:
@swift-ci please test

ktopley-apple:
@swift-ci please test

PR #441, opened 2019-02-02
dispatch: fix logging to a file on Windows

adierking:
Port the dispatch_debug test to Windows and fix a bug that it found related to
logfile initialization.

_dispatch_logv_init()'s Windows implementation uses _fdopen() and fprintf() to
write the log header. However, this gives ownership of the file to stdio, so
fclose() closes the descriptor and causes future writes to trigger CRT assertion
failures. Avoid this by using snprintf() and _write() instead.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -733,15 +733,19 @@ _dispatch_logv_init(void *context DISPATCH_UNUSED)
 			dispatch_log_basetime = _dispatch_absolute_time();
 #endif
 #if defined(_WIN32)
-			FILE *pLogFile = _fdopen(dispatch_logfile, "w");
-
 			char szProgramName[MAX_PATH + 1] = {0};
 			GetModuleFileNameA(NULL, szProgramName, MAX_PATH);
 
-			fprintf(pLogFile, "=== log file opened for %s[%lu] at "
-					"%ld.%06u ===\n", szProgramName, GetCurrentProcessId(),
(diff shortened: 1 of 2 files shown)


adierking:
cc @compnerd @MadCoder @ktopley-apple

ktopley-apple:
@swift-ci please test

src/init.c
```diff
@@ -733,15 +733,17 @@ _dispatch_logv_init(void *context DISPATCH_UNUSED)
...
-			fprintf(pLogFile, "=== log file opened for %s[%lu] at "
-					"%ld.%06u ===\n", szProgramName, GetCurrentProcessId(),
-					tv.tv_sec, (int)tv.tv_usec);
-			fclose(pLogFile);
+			char szMessage[512];
+			int len = snprintf(szMessage, sizeof(szMessage),
+					"=== log file opened for %s[%lu] at %ld.%06u ===\n",
+					szProgramName, GetCurrentProcessId(), tv.tv_sec,
+					(int)tv.tv_usec);
+			if (len > 0 && (size_t)len < sizeof(szMessage)) {
```
ktopley-apple: If the second check fails, wouldn't it be better to write the first sizeof(szMessage) bytes to the log rather than discard everything?

ktopley-apple:
@swift-ci please test

PR #442, opened 2019-02-04
Retry sem_wait on signal interrupts

gmilos:


ktopley-apple:
@swift-ci please test

PR #445, opened 2019-02-12
Swift: explicitly cast `UInt32` to `dispatch_flags_t`

compnerd:
Because `dispatch_flags_t` is a sized enumeration on Windows, the type
is not implicitly constructed (as in C/C++).  Add the explicit
constructor call, which should be portable to all the targets.

### DIFF
--- a/src/swift/Block.swift
+++ b/src/swift/Block.swift
@@ -40,7 +40,7 @@ public class DispatchWorkItem {
 	internal var _block: _DispatchBlock
 
 	public init(qos: DispatchQoS = .unspecified, flags: DispatchWorkItemFlags = [], block: @escaping @convention(block) () -> ()) {
-		let flags: dispatch_block_flags_t = numericCast(flags.rawValue)
+		let flags: dispatch_block_flags_t = dispatch_block_flags_t(numericCast(flags.rawValue))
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@ktopley-apple - this fixes the Windows build and should work across all the targets

jrose-apple:
I'm concerned about what type is being picked for the `numericCast` here. This might cause problems when using a flag whose high bit is set.

ktopley-apple:
> I'm concerned about what type is being picked for the `numericCast` here. This might cause problems when using a flag whose high bit is set.

dispatch_block_flags_t only has 6 defined bits, all at the low-order end, so in reality this should not be a concern. We have the practical problem that the raw value for this type is not the same on all platforms (and not even on all versions of the same platform) and there seems to be no other way to write one line of code that will convince the compiler to do the "right thing".

ktopley-apple:
The build failed. I saw this same failure when I last tried to fix this in a similar way. I filed a Radar on the compiler for that. It looks like the only way around this for now is to put back the Win32 conditional and use the previous version of the code for the Windows build.

compnerd:
Closing this since it doesn't work in favour of #446

PR #456, opened 2019-03-14
CMake: Use the Swift Driver and Response Files

gmittert:
Rather than rolling our own dependencies tracking, use the Swift driver
to handle dependencies and dispatching jobs. Additionally use response
files in the compilation to prevent extremely long command lines.

As requested to be copied from https://github.com/apple/swift-corelibs-foundation/pull/1959

### DIFF
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -1,5 +1,37 @@
 
 include(CMakeParseArguments)
+include(ProcessorCount)
+
+# Creates an output file map give a target and its list of sources at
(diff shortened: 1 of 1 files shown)


jmittert:
@compnerd

gmittert:
Rebased and now pass in `-j` to the swift driver.

gmittert:
@swift-ci please test

compnerd:
I think that we should actually just switch to the builtin support for Swift instead of this.  That will enable a lot of additional benefits.  Thanks for doing this though!

PR #457, opened 2019-03-19
Fixed conflicting method signature for _dispatch_install_thread_detach_callback()

triplef:
This fixes various "conflicting types for XXX" errors I get when building the project for Android after the recent libdispatch-1121 merge. The changes are not Android-specific though and probably result in warnings or errors on other platforms (depending on -Werror).

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -7303,7 +7303,7 @@ gettid(void)
 static void (*_dispatch_thread_detach_callback)(void);
 
 void
-_dispatch_install_thread_detach_callback(dispatch_function_t cb)
+_dispatch_install_thread_detach_callback(void (*cb)(void))
 {
 	if (os_atomic_xchg(&_dispatch_thread_detach_callback, cb, relaxed)) {
 		DISPATCH_CLIENT_CRASH(0, "Installing a thread detach callback twice");


private/queue_private.h
```diff
@@ -493,7 +493,7 @@ dispatch_async_enforce_qos_class_f(dispatch_queue_t queue,
  * "detached" before the thread exits or the application will crash.
  */
 DISPATCH_EXPORT
-void _dispatch_install_thread_detach_callback(void (*cb)(void));
+void _dispatch_install_thread_detach_callback(dispatch_function_t cb);
```
compnerd: No, this should remain as is.  The parameter is never passed to the callback.  @adierking fixed this.

src/queue.c
```diff
@@ -571,7 +571,7 @@ dispatch_block_testcancel(dispatch_block_t db)
 	return (bool)(dbpd->dbpd_atomic_flags & DBF_CANCELED);
 }
 
-long
+intptr_t
```
compnerd: This is already fixed in #453

src/source.c
```diff
@@ -197,7 +197,7 @@ dispatch_source_get_extended_data(dispatch_source_t ds,
 }
 
 void
-dispatch_source_merge_data(dispatch_source_t ds, unsigned long val)
+dispatch_source_merge_data(dispatch_source_t ds, uintptr_t val)
```
compnerd: This seems Linux specifically.

private/queue_private.h
triplef: I’m not sure I follow, but the [current implementation](https://github.com/apple/swift-corelibs-libdispatch/blob/13129d9/src/queue.c#L7306) of this method uses the signature proposed in this patch. Should I change the implementation signature instead?

src/source.c
triplef: This change was based on the [header declaration](https://github.com/apple/swift-corelibs-libdispatch/blob/13129d945443c8b6e8ceaa8f7ce922fc9751eebf/dispatch/source.h#L658), but the same change was again already fixed in #453.

private/queue_private.h
adierking: I changed this from `dispatch_function_t` in #403. It seems that the change to queue.c may have been reverted on accident. Let's fix it there.

triplef:
Thanks, I’ve updated the patch to only include the change for `_dispatch_install_thread_detach_callback()` in queue.c to align with queue_private.h.

compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple

PR #458, opened 2019-03-19
Added missing queue macros for building on Android

triplef:
TAILQ_CONCAT() and LIST_FOREACH_SAFE() are not defined in [queue.h](https://android.googlesource.com/platform/development/+/73a5a3b/ndk/platforms/android-20/include/sys/queue.h) in the Android NDK.

I simply added the macros to internal.h, but please let me know if you’d prefer another place to add them.

### DIFF
--- a/src/internal.h
+++ b/src/internal.h
@@ -269,12 +269,12 @@ upcast(dispatch_object_t dou)
 #if defined(_WIN32)
 #include <time.h>
 #else
-#include <sys/queue.h>
 #include <sys/mount.h>
 #ifdef __ANDROID__
 #include <linux/sysctl.h>
 #else
 #include <sys/sysctl.h>
+#include <sys/queue.h>
 #endif /* __ANDROID__ */
 #include <sys/socket.h>
 #include <sys/time.h>
--- a/src/shims.h
+++ b/src/shims.h
@@ -31,8 +31,11 @@
 #include <pthread.h>
 #else // defined(_WIN32)
 #include "shims/generic_win_stubs.h"
-#include "shims/generic_sys_queue.h"
 #endif // defined(_WIN32)
+
+#if defined(_WIN32) || defined(__ANDROID__)
+#include "shims/generic_sys_queue.h"
+#endif
 
 #ifdef __ANDROID__
 #include "shims/android_stubs.h"
--- a/src/shims/generic_sys_queue.h
+++ b/src/shims/generic_sys_queue.h
@@ -110,12 +110,23 @@
 		struct type *le_prev; \
 	}
 
+#define	LIST_EMPTY(head) ((head)->lh_first == NULL)
+
 #define LIST_FIRST(head) ((head)->lh_first)
 
 #define LIST_FOREACH(var, head, field) \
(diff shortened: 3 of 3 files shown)


compnerd:
This is insufficient, there are additional ones, #453 covers this as well.

src/internal.h
```diff
@@ -432,6 +432,24 @@ DISPATCH_EXPORT DISPATCH_NOTHROW void dispatch_atfork_child(void);
...
 
+#ifndef TAILQ_CONCAT
+#define	TAILQ_CONCAT(head1, head2, field) do {				\
+	if (!TAILQ_EMPTY(head2)) {					\
+		*(head1)->tqh_last = (head2)->tqh_first;		\
+		(head2)->tqh_first->field.tqe_prev = (head1)->tqh_last;	\
+		(head1)->tqh_last = (head2)->tqh_last;			\
+		TAILQ_INIT((head2));					\
+	}								\
+} while (0)
```
compnerd: This macro is entirely incorrect and wont even build.  The fields are named differently.

triplef: This is taken straight from the macOS SDK and looks identical to me to [the implementation](https://github.com/apple/swift-corelibs-libdispatch/pull/453/files#diff-2731630646bf7c7e29667b3b6cfc03e2R95) from #453.

triplef:
The thing is that (probably unlike Windows) the Android NDK implements *some* of these macros, but not all of them. I see a couple of alternate ways that I think should work and would probably be better than just sticking them in internal.h like I did here:

a) Add missing macros to android_stubs.h.
b) Wait for #453, then include `generic_sys_queue.h` on Android and *don’t* include `sys/queue.h`.
c) Wait for #453, then include `generic_sys_queue.h` and add #ifndef’s around each definition.

Do any of these approaches sound good or is there another way you’d prefer?

compnerd:
I'm totally happy with b.  It seems much nicer.  The file is already annoying to work with, so c isn't very appealing.  a would really increase the duplication, so I really think b is the best approach.

triplef:
Sounds good! I’ll wait for #453 and re-do this.

compnerd:
@triplef - #453 is merged :-)

triplef:
Thanks! :) I’ve updated the patch as discussed.

src/shims.h
```diff
@@ -31,9 +31,12 @@
 #include <pthread.h>
 #else // defined(_WIN32)
 #include "shims/generic_win_stubs.h"
-#include "shims/generic_sys_queue.h"
 #endif // defined(_WIN32)
 
+#if defined(_WIN32) || defined(__ANDROID__)
```
compnerd: I'm tempted to add Linux to this clause as well ...

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

src/shims.h
triplef: Makes sense, let me know if you want me add it to this PR.

compnerd: A separate PR is probably better.  Lets get android working again!

compnerd:
CC: @ktopley-apple

PR #459, opened 2019-03-19
Use sched_yield() as fallback for preemption yield

triplef:
pthread_yield() is not available on Android.

In combination with #458 this fixes all build issues using Android NDK targeting API level 21 and up.

### DIFF
--- a/src/shims/yield.h
+++ b/src/shims/yield.h
@@ -156,9 +156,9 @@ void *_dispatch_wait_for_enqueuer(void **ptr);
 #elif defined(_WIN32)
 #define _dispatch_preemption_yield(n) { (void)n; Sleep(0); }
 #define _dispatch_preemption_yield_to(th, n) { (void)n; Sleep(0); }
-#else 
-#define _dispatch_preemption_yield(n) { (void)n; pthread_yield(); }
-#define _dispatch_preemption_yield_to(th, n) { (void)n; pthread_yield(); }
+#else
+#define _dispatch_preemption_yield(n) { (void)n; sched_yield(); }
+#define _dispatch_preemption_yield_to(th, n) { (void)n; sched_yield(); }
 #endif // HAVE_MACH
 
 #pragma mark -


triplef:
I see. I’ll wait for #453 to be merged and re-do the this to also use sleep on Android.

triplef:
I’ve updated the patch to use `sleep(0)` if `pthread_yield_np()` is not available.

src/shims/yield.h
```diff
@@ -156,9 +156,9 @@ void *_dispatch_wait_for_enqueuer(void **ptr);
 #elif defined(_WIN32)
 #define _dispatch_preemption_yield(n) { (void)n; Sleep(0); }
 #define _dispatch_preemption_yield_to(th, n) { (void)n; Sleep(0); }
-#else 
-#define _dispatch_preemption_yield(n) { (void)n; pthread_yield(); }
-#define _dispatch_preemption_yield_to(th, n) { (void)n; pthread_yield(); }
+#else
+#define _dispatch_preemption_yield(n) { (void)n; sleep(0); }
+#define _dispatch_preemption_yield_to(th, n) { (void)n; sleep(0); }
```
compnerd: This impacts the Linux and FreeBSD paths which, unlike android, have a reasonable libc, and a working pthread implementation.  I think that you should add an `__ANDROID__` clause here and use `sleep`, as the other environments have `pthread_yield`.

adierking:
Using `sleep(0)` to yield on non-Windows systems isn't correct because there's no guarantee that it will actually do anything (and in fact some versions of glibc will return immediately). The Windows path uses it because Microsoft explicitly documents that `Sleep(0)` relinquishes the thread's timeslice. The correct fallback should be `sched_yield()`, which is mandated by POSIX.

src/shims/yield.h
triplef: Got it, thanks. I’ve updated the patch.

triplef:
@adierking that’s what I thought and had used originally. :-) Just to confirm, would you use `sched_yield()` just on Android or everywhere as a fallback?

adierking:
> @adierking that’s what I thought and had used originally. :-) Just to confirm, would you use `sched_yield()` just on Android or everywhere as a fallback?

I think it should be used as a fallback everywhere. Nothing should be using `pthread_yield()` anymore because `sched_yield()` is a standardized function which accomplishes the same thing.

triplef:
Great, I updated the patch to use `sched_yield()`.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

ktopley-apple:
@swift-ci please test

compnerd:
@swift-ci please test

PR #460, opened 2019-03-19
Added missing private headers to cmake install.

triplef:
These are included by private.h and should be installed alongside.


compnerd:
@swift-ci please test

PR #461, opened 2019-03-19
Added underscores to gettid(void) to remove name clash

tachoknight:
with system-provided gettid() in /usr/include/bits/unistd_ext.h that is not static. 

According to https://raw.githubusercontent.com/bminor/glibc/master/ChangeLog, Florian Weimer moved `units_ext.h` from `sysdeps/generic/bits` to `bits` and libdispatch won't compile due to the function having different declarations between the system-provided one and the one defined in `queue.c`. 

I figure that, since it's only being called in one spot in the same file, it was safe to rename the function in without having to make a special #ifdef Linux-only version.

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -7300,21 +7300,21 @@ libdispatch_init(void)
 #ifdef SYS_gettid
 DISPATCH_ALWAYS_INLINE
 static inline pid_t
-__gettid(void)
+_gettid(void)
 {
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple

ktopley-apple:
@swift-ci please test

PR #464, opened 2019-03-20
CMake: Correctly detect MSVC

MarkIngramUK:
CMake: Correctly detect MSVC, or other compilers in MSVC compatibility modes (e.g. clang-cl, or clang -fms-compatibility).

~~According to CMake [documentation](https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_SIMULATE_ID.html), when CMake detects a compiler is simulated, it will set the `CMAKE_<LANG>_SIMULATE_ID` variable. However it appears this isn't happening, and both MSVC and clang-cl fail to identify as MSVC during project generation.~~

`CMAKE_<LANG>_SIMULATE_ID` != `MSVC` when generating for Visual Studio. The CMake variable [MSVC](https://cmake.org/cmake/help/latest/variable/MSVC.html) is set when a compiler is simulating MSVC (by defining `_MSC_VER`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ project(dispatch
         LANGUAGES C CXX)
 enable_testing()
 
-if("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")
+if(MSVC)
   include(ClangClCompileRules)
 endif()
 
--- a/cmake/modules/DispatchCompilerWarnings.cmake
+++ b/cmake/modules/DispatchCompilerWarnings.cmake
@@ -1,5 +1,5 @@
 
-if("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")
+if(MSVC)
   # TODO: someone needs to provide the msvc equivalent warning flags
   macro(dispatch_common_warnings)
   endmacro()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -179,7 +179,7 @@ if(WIN32)
                              PRIVATE
                                _CRT_NONSTDC_NO_WARNINGS)
 endif()
-if("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")
+if(MSVC)
   target_compile_options(dispatch PRIVATE /EHs-c-)
 else()
   target_compile_options(dispatch PRIVATE -fno-exceptions)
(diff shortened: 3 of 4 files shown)


src/BlocksRuntime/BlocksRuntime.def
```diff
@@ -1,4 +1,4 @@
 LIBRARY BlocksRuntime
 EXPORTS
-  _NSConcreteGlobalBlock CONSTANT
-  _NSConcreteStackBlock CONSTANT
+  _NSConcreteGlobalBlock DATA
+  _NSConcreteStackBlock DATA
```
compnerd: Please don't change this.  It was using `CONSTANT` for a very specific reason - the generated import library is different.

compnerd:
BTW, `clang-cl` absolutely does identify correctly, sounds like someone is wrong with your setup?

src/BlocksRuntime/BlocksRuntime.def
MarkIngramUK: Apologies, that shouldn't have been in there. Was just experimenting with getting it to build.

MarkIngramUK:
> BTW, `clang-cl` absolutely does identify correctly, sounds like someone is wrong with your setup?

I was using the VS generator with CMAKE_GENERATOR_TOOLSET=LLVM, so I assume the CMAKE_C_SIMULATE_ID isn't set to to MSVC in that case. Using the following command line does simulate correctly:

`cmake -GNinja ../libdispatch -DCMAKE_C_COMPILER=clang-cl -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_LINKER="C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\bin\Hostx64\x64\link.exe"`

Though it does fail to build with `'sys/sysctl.h' file not found`.

I suppose I should have asked ahead of making the PR, but is MSVC support something that should be added? Or is intentional that the only supported compiler is clang?

compnerd:
MSVC is not supported, and is unlikely to be useful.  The only supported compiler is clang, even GCC is unsupported.  The problem is that the code base relies on clang specific extensions, and the blocks support in clang.  clang-cl is supported as it is just a different frontend to clang, but clang nonetheless.  I had tried at one point to try to support MSVC, but found it too difficult to work around the limitations of the compiler.

MarkIngramUK:
OK - in that case this PR should probably be abandoned. We have a very old Win32 port of libdispatch, so I was looking to move to the official fork.

ktopley-apple:
> OK - in that case this PR should probably be abandoned. We have a very old Win32 port of libdispatch, so I was looking to move to the official fork.

In that case,  would you mind closing this PR?

MarkIngramUK:
Done!

PR #465, opened 2019-03-22
Use SYS_gettid on Android

triplef:
The system-provided gettid() was previously used on Android, which broke with the recent symbol change to _gettid() (#461).

Since SYS_gettid is available in recent NDKs (at least since version 21) we can now use that instead and remove the special-casing.


ktopley-apple:
@swift-ci please test

PR #469, opened 2019-04-05
event: create a typedef for the unote ident type

compnerd:
File handles and sockets on Windows are pointer sized.  Ensure that we
have enough space in the unote to track the source properly.

### DIFF
--- a/src/event/event.c
+++ b/src/event/event.c
@@ -51,7 +51,7 @@ _dispatch_unote_create(dispatch_source_type_t dst,
 	}
 	du->du_type = dst;
 	du->du_can_be_wlh = dst->dst_per_trigger_qos;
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

ktopley-apple:
@compnerd do you plan to follow up with another PR to rearrange the fields to reclaim the extra space, as discussed offline?

compnerd:
@ktopley-apple - as it turns out, I think that the `du_fflags` field is touched upon by `event.c`, so, I'm not sure if it is the wisest.  I think that I want to hold off on that until I have a working implementation of the file sources on Windows.  After that, if possible, I'll reclaim the bits.

ktopley-apple:
@compnerd OK, since this only affects the size on Windows, let's merge.

PR #470, opened 2019-04-05
Port most remaining tests to Windows

adierking:
This patch set provides Windows support for most of the remaining test suite. Many of the tests do not pass yet, and I do not expect the test ports themselves to be perfect, but being able to easily build and run everything should be a big step towards fixing all of the remaining issues with the Windows port of Dispatch. In particular, @compnerd and I have been working on implementing socket and file event sources, and we need to be able to easily test everything.

The only test that still needs to be ported is dispatch_io, which may take some time to do since it uses <fts.h>. For now, that test is disabled on Windows just so that it's easy to build everything without ignoring errors. (It won't pass anyway until we have more event sources implemented.)

Most of the complexity here is related to file I/O, because Dispatch will use `HANDLE`s for event sources. In order to cut down on the amount of platform-specific code in tests, and to make these changes easy to inspect, I introduced wrapper functions around the `dispatch_fd_t` type. On Windows, these functions will use Win32 APIs and work with `HANDLE`s, and on other platforms they will forward to the corresponding system calls. I also replaced usages of `fstat()` (for getting file sizes) with calls to `dispatch_test_fd_lseek()` so that we don't need an `fstat()` equivalent on Windows.

While msvcrt already provides POSIX-like I/O functions, they work with virtual file descriptors instead of handles, and these custom implementations let us manage temporary files correctly. (msvcrt does not open files with `FILE_SHARE_DELETE`!)

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -196,7 +196,9 @@ add_unit_test(dispatch_c99 NO_BSD_OVERLAY SOURCES dispatch_c99.c)
 add_unit_test(dispatch_plusplus SOURCES dispatch_plusplus.cpp)
 
 # test-specific link options
-if(NOT WIN32)
+if(WIN32)
+  target_link_libraries(dispatch_io_net PRIVATE WS2_32)
+else()
   target_link_libraries(dispatch_group PRIVATE m)
   target_link_libraries(dispatch_timer_short PRIVATE m)
 endif()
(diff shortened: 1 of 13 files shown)


adierking:
cc @compnerd @ktopley-apple @MadCoder

compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple

PR #471, opened 2019-04-10
build: install swiftDispatch.dll to bin

compnerd:
The DLLs are runtime components.  The Windows target installs the
runtime components into the bin directory, and the import libraries into
the lib directory.  This enables the runtime setup to be simplified on
Windows.


compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple

PR #477, opened 2019-05-03
SR-9033: handle EPOLLHUP on Linux

adierking:
If epoll_wait() reports EPOLLHUP, we must respond immediately and
unregister the event or else Dispatch will go into a spinloop.

This makes dispatch_io_pipe_close (#476) pass on Linux with identical
output to Darwin.

PR #476, opened 2019-05-03
dispatch_io pipe close test

weissi:
this is a test that documents the problem seen in
- https://bugs.swift.org/browse/SR-9033
- https://bugs.swift.org/browse/SR-5773

PR #478, opened 2019-05-06
Combined PR: SR-9033 handle EPOLLHUP

gottesmm:
This is just a combination of these two PRs combining authorship information:

#477 
#476 

As one can see #476 fails with ToT. Lets see if with 477 it passes and then lets land this sucker!


gottesmm:
@swift-ci test

weissi:
We should also back-port this to the swift-5.0-branch branch

PR #479, opened 2019-05-09
Swift: correct dispatch source construction on Win32

compnerd:
Windows uses handles internally for the dispatch sources.  Convert the
`fileDescriptor` parameter on the incoming point to `HANDLE`s when
building the `DispatchSource`.  This fixes passing invalid parameters to
the Windows system functions.

### DIFF
--- a/src/swift/Source.swift
+++ b/src/swift/Source.swift
@@ -12,6 +12,9 @@
 
 import CDispatch
 import _SwiftDispatchOverlayShims
+#if os(Windows)
+import WinSDK
+#endif
 
 extension DispatchSourceProtocol {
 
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple

src/swift/Source.swift
```diff
@@ -179,7 +182,13 @@ extension DispatchSource {
 #endif
 
 	public class func makeReadSource(fileDescriptor: Int32, queue: DispatchQueue? = nil) -> DispatchSourceRead {
-		let source = dispatch_source_create(_swift_dispatch_source_type_READ(), UInt(fileDescriptor), 0, queue?.__wrapped)
+#if os(Windows)
+		let handle: UInt = UInt(_get_osfhandle(fileDescriptor))
```
ktopley-apple: Strictly speaking you don't need to declare the type here.

PR #480, opened 2019-05-10
build: ensure that we link against the correct library

compnerd:
Ensure that we link against the correct VC runtime libraries.
Additionally, enable the macros to indicate that we are linking against
the VC runtimes dynamically to get proper DLL storage.  This fixes
memory issues and file descriptor table synchrony.


compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple

PR #481, opened 2019-05-12
swift: add a set of Windows extensions

compnerd:
This adds a new set of API overloads which provide an interface which
use the HANDLE rather than file descriptor.  This makes it easier to
create the dispatch sources without having to convert the file handle to
a file descriptor.


compnerd:
@swift-ci please test

PR #483, opened 2019-05-16
When adding conditionally included compiler flags, dont use the CFLAGS section

spevans:
- cmake/modules/SwiftSupport.cmake prepends each flag with '-Xcc' but
  if the flag evaluates to empty then an extra -Xcc is added to the
  compiler options.

- Explicitly add the option with a conditional '-Xcc' before it as well.

- This fixes the "warning: argument unused during compilation: '-Xcc'"
  seen in the error logs.


spevans:
@compnerd I cant actually test this on windows, does it look correct to you?

spevans:
@swift-ci test

compnerd:
Yeah, this seems reasonable - though its all going to go away soon.

compnerd:
CC: @ktopley-apple

PR #484, opened 2019-05-16
dispatch_group and dispatch_starfish are failing sporadically in CI

jckarter:
rdar://problem/50861949


jckarter:
@swift-ci Please test

PR #486, opened 2019-05-17
build: use the new `-libc` flag for Windows

compnerd:
Windows has ABI incompatible system libraries which must be explicitly
selected.  Adjust the build system to enable the user to control the
variant of the library that we build against.  This simplifies the build
and ensures that the correct library is selected.


compnerd:
apple/swift#24849

@swift-ci please test

compnerd:
Ugh, this isn't ready yet here since this is a driver level flag and we are still using the frontend here.

PR #493, opened 2019-06-11
io: free Windows operation data with _aligned_free()

adierking:
`_dispatch_operation_perform()` uses `_aligned_malloc()` on Windows to
allocate the data buffer. This means that we cannot use
`DISPATCH_DATA_DESTRUCTOR_FREE` when we create the `dispatch_data_t`.
Provide a custom destructor which correctly frees the buffer.

Discovered while porting the dispatch_io test to Windows.


adierking:
cc @compnerd @ktopley-apple

ktopley-apple:
@swift-ci please test

PR #494, opened 2019-06-11
io: fix dispatch_io_create_with_path() on Windows

adierking:
This function assumes POSIX-style paths. Add Windows-specific checks for
absolute paths and separators.

Additionally, `_dispatch_fd_entry_guarded_open()` does not handle the
open flags correctly. `O_RDONLY` is 0, so we need to use a switch
statement to check for it. `O_EXCL` is not handled at all. `O_TRUNC`'s
behavior needs to change depending on whether `O_CREAT` is specified.
Finally, we must open files with maximum sharing to match the semantics
of other platforms. (The dispatch_io test already has a handle open when
it calls `dispatch_io_create_with_path()`.)

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -406,15 +406,29 @@ dispatch_io_create_f(dispatch_io_type_t type, dispatch_fd_t fd,
 			^(int error){ cleanup_handler(context, error); });
 }
 
+#if defined(_WIN32)
+#define _is_separator(ch) ((ch) == '/' || (ch) == '\\')
+#else
+#define _is_separator(ch) ((ch) == '/')
+#endif
+
 dispatch_io_t
 dispatch_io_create_with_path(dispatch_io_type_t type, const char *path,
 		int oflag, mode_t mode, dispatch_queue_t queue,
 		void (^cleanup_handler)(int error))
 {
-	if ((type != DISPATCH_IO_STREAM && type != DISPATCH_IO_RANDOM) ||
-			!(*path == '/')) {
+	if (type != DISPATCH_IO_STREAM && type != DISPATCH_IO_RANDOM) {
 		return DISPATCH_BAD_INPUT;
 	}
+#if defined(_WIN32)
+	if (PathIsRelativeA(path)) {
+		return DISPATCH_BAD_INPUT;
+	}
+#else
+	if (!_is_separator(*path)) {
+		return DISPATCH_BAD_INPUT;
+	}
+#endif
 	size_t pathlen = strlen(path);
 	dispatch_io_path_data_t path_data = malloc(sizeof(*path_data) + pathlen+1);
 	if (!path_data) {
@@ -449,9 +463,15 @@ dispatch_io_create_with_path(dispatch_io_type_t type, const char *path,
 				break;
 			default:
 				if ((path_data->oflag & O_CREAT) &&
-						(*(path_data->path + path_data->pathlen - 1) != '/')) {
+						!_is_separator(*(path_data->path + path_data->pathlen - 1))) {
 					// Check parent directory
-					char *c = strrchr(path_data->path, '/');
+					char *c = NULL;
+					for (ssize_t i = (ssize_t)path_data->pathlen - 1; i >= 0; i--) {
+						if (_is_separator(path_data->path[i])) {
+							c = &path_data->path[i];
+							break;
+						}
+					}
 					dispatch_assert(c);
 					*c = 0;
 					int perr;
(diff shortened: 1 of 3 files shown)


adierking:
cc @compnerd @ktopley-apple

src/io.c
```diff
@@ -406,15 +406,34 @@ dispatch_io_create_f(dispatch_io_type_t type, dispatch_fd_t fd,
...
+	if (type != DISPATCH_IO_STREAM && type != DISPATCH_IO_RANDOM) {
+		return DISPATCH_BAD_INPUT;
+	}
+#if defined(_WIN32)
+	// Absolute paths begin with either a separator or a character followed by a
+	// colon and a separator
+	if (path[0] == '\0') {
+		return DISPATCH_BAD_INPUT;
+	}
+	if (!is_separator(path[0]) && !(path[1] == ':' && is_separator(path[2]))) {
```
compnerd: This is correct, but a little subtle as it happens to also correctly handle UNC style paths which are absolute.

src/io.c
```diff
@@ -449,9 +468,14 @@ dispatch_io_create_with_path(dispatch_io_type_t type, const char *path,
 				break;
 			default:
 				if ((path_data->oflag & O_CREAT) &&
-						(*(path_data->path + path_data->pathlen - 1) != '/')) {
+						!is_separator(*(path_data->path + path_data->pathlen - 1))) {
 					// Check parent directory
-					char *c = strrchr(path_data->path, '/');
+					char *c = NULL;
+					for (char *i = path_data->path; *i != '\0'; i++) {
```
compnerd: This is a slight change in semantics: `strchr` rather than than `strrchr`.

src/io.c
```diff
@@ -1287,18 +1311,31 @@ _dispatch_fd_entry_guarded_open(dispatch_fd_entry_t fd_entry, const char *path,
...
-	return (dispatch_fd_t)CreateFile(path, dwDesiredAccess, 0, NULL, dwCreationDisposition, 0, NULL);
+		if (oflag & _O_EXCL) {
+			dwCreationDisposition = CREATE_NEW;
+		} else if (oflag & _O_TRUNC) {
+			dwCreationDisposition = CREATE_ALWAYS;
+		}
+	} else if (oflag & _O_TRUNC) {
+		dwCreationDisposition = TRUNCATE_EXISTING;
+	}
+	return (dispatch_fd_t)CreateFile(path, dwDesiredAccess,
```
compnerd: Would be nice to use `CreateFileW` so that we can handle file names which may contain non-ASCII values.

src/io.c
adierking: The loop doesn't break after it finds a separator.

src/io.c
adierking: I'll import ShLwApi and switch this to `!PathIsRelative()`.

src/io.c
adierking: I agree, however, the Windows port of dispatch currently does not handle Unicode well in general. I'd prefer to do a follow-up which fixes this and other similar cases.

compnerd: Sounds reasonable.

src/io.c
compnerd: True, but, why go forward rather than backwards?

src/io.c
compnerd: Sounds good

src/io.c
adierking: Ah, I missed that we know the length of the path. I can reverse it then.

adierking:
Addressed review comments. cc @compnerd

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

src/io.c
```diff
@@ -449,9 +463,15 @@ dispatch_io_create_with_path(dispatch_io_type_t type, const char *path,
 				break;
 			default:
 				if ((path_data->oflag & O_CREAT) &&
-						(*(path_data->path + path_data->pathlen - 1) != '/')) {
+						!_is_separator(*(path_data->path + path_data->pathlen - 1))) {
 					// Check parent directory
-					char *c = strrchr(path_data->path, '/');
+					char *c = NULL;
+					for (size_t i = path_data->pathlen; i-- > 0;) {
```
ktopley-apple: Although this code is correct, I found it confusing at first because it appears that the start index should be path_data->pathlen - 1, but this is not the case because i-- decrements before each pass of the loop. I understand the reason for the - size_t is unsigned. Maybe it would be clearer to cast to ssize_t:

for (ssize_t i = (ssize_t)path_data->pathlen - 1; i >= 0; i--)

adierking: I forgot that Dispatch declares `ssize_t` on Windows. (It normally isn't present.) That sounds better, thanks!

ktopley-apple:
@swift-ci please test

PR #496, opened 2019-06-12
shims: fix TAILQ_CONCAT() edge cases

adierking:
This needs to set the `te_next` of the last element in `head1` and also
handle the case where `head1` is empty. This fixes a hang in the
dispatch_context_for_key test on Windows because the the queue-specific
list was getting corrupted in `_dispatch_queue_specific_head_dispose()`.


adierking:
cc @compnerd @ktopley-apple

compnerd:
@swift-ci please test

PR #498, opened 2019-06-12
shims: use the precise Windows time functions

adierking:
`_dispatch_uptime()` and `_dispatch_monotonic_time()` are currently
using interrupt time APIs which return values cached by the scheduler.
This is causing the dispatch_after test to fail on Windows. Switch them
to the "Precise" APIs which incorporate a hardware time reading into
their results. Unfortunately, these functions are not exported by any
import libraries and we have to locate them at runtime.

Also, `_dispatch_monotonic_time()` should not be using the unbiased
version of the API because it is expected to include suspend time.

### DIFF
--- a/src/shims/generic_win_stubs.c
+++ b/src/shims/generic_win_stubs.c
@@ -1,24 +1,40 @@
 #include "internal.h"
 
-/*
- * This file contains stubbed out functions we are using during
- * the initial Windows port.  When the port is complete, this file
- * should be empty (and thus removed).
- */
+typedef void (WINAPI *_precise_time_fn_t)(PULONGLONG);
 
-void
-_dispatch_runloop_queue_dispose(dispatch_queue_t dq DISPATCH_UNUSED,
-		bool *allow_free DISPATCH_UNUSED)
+DISPATCH_STATIC_GLOBAL(dispatch_once_t _dispatch_precise_time_pred);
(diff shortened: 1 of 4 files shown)


adierking:
cc @compnerd @ktopley-apple

src/shims/generic_win_stubs.c
```diff
@@ -1,24 +1,40 @@
...
+	if (!_dispatch_QueryInterruptTimePrecise_ptr) {
+		DISPATCH_INTERNAL_CRASH(0, "could not locate QueryInterruptTimePrecise");
+	}
+	if (!_dispatch_QueryUnbiasedInterruptTimePrecise_ptr) {
+		DISPATCH_INTERNAL_CRASH(0, "could not locate QueryUnbiasedInterruptTimePrecise");
+	}
 }
 
 void
-_dispatch_runloop_queue_xref_dispose(dispatch_queue_t dq DISPATCH_UNUSED)
```
compnerd: Is this not used anymore?

compnerd:
@swift-ci please test

src/shims/generic_win_stubs.c
adierking: This file had been removed from the build but not deleted. We have to remove these stubs or else we get duplicate symbol errors.

compnerd: I think maintaining the existing structure is preferable since it mirrors the other targets.

PR #501, opened 2019-06-18
event: support pipe sources on Windows

adierking:
This implements full support for using pipe sources on Windows to
trigger an event when a pipe becomes readable or writable.

Due to differences between the Windows asynchronous I/O model and the
Unix model, this is rather complex. On Windows, the standard mechanism
to achieve asynchronous pipe I/O is to just read or write from the pipe
and then get notified once the operation completes. Unlike the Unix
`select()`/`poll()` model, there is no way to simply know when a pipe
becomes readable or writable without actually running an operation. So
we need to resort to several tricks in order to achieve the semantics
that Dispatch wants here.

To monitor a pipe for readability, we take advantage of the fact that a
zero-byte `ReadFile()` on a pipe will block until data becomes available
in the pipe. A muxnote which monitors a pipe for reading will spin up a
lightweight thread which repeatedly calls `ReadFile()` (blocking) on the
pipe and posts back to the I/O completion queue when it returns.

To monitor pipes for writability, we use the `NtQueryInformationFile()`
kernel API to get the amount of available space in the pipe's write
buffer. There is no way to block here, so we have no choice but to
continually disarm and rearm unotes until space becomes available. This
is inefficient, but it generally seems to work OK.

In order to test this, I implemented a new dispatch_io_pipe test which
performs various read and write operations on pipes. On Windows, this
will run the tests on most of the different pipe kinds (anonymous,
named, inbound, outbound, overlapped). This caught a lot of issues in
the Windows `_dispatch_operation_perform()` which I fixed along the way.
The dispatch_io and dispatch_io_pipe_close tests pass as well with my
other pull request applied.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -2410,7 +2410,43 @@ syscall:
 	if (op->direction == DOP_DIR_READ) {
 		if (op->params.type == DISPATCH_IO_STREAM) {
 #if defined(_WIN32)
-			ReadFile((HANDLE)op->fd_entry->fd, buf, (DWORD)len, (LPDWORD)&processed, NULL);
+			HANDLE hFile = (HANDLE)op->fd_entry->fd;
+			BOOL bSuccess;
+			if (GetFileType(hFile) == FILE_TYPE_PIPE) {
+				OVERLAPPED ovlOverlapped = {};
+				DWORD dwTotalBytesAvail;
+				bSuccess = PeekNamedPipe(hFile, NULL, 0, NULL,
+						&dwTotalBytesAvail, NULL);
+				if (bSuccess) {
+					if (dwTotalBytesAvail == 0) {
+						err = EAGAIN;
+						goto error;
+					}
+					len = MIN(len, dwTotalBytesAvail);
+					bSuccess = ReadFile(hFile, buf, (DWORD)len,
+							(LPDWORD)&processed, &ovlOverlapped);
+				}
+				if (!bSuccess) {
+					DWORD dwError = GetLastError();
+					if (dwError == ERROR_IO_PENDING) {
+						bSuccess = GetOverlappedResult(hFile, &ovlOverlapped,
+								(LPDWORD)&processed, /* bWait */ TRUE);
+						dwError = GetLastError();
+					}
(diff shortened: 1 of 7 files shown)


adierking:
cc @compnerd @ktopley-apple

src/io.c
```diff
@@ -152,7 +152,7 @@ enum {
 #endif // DISPATCH_IO_DEBUG
 
 #define _dispatch_fd_debug(msg, fd, ...) \
-		_dispatch_io_log("fd[0x%" PRIx64 "]: " msg, fd, ##__VA_ARGS__)
+		_dispatch_io_log("fd[0x%lld]: " msg, (long long)(fd), ##__VA_ARGS__)
```
ktopley-apple: Is this a necessary change? Recent PRs have been moving toward the PRxNN style of formatting specifiers.

src/io.c
```diff
@@ -1385,7 +1385,7 @@ _dispatch_fd_entry_create_with_fd(dispatch_fd_t fd, uintptr_t hash)
 	// On fds lock queue
 	dispatch_fd_entry_t fd_entry = _dispatch_fd_entry_create(
 			_dispatch_io_fds_lockq);
-	_dispatch_fd_entry_debug("create: fd %" PRId64, fd_entry, fd);
+	_dispatch_fd_entry_debug("create: fd %lld", fd_entry, (long long)fd);
```
ktopley-apple: Same as above.

src/io.c
adierking: This was giving me trouble on Linux when building with `DISPATCH_ENABLE_ASSERTS` because file descriptors are `int`. This is not required for this change. I'll remove this patch along with the one below and we can fix them as a follow-up.

ktopley-apple:
@swift-ci please test

adierking:
Fixed the build failure in Release mode and a merge conflict that arose. cc @ktopley-apple

ktopley-apple:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #502, opened 2019-06-29
sys/sysctl.h is deprecated on Linux

tachoknight:
A warning has been added to sys/sysctl.h that is treated as an error when compiling; see https://patches-gcc.linaro.org/patch/19443/#34639 for the discussion. This patch adds a guard to prevent the file from being included when compiling on Linux.


ktopley-apple:
@swift-ci please test

PR #503, opened 2019-07-13
build: add DT_SONAME for ELF libraries

compnerd:
ELF libraries should have DT_SONAME set so that they can be loaded
properly.  This is needed currently since we do not use proper CMake
support for Swift and that requires that we pass along the `-soname`
flag to the linker ourselves.  Account for that.  This improves loading
of swiftDispatch on android.

### DIFF
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -133,7 +133,11 @@ function(add_swift_target target)
   endif()
 
   if(AST_LIBRARY)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test and merge

compnerd:
CC: @ktopley-apple

compnerd:
@swift-ci please test and merge

compnerd:
@swift-ci please test

PR #504, opened 2019-07-15
build: correct the `DT_SONAME`

compnerd:
When ninja is invoked from a subdirectory it will convert the path to a
relative path, and that resulted in incorrect test results. This
ensures that we get the correct soname.


compnerd:
@swift-ci please test

PR #505, opened 2019-07-16
Address TODO to unblock master-next.

pschuh:
For reference, this is a build that fails: https://ci.swift.org/view/swift-master-next/job/oss-swift-incremental-RA-linux-ubuntu-16_04-master-next/11406/console
It repeatedly complains about "-momit-leaf-frame-pointer" being unused.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,3 +1,4 @@
+include(CheckCCompilerFlag)
 
 include(SwiftSupport)
 include(DTrace)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -229,11 +230,19 @@ if("${CMAKE_C_SIMULATE_ID}" STREQUAL "MSVC")
                          PRIVATE
                            -Xclang -fblocks)
 else()
-  # FIXME(compnerd) add check for -momit-leaf-frame-pointer?
-  target_compile_options(dispatch
-                         PRIVATE
-                           -fblocks
-                           -momit-leaf-frame-pointer)
+  check_c_compiler_flag("-momit-leaf-frame-pointer -Werror -Wall -O3" C_SUPPORTS_OMIT_LEAF_FRAME_POINTER)
```
compnerd: `-Werror -Wall -O3` is not required when using `check_c_compiler_flag`.  See the documentation at https://cmake.org/cmake/help/latest/module/CheckCCompilerFlag.html.

I think that we should just do two `target_compile_options` calls here.

```
check_c_compile_flag("-momit-leaf-frame-pointer" CC_SUPPORTS_MOMIT_LEAF_FRAME_POINTER)
target_compile_options(dispatch PRIVATE -fblocks)
if(CC_SUPPORTS_MOMIT_LEAF_FRAME_POINTER)
  target_compile_options(dispatch PRIVATE -momit-leaf-frame-pointer)
endif()

```

pschuh: check_c_compiler_flag just runs the compilation command with the provided flag. Unfortunately, clang will only complain with this combination of flags.

compnerd: Blah, that is annoying.

compnerd:
@pschuh - thanks for this patch and for the quick turn-around on the review!

compnerd:

PR #508, opened 2019-07-23
Remove Android Platform level for swiftc invocations

ephemer:
`swiftc` breaks when calling it from a CMake context with the `-DCMAKE_TOOLCHAIN_FILE` parameter referencing the Android NDK's CMake toolchain. The issue reported is `<unknown>:0: error: unable to load standard library for target 'x86_64-none-linux-android21'`

By removing the Android API level number from the end of the LLVM triple in the swift support we can ensure the triple containing the API level still gets provided to `clang` for C compilation without breaking `swiftc`.

Note: this change only affects Android. I am pretty sure there are no other Android triples that end with a number that we'd be breaking here.

### DIFF
--- a/cmake/modules/SwiftSupport.cmake
+++ b/cmake/modules/SwiftSupport.cmake
@@ -12,6 +12,13 @@ function(add_swift_target target)
   set(link_flags ${CMAKE_SWIFT_LINK_FLAGS})
 
   if(AST_TARGET)
+    if(ANDROID)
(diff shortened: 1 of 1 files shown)


compnerd:
I think that we should be doing this in the swift driver not in the CMake rules.

ephemer:
@compnerd fine by me but I couldn't find a working (let alone clean) way of doing this for Android-only in the swift repo, so suggestions and alternatives are very welcome :)

PR #509, opened 2019-07-23
build: support `CMAKE_SWIFT_COMPILER_TARGET`

compnerd:
This adds support for a new variable `CMAKE_SWIFT_COMPILER_TARGET` which
mirror's CMake's `CMAKE_Swift_COMPILER_TARGET`.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -171,7 +171,7 @@ if(ENABLE_SWIFT)
                       $<$<PLATFORM_ID:Windows>:-Xcc>
                       $<$<PLATFORM_ID:Windows>:-D_DLL>
                     TARGET
-                      ${CMAKE_C_COMPILER_TARGET})
+                      ${CMAKE_SWIFT_COMPILER_TARGET})
 endif()
 if(ENABLE_DTRACE)
   dtrace_usdt_probe(${CMAKE_CURRENT_SOURCE_DIR}/provider.d


ephemer:
I'm a bit hazy on CMake's case sensitivity (which has been an issue for the Android build in this repo before)- is there are reason to spell this differently compared to what CMake expects internally?

compnerd:
Yes, it is “misspelt” because right now all the builds use the old spelling of `CMAKE_SWIFT_COMPILER`.

ephemer:
@compnerd maybe we should add a step that sets `CMAKE_SWIFT_COMPILER_TARGET` if (only) `CMAKE_C_COMPILER_TARGET` has been set to avoid breaking everyone's builds

compnerd:
Well, the use of `CMAKE_SWIFT_COMPILER_TARGET` is only useful when doing a cross-compile, most people aren’t doing that, so I don’t really think that this would break most people’s build.

ktopley-apple:
@swift-ci please test

vgorloff:
FYI: Possible values for `CMAKE_SWIFT_COMPILER_TARGET` for Android cross-compile builds:

```
-DCMAKE_SWIFT_COMPILER_TARGET=armv7-none-linux-androideabi
-DCMAKE_SWIFT_COMPILER_TARGET=aarch64-none-linux-android
-DCMAKE_SWIFT_COMPILER_TARGET=i686-none-linux-android
-DCMAKE_SWIFT_COMPILER_TARGET=x86_64-none-linux-android
```

PR #510, opened 2019-08-14
shims: always replace `_Bool` with `bool` in C++

compnerd:
This generalises the path to always perform this substitution in C++
mode.  This should repair the build of libdispatch on android.


compnerd:
@swift-ci please test

compnerd:
CC: @ktopley-apple 
This fixes the build on android which broke recently.

PR #511, opened 2019-08-16
event: simplify the Windows file source implementation

adierking:
Currently, file sources are all registered with a global
`_dispatch_file_handles` list that has a lock around it. Remove this in
favor of posting muxnote pointers to the I/O completion port using the
same mechanism as the other source types.


adierking:
cc @compnerd @ktopley-apple

ktopley-apple:
@swift-ci please test

PR #512, opened 2019-08-21
event: implement source muxing on Windows

adierking:
Implement the cases in `_dispatch_unote_register_muxed()` and
`_dispatch_unote_unregister_muxed()` for when multiple event sources are
open on a handle and we need to combine them.

The test suite doesn't hit these codepaths anywhere and we haven't run
into issues with Foundation yet, so I added a dispatch_io_muxed test
which opens multiple sources on a file/pipe/socket and checks that
events fire correctly.

### DIFF
--- a/src/event/event_windows.c
+++ b/src/event/event_windows.c
@@ -174,27 +174,62 @@ _dispatch_muxnote_create(dispatch_unote_t du,
 }
 
 static void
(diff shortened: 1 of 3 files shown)


adierking:
cc @compnerd @ktopley-apple

ktopley-apple:
@swift-ci please test

PR #513, opened 2019-08-24
build: remove unnecessary flag to control the linker

compnerd:
The `SWIFT_LINKER_FLAGS` allows the user to specify the linker that they
want for Swift and C/C++ have their own flags which let you control
flags which should permit control of the C/C+ linker.  This removes the
unnecessary logic for controlling the linker.  This simplification
always paves the way for migrating further to CMake 3.15+.


compnerd:
@swift-ci please test

PR #515, opened 2019-08-29
build: place executable content into the root of the build

compnerd:
Adjust the output location of the generated executable content to the root of
the build tree.  This is needed primarily on Windows where there is no concept
of a RPATH, and the current directory is scanned for the dependent libraries.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,6 +19,12 @@ set(CMAKE_CXX_STANDARD 11)
 
 set(CMAKE_C_VISIBILITY_PRESET hidden)
 
+# NOTE(compnerd) this is a horrible workaround for Windows to ensure that the
+# tests can run as there is no rpath equivalent and `PATH` is used to lookup the
+# libraries.
+set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
+
 set(CMAKE_THREAD_PREFER_PTHREAD TRUE)
 set(THREADS_PREFER_PTHREAD_FLAG TRUE)
 find_package(Threads REQUIRED)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -124,6 +124,7 @@ if(ENABLE_SWIFT)
                       module-maps
                       DispatchStubs
                     LINK_FLAGS
+                      -L $<TARGET_LINKER_FILE_DIR:DispatchStubs>
                       -lDispatchStubs
                       -L $<TARGET_LINKER_FILE_DIR:BlocksRuntime>
                       -lBlocksRuntime


compnerd:
@swift-ci please test

CMakeLists.txt
```diff
@@ -19,6 +19,12 @@ set(CMAKE_CXX_STANDARD 11)
 
 set(CMAKE_C_VISIBILITY_PRESET hidden)
 
+# NOTE(compnerd) this is a horrible workaround for Windows to ensure that the
```
jakepetroules: Can you please do this for Windows only, and not for other platforms? I don't think having an identical directory layout between platforms is necessary; each have their own conventions which we should follow.

compnerd: This is the same thing that LLVM does on all the platforms, as does the Swift project.  What do we gain by doing it as a special case here?  Furthermore, this is the build tree only, so having the identical directory layout is necessary to keep the overall packaging rules identical.  (e.g. copying of files).

compnerd:
@swift-ci please test

compnerd:
apple/swift-package-manager#2329

@swift-ci please test

compnerd:
apple/swift-package-manager#2329
@swift-ci please test

compnerd:
apple/swift-package-manager#2329
apple/swift-corelibs-libdispatch#2498

@swift-ci please test

compnerd:
apple/swift-package-manager#2329
apple/swift-corelibs-foundation#2498

@swift-ci please test

compnerd:
apple/swift-package-manager#2329
apple/swift-corelibs-foundation#2498

@swift-ci please test

CMakeLists.txt
jakepetroules: Ah, right, this is for the build layout, not the installed layout.

compnerd:

ktopley-apple:
Yes, done :-)

buttaface:
Unfortunately, @jakepetroules was right, as [there are Swift tests that look for libdispatch.so and other files under the old build layout](https://github.com/apple/swift/blob/51193f4c828e2a15fe9a65e212556cd044440905/test/lit.cfg#L1019) and are now disabled on the linux CI as a result of this pull, specifically all tests for the libdispatch and tsan_runtime features.

I confirmed by examining [the build log for this commit](https://ci.swift.org/job/swift-corelibs-libdispatch-PR-Linux/379/) and comparing with [the previous commit's build log](https://ci.swift.org/job/swift-corelibs-libdispatch-PR-Linux/369/): note that libdispatch and tsan_runtime are missing from the available features for lit.py before the Swift testsuite is run only in the first linked log. This can be verified further through tests like Sanitizers/tsan or stdlib/DispatchData going from passing to unsupported after this pull.

I ran into this because I noticed that these tests were being run in prior Swift test logs I had run natively on Android, but were disabled recently, which I tracked down to this pull.

buttaface:
@compnerd or @drodriguez, should I file a bug in the Swift JIRA for this tsan/libdispatch issue with the linux CI, so that enabling those tests again isn't forgotten?

PR #516, opened 2019-09-02
tests: for Android, check TMPDIR first in dispatch_io_muxed

buttaface:
With this fix, all the tests pass again when built and run natively on Android, as Android has no global /tmp directory but [the Termux app that I'm using](https://github.com/termux/termux-app) sets TMPDIR, which is the usual variable for POSIX.


ktopley-apple:
@swift-ci please test

PR #520, opened 2019-09-18
tests: correct the check function (NFCI)

compnerd:
The tests were using `test_double_less_than_or_equal` which raised a warning
when building on Linux-x86_64:

```
tests/dispatch_select.c:72:51: error: implicit conversion from 'ssize_t' (aka 'long') to 'double' may lose precision
tests/dispatch_select.c:131:61: error: implicit conversion from 'long' to 'double' may lose precision
```

Use `test_sizet_less_than_or_equal` instead.


compnerd:
@swift-ci please test

compnerd:

PR #522, opened 2019-09-22
queue: use the default stack size on Windows

adierking:
The Windows thread pool implementation is calling `_beginthreadex` with
a stack size of 64 KB. This seems like a misunderstanding to me. In the
pthread code, `_dispatch_mgr_root_queue_init` does call
`pthread_attr_setstacksize` with a 64 KB size. However, this only
applies to the pthread root queue manager, and pthread root queues are
only supported on Apple platforms anyway.

We can confirm that using the OS default stack size is the intended
semantic by looking at the `dispatch_select` test, which expects to be
able to stack-allocate a 500 KB buffer in an event handler. This test is
failing on Windows because there is not enough stack space. We've also
run into various related issues in Swift which only happen on Windows,
and I suspect that this is just because the stack size isn't so small on
other platforms.

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -5769,13 +5769,8 @@ _dispatch_root_queue_poke_slow(dispatch_queue_global_t dq, int n, int floor)
 #endif
 	do {
 		_dispatch_retain(dq); // released in _dispatch_worker_thread
-#if DISPATCH_DEBUG
-		unsigned dwStackSize = 0;
-#else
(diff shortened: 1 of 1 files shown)


adierking:
cc @compnerd @ktopley-apple

compnerd:
@swift-ci please test

compnerd:
Windows test run: https://dev.azure.com/compnerd/windows-swift/_build/results?buildId=9441&view=results

PR #530, opened 2019-09-30
build: make some linkage flags public

compnerd:
Some of the compiler and linker flags need to be used by dependent libraries
which requires making the flags public so that users of the libraries and the
export targets are setup properly.


compnerd:
@swift-ci please test

compnerd:

PR #531, opened 2019-10-24
Silence Clang -Wimplicit-int-float-conversion.

drodriguez:
In LLVM https://reviews.llvm.org/D64666 a new warning is introduced that
warns about possible lost in precision when converting some integrals
into floating point numbers.

When the code is compiled with a newer Clang, the warning will be
emitted for the code branches that use `long double` and reduce them to
`uint64_t`. To avoid that, a couple of static constants are created in
both the `long double` and the `__uint128_t` which holds the maximum
(casted in the case of `long double` to avoid the warning), and those
constants are used for the comparison.

In the case at hand, `UINT64_MAX` is not representable as a `long
double`, but `UINT64_MAX+1` is, which is the value that the constant
will be.

/cc @adierking @compnerd

### DIFF
--- a/src/benchmark.c
+++ b/src/benchmark.c
@@ -43,8 +43,10 @@ _dispatch_benchmark_init(void *context)
 	uint64_t start, delta;
 #if DISPATCH_SIZEOF_PTR == 8 && !defined(_WIN32)
 	__uint128_t lcost;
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @ktopley-apple

compnerd:
@swift-ci please test

shahmishal:

PR #532, opened 2019-10-29
Redundant return statements removed and some if statements flipped to better reflect the function's purpose.

pi1024e:
Small change, but many of these functions have inconsistent structures, so I flipped some of the if statements so they better align with the others. What this means is that the branch will have a call to the intended function and the second one what happens when the initial check fails.

I hope this helps!


pi1024e:
No update?

PR #539, opened 2020-03-25
IO: match the signature for `DispatchIO` with `dispatch_io_create`

compnerd:
The `DispatchIO` constructor maps to `dispatch_io_create`.  The value on
Windows for the `fd` (which is actually a `HANDLE`) is being truncated
since `dispatch_fd_t` is actually `Int` not `Int32` on Windows.  Use the
`dispatch_fd_t` to ensure that the right size is always passed along.

### DIFF
--- a/src/swift/IO.swift
+++ b/src/swift/IO.swift
@@ -67,7 +67,7 @@ extension DispatchIO {
 
 	public convenience init(
 		type: StreamType,
-		fileDescriptor: Int32,
+		fileDescriptor: dispatch_fd_t,
 		queue: DispatchQueue,
 		cleanupHandler: @escaping (_ error: Int32) -> Void)
 	{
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

compnerd:
With this change, it is now possible to write an echo server for Windows:

```swift
import Dispatch
import ucrt
import WinSDK

let rx: DispatchQueue = DispatchQueue(label: "rx", qos: .userInitiated)

let hIn: HANDLE = GetStdHandle(STD_INPUT_HANDLE)
if hIn == INVALID_HANDLE_VALUE { exit(EXIT_FAILURE) }

let io: DispatchIO = DispatchIO(type: .stream, fileDescriptor: Int(bitPattern: hIn), queue: rx) { error in
  guard error == 0 else { exit(EXIT_FAILURE) }
}
io.setLimit(lowWater: 1)
io.setLimit(highWater: Int.max)

while true {
  io.read(offset: 0, length: Int.max, queue: rx) { done, data, errorCode in
    guard errorCode == 0 else {
      print("IO Error \(errorCode)");
      exit(EXIT_FAILURE)
    }

    guard let data = data else { return }
    data.withUnsafeBytes { (ptr: UnsafePointer<CChar>) -> () in
      print(">> \(String(cString: ptr))")
    }
  }
}
```

gwynne:
Looks nice and straightforward to me.

hafixo:
Pull request

PR #540, opened 2020-05-02
tests: Suppress unused return value warnings

dduan:
These cause compile failures due to `-Werror -Wunused-result` when building on
NixOS. This PR upstreams patches need to package Swift on that OS.


dduan:
@swift-ci please test

dduan:
@swift-ci please test

PR #543, opened 2020-07-23
Modify headers so that GCC can use libdispatch

Ignition:
GCC doesn't have serveral extensions that clang has. This is fine for building libdispatch shared library, but using libdispatch's with just the function based variants should have headers which GCC can use.


compnerd:
@swift-ci please test

shahmishal:

PR #544, opened 2020-08-03
Add modulemap for static compilation

drexin:
When compiling statically, we have to link against DispatchStubs in addition to the other dependencies, so we are defining a separate modulemap for that.


drexin:
@swift-ci test linux

drexin:
@compnerd The only difference between the existing modulemap and the new one is 

```
link "DispatchStubs"
```

otherwise they are exactly the same. So as to why it is required, I don't know.

compnerd:
Oh, I get it.  Its for the autolinking to its dependencies from the Swift module.  LGTM with grumbling.

PR #546, opened 2020-08-28
[5.3] Add modulemap for static compilation

drexin:
When compiling statically, we have to link against DispatchStubs in addition to the other dependencies, so we are defining a separate modulemap for that.


drexin:
@swift-ci test

PR #541, opened 2020-05-06
[CMake] fix runpath for ELF platforms

buttaface:
Remove the absolute path to the host stdlib for libswiftDispatch.so and add `$ORIGIN` to it and libdispatch.so.

Otherwise, you see the following in the current official release for linux:
```
> readelf -d swift-5.2.3-RELEASE-ubuntu18.04/usr/lib/swift/linux/lib*ispatch.so | ag "File:|runpath"
File: swift-5.2.3-RELEASE-ubuntu18.04/usr/lib/swift/linux/libdispatch.so
File: swift-5.2.3-RELEASE-ubuntu18.04/usr/lib/swift/linux/libswiftDispatch.so
0x000000000000001d (RUNPATH)            Library runpath: [/home/buildnode/jenkins/workspace/oss-swift-5.2-package-linux-ubuntu-18_04/build/buildbot_linux/swift-linux-x86_64/lib/swift/linux]
```

PR #547, opened 2020-08-29
[5.3][CMake] fix runpath for ELF platforms

buttaface:
Remove the absolute path to the host stdlib for libswiftDispatch.so and add $ORIGIN to it and libdispatch.so.

It has been suggested to me by some Swift devs that this would be good to get in before the 5.3 release, a cherry-pick from #541. @compnerd, I can remove the `$ORIGIN` portions and just make this about removing the toolchain stdlib rpath if you prefer.

This should be completely safe for the 5.3 branch, as it's just removing an unused rpath and adding rpaths that are already used, in a more explicit way.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -155,6 +155,10 @@ if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
     "-Xlinker -alias_list -Xlinker ${PROJECT_SOURCE_DIR}/xcodeconfig/libdispatch.aliases")
 endif()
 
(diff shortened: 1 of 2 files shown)


buttaface:
@MadCoder or @drexin, would be good to get this in before the 5.3 release, has been in trunk safely for months now.

drexin:
@swift-ci test

buttaface:
Thanks, it would be good to get this and the equivalent Foundation pull in for the 5.3 release, as user executables link against them and this patch would close a security hole for that. The same XCTest and SPM pulls can wait for the 5.3.1 release, as those are holes when running the toolchain, which I presume is less serious.

PR #548, opened 2020-10-03
Light cleanup of the kevent backend.

3405691582:
Some functionality requires particular preprocessor feature flags to be
set but errors are raised if they aren't, so the remedy is to put these
behind relevant feature flags.
  * dispatch_kevent_t has a qos field only when DISPATCH_USE_KEVENT_QOS
  * _dispatch_kq_unote_update requires DISPATCH_HAVE_DIRECT_KNOTES as it
    is only called from functions behind this
  * _dispatch_workloop_actions and the associated enum are only used when
    DISPATCH_USE_KEVENT_WORKLOOP.

Some unused variable void casts are removed (guard_ptr and pp), which do
not refer to variables -- these may be upstream merge artefacts.

A type cast is made to match types for _dispatch_bug_kevent_client. This
function is incorrectly placed behind HAVE_MACH in src/init.c, but to
keep the scope and impact of this commit narrow, we will defer that
change.

Since EVFILT_FS is not supported on all kevent implementations, put these
behind an #ifdef. EVFILT_USER also requires similar handling, but since
EVFILT_USER kevents are used intrinsically as part of the functionality,
these will be handled in a separate commit.

### DIFF
--- a/src/event/event_kevent.c
+++ b/src/event/event_kevent.c
@@ -101,7 +101,9 @@ _evfiltstr(short filt)
 	_evfilt2(EVFILT_MACHPORT);
 	_evfilt2(DISPATCH_EVFILT_MACH_NOTIFICATION);
 #endif
+#ifdef EVFILT_FS
 	_evfilt2(EVFILT_FS);
+#endif
 	_evfilt2(EVFILT_USER);
 #ifdef EVFILT_SOCK
 	_evfilt2(EVFILT_SOCK);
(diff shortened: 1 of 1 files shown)


3405691582:
cc @ktopley-apple who may find some of this of interest.

3405691582:
ping.

3405691582:
ping.

3405691582:
ping.

3405691582:
Closing in favor of #559.

PR #549, opened 2020-10-14
HAVE_MACH is unnecessary here.

3405691582:
The kevent backend requires _dispatch_bug_kevent_client unconditionally.
We don't need to require this entire function require HAVE_MACH, just the
inner few that is already present behind this conditional.

### DIFF
--- a/src/init.c
+++ b/src/init.c
@@ -961,7 +961,6 @@ _dispatch_continuation_get_function_symbol(dispatch_continuation_t dc)
 	return dc->dc_func;
 }
 
-#if HAVE_MACH
 void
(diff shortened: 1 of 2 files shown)


3405691582:
cc @ktopley-apple 

In f03fa15144a4cfaeb5ac8d2ef29ce556c3d070ea this was put in to possibly work around Ubuntu clang crashes -- let me know if this issue is still relevant. Perhaps changing it to the DISPATCH_EVENT_BACKEND_KEVENT conditional is better, which might necessitate further code movement and refactoring.

3405691582:
ping.

3405691582:
Closing in favor of #559.

PR #550, opened 2020-10-14
fadvise need not be mandatory.

3405691582:
It is just a hint, after all, so it is not an error if it is unavailable.


3405691582:
ping.

3405691582:
Closing in favor of #559.

PR #551, opened 2020-11-03
Install Dispatch Exports in the Install Tree

gmittert:
When building foundation (or any project that depends on swiftDispatch)
using a built and installed toolchain that already contains libdispatch,
setting Dispatch_DIR to the source directory creates duplicate
definitions of the libdispatch headers.

This creates an additional dispatchConfig that gets installed to the
install tree that points to the installed headers rather than the source
headers.


gmittert:
cc @compnerd for cmake things

Note that I'm not terribly familiar with CMake exports so this might not be 100% proper.

gmittert:
@swift-ci please test

CMakeLists.txt
```diff
@@ -254,12 +254,19 @@ configure_file("${PROJECT_SOURCE_DIR}/cmake/config.h.in"
...
 
+if (BUILD_SHARED_LIBS)
+  set(swift_dir "swift")
+else()
+  set(swift_dir "swift_static")
+endif()
+string(TOLOWER ${CMAKE_SYSTEM_NAME} lower_system_name)
 
 if(ENABLE_SWIFT)
-  set(INSTALL_TARGET_DIR "${CMAKE_INSTALL_LIBDIR}/swift$<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:_static>/$<LOWER_CASE:${CMAKE_SYSTEM_NAME}>" CACHE PATH "Path where the libraries will be installed")
```
gmittert, line 259: Sadly, `INSTALL(EXPORT` doesn't expand macros.

gmittert:
@swift-ci please test

PR #552, opened 2020-11-04
cmake: Test compiler flags before using.

orbea:
I am not sure if you are interested in supporting gcc builds, but this fixes most if not all of the gcc problems in the cmake scripts. Of course the actual build still fails, but I would rather see how this is received before considering working more on this.


Ezra-Black:
To compile multiple sources and link them together?

orbea:
Incomplete, no upstream response.

PR #554, opened 2020-12-02
[android] Put in fixes for librt and armv7-a

buttaface:
Android doesn't have a separate librt, it's just part of libc. Also, the static build wasn't working for armv7-a, because the test executables wouldn't link with the multiple definition errors listed in android/ndk#176, so use the workaround given there.

I didn't have any problems when cross-compiling for 64-bit arches, but hit the first one because I happen to have an i686 libc package installed in Ubuntu and the armv7 build would try to use librt.so from that. The second libunwind issue has been around for awhile, and I worked around it before [by not building the tests for libdispatch](https://github.com/buttaface/swift-android-sdk/blob/main/swift-android-531.patch#L123), but this gets them to build and run.

@compnerd, let me know what you think.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -113,7 +113,9 @@ if(__BUILTIN_TRAP)
   set(HAVE_NORETURN_BUILTIN_TRAP 1)
 endif()
 
-find_package(LibRT)
+if(NOT CMAKE_SYSTEM_NAME STREQUAL Android)
+  find_package(LibRT)
+endif()
 
 check_function_exists(_pthread_workqueue_init HAVE__PTHREAD_WORKQUEUE_INIT)
 check_function_exists(getprogname HAVE_GETPROGNAME)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -88,6 +88,14 @@ function(add_unit_test name)
     target_compile_options(${name} PRIVATE -fblocks)
     target_compile_options(${name} PRIVATE -Wall -Wno-deprecated-declarations)
   endif()
+  # Without this flag, cross-compiling static test executables for Android armv7
+  # fails with the multiple definition errors seen in android/ndk#176, so I
+  # pulled in this workaround noted there. The tests build and run with this
+  # flag applied.
+  if(NOT BUILD_SHARED_LIBS AND CMAKE_SYSTEM_NAME STREQUAL Android AND
+     CMAKE_SYSTEM_PROCESSOR STREQUAL armv7-a)
+    target_link_options(${name} PRIVATE "LINKER:--allow-multiple-definition")
+  endif()
   target_link_libraries(${name}
                         PRIVATE
                           dispatch


CMakeLists.txt
```diff
@@ -113,7 +113,9 @@ if(__BUILTIN_TRAP)
   set(HAVE_NORETURN_BUILTIN_TRAP 1)
 endif()
 
-find_package(LibRT)
+if(NOT CMAKE_SYSTEM_NAME STREQUAL Android)
+  find_package(LibRT)
+endif()
```
compnerd, line 118: This is fine to do, but really unfortunate.  CMake already already can deal with librt properly on various targets, but we don't use CMake properly :-(.

tests/CMakeLists.txt
```diff
@@ -88,6 +88,10 @@ function(add_unit_test name)
     target_compile_options(${name} PRIVATE -fblocks)
     target_compile_options(${name} PRIVATE -Wall -Wno-deprecated-declarations)
   endif()
+  if(NOT BUILD_SHARED_LIBS AND CMAKE_SYSTEM_NAME STREQUAL Android AND
+     CMAKE_SYSTEM_PROCESSOR STREQUAL armv7-a)
+    target_link_options(${name} PRIVATE "LINKER:--allow-multiple-definition")
+  endif()
```
compnerd, line 94: I think this needs a huge comment.  This is indicative of an issue - the multiple definitions shouldnt be there.  The warnings indicate that there is a mix up between libgcc and libunwind, which is a really bad thing as it will crash at runtime in a difficult to diagnose manner (the internal layouts of the unwind context is not compatible/guaranteed).

CMakeLists.txt
buttaface: Yes, you could use `CMAKE_FIND_ROOT_PATH`, [like I do when cross-compiling the corelibs](https://github.com/apple/swift/pull/33724/files#diff-1a5a064af373f742294e3b312a5c8d0ab52f5aef78052490e6005165472cda48R2179), if there is an `librt.so` to find, but there is none on Android- it's just part of `libc.so`- so this is the right change for Android.

tests/CMakeLists.txt
buttaface: Yeah, I don't know exactly what's going on here, [their doc doesn't clear it up for me](https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md#Unwinding). All I know is that applying this workaround flag gets the test executables to build, and I tried moving a handful to Android and they ran fine. I will add a comment.

buttaface: Added the comment.

buttaface:
Ping, would be good to get this in before the 5.4 branch. @MadCoder or @drexin, maybe one of you could merge if Saleem is busy?

buttaface:
Ping, if someone can run the CI, we can get this in.

buttaface:
@benlangmuir, mind running the CI on this?

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

benlangmuir:
@swift-ci please test

buttaface:
A final CI run and this can be merged.

drexin:
@swift-ci test

buttaface:
CI failed because of unrelated lldb compilation error.

benlangmuir:
@swift-ci please test

buttaface:
Passed CI, can someone merge?

buttaface:
I just ran into the build failing when cross-compiling for Android AArch64 too, because it tries to link against the system libRT.so on Fedora Core x86_64 too.

@benlangmuir, mind merging?

benlangmuir:
Someone who works on this project should review/merge.  I'm not the right person to look at this.  It looks like @compnerd was looking at it earlier.

buttaface:
Alright, @MadCoder, mind reviewing?

buttaface:
Ping @drexin, can you get this simple pull in?

PR #555, opened 2021-01-22
cmake: link BlocksRuntime only on non-Darwin

holymonson:
On Darwin we have BlocksRuntime builtin, don't need to link BlocksRuntime::BlocksRuntime.


holymonson:
@swift-ci test

AtariDreams:
cc @etcwilde

triplef:
FYI this was previously also done a bit more comprehensively by @compnerd in #535, but never merged.

PR #556, opened 2021-01-29
Properly annotate fallthrough in kevent backend.

3405691582:
Building dispatch from swift sets -Werror -Wimplicit-fallthrough. It is
not enough to just comment the fallthrough; we have elsewhere defined
DISPATCH_FALLTHROUGH, so we might as well use it.


3405691582:
Closing in favor of #559.

PR #557, opened 2021-02-15
Fix _libdispatch_tsd_cleanup signature.

triplef:
Fixes the following error when building for Windows x86:

```
..\src\queue.c(7376,1): error: function declared 'stdcall' here was previously declared without calling convention
_libdispatch_tsd_cleanup(void *ctx)
^
..\src/shims/tsd.h(161,13): note: previous declaration is here
extern void _libdispatch_tsd_cleanup(void *ctx);
```

`DISPATCH_TSD_DTOR_CC` is part of the `_libdispatch_tsd_cleanup` signature in queue.c, but not in the external declaration in `tsd.h`. `DISPATCH_TSD_DTOR_CC` resolves to `__stdcall` on Windows.

### DIFF
--- a/src/shims/tsd.h
+++ b/src/shims/tsd.h
@@ -158,7 +158,7 @@ struct dispatch_tsd {
 extern _Thread_local struct dispatch_tsd __dispatch_tsd;
 
 extern void libdispatch_tsd_init(void);
-extern void _libdispatch_tsd_cleanup(void *ctx);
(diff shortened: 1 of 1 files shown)


AtariDreams:
@etcwilde we can close this since this was merged already

triplef:
Yup, this was done in #569.

PR #559, opened 2021-03-18
Support OpenBSD.

3405691582:
Prior separated PRs were closed in favor of this omnibus.

This PR implements functionality to properly support Dispatch on OpenBSD. This includes fixing some errors, portability cleanup, as well as a portable implementation of the Dispatch runloop and making sure that the kevent backend works properly and portably on kernels that may not support all kevent features (like user-driven kevents).

This PR influences the design of apple/swift-corelibs-foundation#3004 because CFRunLoop in Foundation peeks under the curtain of how the Dispatch runloop handles are implemented.


src/event/event_internal.h
```diff
@@ -125,6 +125,8 @@ _dispatch_timer_flags_from_clock(dispatch_clock_t clock)
 
 #if defined(_WIN32)
 typedef uintptr_t dispatch_unote_ident_t;
+#elif defined(__OpenBSD__)
+typedef uint32_t dispatch_unote_ident_t;
```
kithrup, line 129: Unless the typedef is going to change, that is an unnecessary ifdef, no?

src/event/event_kevent.c
```diff
@@ -727,6 +745,10 @@ _dispatch_kq_poll(dispatch_wlh_t wlh, dispatch_kevent_t ke, int n,
 		(void)avail;
 		const struct timespec timeout_immediately = {}, *timeout = NULL;
 		if (flags & KEVENT_FLAG_IMMEDIATE) timeout = &timeout_immediately;
+#if EVFILT_USER
```
kithrup, line 748: Should that be an ifdef, not an if?

3405691582: Fixed.

src/event/event_internal.h
3405691582: I was experimenting a little here and forgot to change this back. This needs to match `struct kevent`, so should be a `uintptr_t`, and necessitates some further changes.

3405691582:
Closed smaller PRs in favor of this omnibus, so it is easier to track and will centralize any future review.

buttaface:
@etcwilde, mind reviewing?

src/queue.c
```diff
@@ -6592,6 +6606,11 @@ _dispatch_runloop_queue_handle_dispose(dispatch_lane_t dq)
 #elif defined(__linux__)
 	int rc = close(handle);
 	(void)dispatch_assume_zero(rc);
+#elif defined(__unix__)
+	int rc = close((int)(handle & 0xffffffff));
```
kithrup, line 6610: That seems a strange constant to use.

kithrup (approved):
Other than the strange (1<<32 - 1) constant, this seems good to me, for whatever that's worth.

src/queue.c
3405691582: The handle packs both fds of the pipe into a single 64-bit integer -- this plucks out the lower 32-bits.

I'll define some macros to make this a little more obvious.

kithrup: Oh! How ... fun. ;)

buttaface:
@drexin, would you review?

buttaface:
@rokhinip, would you review?

3405691582:
Now that apple/swift-corelibs-foundation#3004 is merged and Dispatch is becoming more of a foundational library in Swift, it's all the more necessary that this gets some more eyeballs if at all possible.

tkremenek:
I spoke to @rokhinip offline, and @rokhinip will try and review this week or early next.

cmake/modules/SwiftSupport.cmake
```diff
@@ -27,6 +27,8 @@ function(get_swift_host_arch result_var_name)
     set("${result_var_name}" "armv7" PARENT_SCOPE)
   elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "armv7l")
     set("${result_var_name}" "armv7" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "amd64")
+    set("${result_var_name}" "amd64" PARENT_SCOPE)
   elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "AMD64")
```
compnerd, line 32: I wonder if we should change this to `MATCHES` instead.

private/private.h
```diff
@@ -191,6 +191,8 @@ void _dispatch_prohibit_transition_to_multithreaded(bool prohibit);
 typedef mach_port_t dispatch_runloop_handle_t;
 #elif defined(__linux__) || defined(__FreeBSD__)
 typedef int dispatch_runloop_handle_t;
+#elif defined(__unix__)
+typedef uint64_t dispatch_runloop_handle_t;
```
compnerd, line 195: This feels a bit ... worrisome to me.  If this ever gets re-ordered, we will have a subtle way to change the semantics.  Currently `__linux__` will define `dispatch_runloop_handle_t` to `int` rather than `uint64_t`.  Is there a reason to prefer the `__unix__` spelling here over `__OpenBSD__`?

src/event/event.c
```diff
@@ -768,7 +768,7 @@ _dispatch_timer_heap_update(dispatch_timer_heap_t dth,
 #define _dispatch_timer_du_debug(what, du) \
 		_dispatch_debug("kevent-source[%p]: %s kevent[%p] { ident = 0x%llx }", \
 				_dispatch_wref2ptr((du)->du_owner_wref), what, \
-				(du), (unsigned long long)((du)->du_ident))
+				(du), (unsigned long long)(du)->du_ident)
```
compnerd, line 771: This seems like an unnecessary change ... and I kinda do prefer the original variant.  Does it cause some issue to leave it as is?

src/event/event.c
```diff
@@ -792,7 +792,7 @@ static void
 _dispatch_timer_unote_disarm(dispatch_timer_source_refs_t dt,
 		dispatch_timer_heap_t dth)
 {
-	uint32_t tidx = dt->du_ident;
+	uint32_t tidx = (uint32_t)dt->du_ident;
```
compnerd, line 795: Do we need to worry about truncation?  IIRC, a timer on Windows would be a `HANDLE` which would be truncated by `uint32_t` as it should be `uintptr_t`.

src/event/event_kevent.c
```diff
@@ -388,16 +392,18 @@ _dispatch_kevent_print_error(dispatch_kevent_t ke)
...
+#endif
 	default:
 		// log the unexpected error
 		_dispatch_bug_kevent_client("kevent", _evfiltstr(ke->filter),
 				!ke->udata ? NULL :
 				ke->flags & EV_DELETE ? "delete" :
 				ke->flags & EV_ADD ? "add" :
 				ke->flags & EV_ENABLE ? "enable" : "monitor",
-				(int)ke->data, ke->ident, ke->udata, du);
+				(int)ke->data, ke->ident, (uint64_t)ke->udata, du);
```
compnerd, line 406: Is this explicitly sized to `uint64_t` or would `uintpr_t` be more appropriate?

src/event/event_kevent.c
```diff
@@ -528,11 +534,17 @@ _dispatch_kevent_merge_muxed(dispatch_kevent_t ke)
 	}
 }
 
+#define DISPATCH_KEVENT_ERSATZ_EVFILT_USER_MASK 0xfffffe00
```
compnerd, line 537: Does it make sense to have a top level macro for the use of this?  Something similar to `DISPATCH_USE_KEVENT_ERSATZ_USER_FILTER`?

src/io.c
```diff
@@ -2308,7 +2308,7 @@ _dispatch_operation_advise(dispatch_operation_t op, size_t chunk_size)
 		default: (void)dispatch_assume_zero(err); break;
 	}
 #else
-#error "_dispatch_operation_advise not implemented on this platform"
+	(void)err;
```
compnerd, line 2311: I'm kinda torn on this.  This makes it easier to port - you know immediately to handle this.

src/queue.c
```diff
@@ -6469,6 +6469,8 @@ _dispatch_runloop_handle_is_valid(dispatch_runloop_handle_t handle)
 	return MACH_PORT_VALID(handle);
 #elif defined(__linux__)
 	return handle >= 0;
+#elif defined(__unix__)
```
compnerd, line 6472: If we want to use `__unix__` here, perhaps drop the Linux case?

src/queue.c
```diff
@@ -6485,6 +6487,8 @@ _dispatch_runloop_queue_get_handle(dispatch_lane_t dq)
 #elif defined(__linux__)
 	// decode: 0 is a valid fd, so offset by 1 to distinguish from NULL
 	return ((dispatch_runloop_handle_t)(uintptr_t)dq->do_ctxt) - 1;
+#elif defined(__unix__)
+	return ((dispatch_runloop_handle_t)(uintptr_t)dq->do_ctxt);
```
compnerd, line 6491: The off by one difference worries me if this ever gets re-ordered.

src/queue.c
```diff
@@ -6502,13 +6506,21 @@ _dispatch_runloop_queue_set_handle(dispatch_lane_t dq,
 #elif defined(__linux__)
 	// encode: 0 is a valid fd, so offset by 1 to distinguish from NULL
 	dq->do_ctxt = (void *)(uintptr_t)(handle + 1);
+#elif defined(__unix__)
+	dq->do_ctxt = (void *)(uintptr_t)handle;
```
compnerd, line 6510: Similar.

src/queue.c
```diff
@@ -6558,6 +6570,14 @@ _dispatch_runloop_queue_handle_init(void *ctxt)
 		}
 	}
 	handle = fd;
+#elif defined(__unix__)
```
compnerd, line 6573: And here.

src/queue.c
```diff
@@ -6592,6 +6612,11 @@ _dispatch_runloop_queue_handle_dispose(dispatch_lane_t dq)
 #elif defined(__linux__)
 	int rc = close(handle);
 	(void)dispatch_assume_zero(rc);
+#elif defined(__unix__)
```
compnerd, line 6615: Similar

src/queue.c
```diff
@@ -6628,6 +6653,13 @@ _dispatch_runloop_queue_class_poke(dispatch_lane_t dq)
 		result = eventfd_write(handle, 1);
 	} while (result == -1 && errno == EINTR);
 	(void)dispatch_assume_zero(result);
+#elif defined(__unix__)
```
compnerd, line 6656: Similar

src/source.c
```diff
@@ -1402,8 +1402,7 @@ _dispatch_source_debug_attr(dispatch_source_t ds, char* buf, size_t bufsiz)
 			"mask = 0x%x, pending_data = 0x%llx, registered = %d, "
 			"armed = %d, %s%s%s",
 			target && target->dq_label ? target->dq_label : "", target,
-			(unsigned long long)dr->du_ident, dr->du_fflags,
-			(unsigned long long)dr->ds_pending_data,
+			(unsigned long long)dr->du_ident, dr->du_fflags, (unsigned long long)dr->ds_pending_data,
```
compnerd, line 1405: 80-col?

tests/dispatch_workqueue.c
```diff
@@ -0,0 +1,72 @@
+#include <bsdtests.h>
```
compnerd, line 1: Oh this looks like a very nice addition.  This was something that we didn't catch on the original windows port and had to figure it out much later.

3405691582:
Answered some of the questions with some AIs for me to make tweaks and re-test -- I'll answer the others after that.

src/io.c
3405691582: fadvise is just a hint, after all, so it is not an error if it is unavailable.

src/event/event.c
3405691582: `du_ident` on Windows is indeed a `uintptr_t`, as it is on OpenBSD. The timer index here is a `uint32_t`, so as written there is already a truncation occurring. `du_ident` is already initialized in various places from `_dispatch_timer_unote_idx` which returns an `unsigned int`, but here we access `du_ident` directly, which is not guaranteed to be 32-bit wide. 

Casting here just suppresses the error. I don't think there is actually a problem here on looking how du_ident is used in this module; the 32-bit assumption is baked a little hard in the timer code and disentangling that should be handled separately.

src/source.c
3405691582: Rewrapped this function. (Dispatch likes hardtabs, which makes 80-col issues difficult to see.)

src/event/event_kevent.c
3405691582: I renamed it to `DISPATCH_KEVENT_ERSATZ_EVFILT_USER_IDENT`, since this isn't really being used as a mask. The "use" verb doesn't really apply here, because we check `#ifdef EVFILT_USER` -- if there's no EVFILT_USER, the platform doesn't support this filter type and should use the ersatz implementation.

src/queue.c
3405691582: Done.

src/queue.c
3405691582: Seems to work.

cmake/modules/SwiftSupport.cmake
3405691582: Is it Windows that aliases `AMD64` to `x86_64`, or is it something else?

compnerd: Windows does that.  Linux uses x86_64.

3405691582:
FTR, also slightly unwound a little from the futex commit, as it was buggy, so that part of existing code is now unchanged. As mentioned, also performed the MATCHES suggestion based on your comment above.

cmake/modules/SwiftSupport.cmake
```diff
@@ -27,8 +27,12 @@ function(get_swift_host_arch result_var_name)
     set("${result_var_name}" "armv7" PARENT_SCOPE)
   elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "armv7l")
     set("${result_var_name}" "armv7" PARENT_SCOPE)
-  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "AMD64")
-    set("${result_var_name}" "x86_64" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "amd64|AMD64")
+    if(CMAKE_SYSTEM_NAME MATCHES "Windows")
+      set("${result_var_name}" "x86_64" PARENT_SCOPE)
+    else()
+      set("${result_var_name}" "amd64" PARENT_SCOPE)
```
compnerd, line 34: Hmm, I think that Linux also uses x86_64 (see /usr/lib/swift/linux/...).

src/event/event_kevent.c
compnerd: But doesn't that suggest that this is a pointer type (i.e. `uintptr_t`)?

src/event/event_kevent.c
```diff
@@ -2363,6 +2363,12 @@ _dispatch_event_loop_timer_arm(dispatch_timer_heap_t dth, uint32_t tidx,
 		target += range.leeway;
 		range.leeway = 0;
 	}
+#if !NOTE_ABSOLUTE
```
compnerd, line 2402: This won't be defined on all platforms (e.g. Windows).

src/event/event_kevent.c
```diff
@@ -730,6 +745,10 @@ _dispatch_kq_poll(dispatch_wlh_t wlh, dispatch_kevent_t ke, int n,
 		(void)avail;
 		const struct timespec timeout_immediately = {}, *timeout = NULL;
 		if (flags & KEVENT_FLAG_IMMEDIATE) timeout = &timeout_immediately;
+#ifdef EVFILT_USER
+		const struct timespec timeout_1ms = {.tv_sec = 0, .tv_nsec = 1000000}
```
compnerd, line 749: Missing `;`.

src/event/event_kevent.c
```diff
@@ -730,6 +745,10 @@ _dispatch_kq_poll(dispatch_wlh_t wlh, dispatch_kevent_t ke, int n,
 		(void)avail;
 		const struct timespec timeout_immediately = {}, *timeout = NULL;
 		if (flags & KEVENT_FLAG_IMMEDIATE) timeout = &timeout_immediately;
+#ifdef EVFILT_USER
```
compnerd, line 748: This is confusing me.  If `EVFILT_USER` is present, the ersatz user filter shouldn't be used right?  Should this be `#ifndef EVFILT_USER`?

rokhinip (commented):
Firstly I apologize for the delay in reviewing this. 

Secondly, I've done an initial overall pass and highlighted one thing which I think seems like it is in need of changing around EVFILT_USER. Otherwise, most of the other commits looks reasonable to me. Stay tuned for more feedback cause I do want to go back and do a more thorough review of the runloop with pipe commit but wanted to publish initial feedback first.

src/event/event_kevent.c
```diff
@@ -1887,10 +1916,17 @@ _dispatch_event_loop_poke(dispatch_wlh_t wlh, uint64_t dq_state, uint32_t flags)
 {
 	if (wlh == DISPATCH_WLH_MANAGER) {
 		dispatch_kevent_s ke = (dispatch_kevent_s){
+#ifdef EVFILT_USER
 			.ident  = 1,
 			.filter = EVFILT_USER,
 			.fflags = NOTE_TRIGGER,
 			.udata = (dispatch_kevent_udata_t)DISPATCH_WLH_MANAGER,
+#else
+			.ident = DISPATCH_KEVENT_ERSATZ_EVFILT_USER_MASK,
```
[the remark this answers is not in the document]
3405691582: Because not all kqueue implementations have EVFILT_USER. We have to make do on OpenBSD by using a timer with a short timeout to fake it.

src/event/event_kevent.c
3405691582: Right, but Windows is using src/event/event_windows.c. This entire file gets #ifdef'd out on Windows.

cmake/modules/SwiftSupport.cmake
3405691582: Right, but `CMAKE_SYSTEM_PROCESSOR` is `x86_64` on Linux normally, isn't it?

src/event/event_kevent.c
3405691582: Yes, it's a `void *`. The code deliberately passes the pointer to `_dispatch_bug_kevent_client` but the argument is integral (as is `uintptr_t`). Removing the cast gives an error:
```
error: incompatible pointer to integer conversion passing 'void *' to parameter of type 'uint64_t' (aka 'unsigned long long')
```

src/event/event_kevent.c
3405691582: You're right. I think perhaps I added something like this while debugging, didn't realize it was basically inoperative while testing, and didn't remove it. I've verified things work without this particular fragment.

src/event/event_kevent.c
3405691582: (Moot.)

src/event/event_kevent.c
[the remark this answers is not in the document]
rokhinip: I understand but I'm saying you don't even need kqueue for the thing that this is trying to use EVFILT_USER for. On Darwin, we decided to use the EVFILT_USER to bring up the manager thread through our kernel-integrated workqueue subsystem. 

On linux and OpenBSD, you don't need that - all the caller of this function with `DISPATCH_WLH_MANAGER` trying to do, is wake up the manager queue so let's just do that and enqueue it onto the manager root queue and have the workqueue bring up threads for it. Why are we bothering with kevent and timers and all that additional kernel overhead?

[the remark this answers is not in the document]
3405691582: I'm not sure I understand. A lot of Dispatch internals are undocumented and difficult to understand, especially in the event backend code like this, so I may be confused at your suggestion. 

All of the other event backends (Windows, Linux's epoll) appear to use platform-specific functionality to handle other signal inputs in the event loop (fd read/write, signals, timers), and presumably in `_dispatch_event_loop_poke` this is done with the same platform-specific mechanism so that all signals are ganged together in the one event loop. And the kevent backend that Darwin and OpenBSD use is no different.

Are you saying there is some way that _all_ of these event backends can avoid having to use the platform-specific functionality here? If so, knowing how to improve this is less than obvious and if that is the path you would want to see me go down, I would need more guidance on that. 

That being said, I could try and stare at this code a little more to try and figure out how to implement the suggestion you made, but from a perspective of someone trying to get Dispatch to build on a new platform without breaking an existing one, as well as having those changes work and be reliable, the smallest change to an existing codebase is likely the safest one.

This is why just providing a compatible alternative to EVFILT_USER was done, because it seems like the simplest, cleanest way to get this backend to work on this platform. 

(One alternative I briefly considered was to create an entirely new backend based solely on portable POSIXish primitives, but knowing how to properly implement the semantics of that backend API interface was baffling enough as it is, and ultimately it made much more sense to tweak the kevent backend to be a little more portable.)

3405691582:
While waiting for additional reviews, I made the following changes in this pr from cross-compiling and building this change on Linux:

* Fixed a minor issue where the new `dispatch_workqueue` would work on OpenBSD but not Linux by making the test context a global.

* Fixed an issue where `dispatch_select` would crash because it would run into stack size rlimits on OpenBSD but not Linux by moving the buffers in this test to the heap.

* Fixed a cast to an integral type to match the format specifier, rather than a platform-specific type (like `time_t`)

Additionally, I ran the tests on Linux forcing the queue runloop to use the pipe2 implementation proposed in this pr rather than epoll; it seemed to work fine.

finagolfin:
@swift-ci please test

finagolfin:
@swift-ci please test

src/event/event_kevent.c
3405691582: Okay, I've just changed this to `!defined` instead.

src/io.c
3405691582: Okay, I've just converted this back to a platform conditional and reinstated the `#error`.

src/event/event_kevent.c
3405691582: Actually -- that is wrong. NOTE_ABSOLUTE is defined in event_kevent.h, and is set to 0 if it is undefined. I'm going to go back to `!NOTE_ABSOLUTE`, since the change caused problems.

3405691582:
@swift-ci please test.

3405691582:
@compnerd what do you think is necessary and remaining to land this (especially now that #861 is in the pipeline as well and using some of this work), or what is the ongoing strategy going to be for platform-specific changes to Dispatch?

finagolfin:
@al45tair, as PSG chair, maybe you can help push this four year-old pull forward, just need somebody with the right libdispatch background to review: maybe you can find someone.

al45tair:
> @al45tair, as PSG chair, maybe you can help push this four year-old pull forward, just need somebody with the right libdispatch background to review: maybe you can find someone.

I have an open PR against corelibs-libdispatch also; I've been promised someone will look at some of these soon…

src/event/event_kevent.c
```diff
@@ -528,11 +534,17 @@ _dispatch_kevent_merge_muxed(dispatch_kevent_t ke)
 	}
 }
 
+#define DISPATCH_KEVENT_ERSATZ_EVFILT_USER_IDENT 0xfffffe00
```
rokhinip, line 537: This is unfortunate to have to use timers with this ident to have to signal the manager thread on this platform - should we have a comment to explain the purpose here?

Also, can you clarify what ERSATZ is?

src/queue.c
```diff
@@ -6563,6 +6573,14 @@ _dispatch_runloop_queue_handle_init(void *ctxt)
 		}
 	}
 	handle = fd;
+#elif defined(__unix__) && !defined(__linux__)
+	int fds[2];
```
rokhinip, line 6577: Do we not have eventfd or an equivalent on OpenBSD? What about https://man.openbsd.org/event.3?

rokhinip (approved):
Overall, this looks generally reasonable and is mostly plumbing that AFAICT, is mainly limited to the OpenBSD stuff. I am not familiar with the intricacies of OpenBSD as a platform and what it provides so I am deferring to the author's expertise there.

src/queue.c
3405691582: This is an API similar to libevent, and behaves differently from eventfd.

Here, I thought we could use somewhat more standard POSIX-ish API surface that would be more portable (especially when `defined(__unix__)`.)

A more platform-specific API could come later.

akmorrison (approved):

src/event/event_kevent.c
akmorrison: Will bad things happen if a `DISPATCH_SOURCE_TYPE_INTERVAL` timer is created with this magic value as the interval? I think that's the only EVFILT_TIMER source that uses something other than a clockid as the ident.

3405691582: I don't think so. As I understand it/unless I'm missing something, the only other kevent with EVFILT_TIMER here has timer index masked with `DISPATCH_KEVENT_TIMEOUT_IDENT_MASK` in the ident (`_dispatch_event_loop_timer_program`).   `DISPATCH_KEVENT_ERSATZ_EVFILT_USER_IDENT` shouldn't conflict with `DISPATCH_KEVENT_TIMEOUT_IDENT_MASK`.

(maybe DISPATCH_KEVENT_ERSATZ_EVFILT_USER_IDENT might be best fully sign extended here. I'll try that while I'm adding the comment mentioned above.)

3405691582:
Comment and full sign extension on `DISPATCH_KEVENT_ERSATZ_EVFILT_USER_IDENT` added.

@swift-ci please test.

3405691582:
@swift-ci please test.

3405691582:
Please merge on my behalf if there are no other concerns, thank you!

jakepetroules:
@3405691582 Why was makeProcessSource disabled? I thought OpenBSD supports kevent + EVFILT_PROC.

3405691582:
> @3405691582 Why was makeProcessSource disabled? I thought OpenBSD supports kevent + EVFILT_PROC.

I think I've lost context on the specifics. I'll revisit and see if this can be reenabled.

PR #561, opened 2021-07-12
Fix FreeBSD support

drexin:


drexin:
@swift-ci test

3405691582:
FYI: some of this is already captured in #559. You may want to review that, and then perhaps some of this can be streamlined.

src/init.c
```diff
@@ -1048,7 +1048,7 @@ _dispatch_bug_kevent_vanished(dispatch_unote_t du)
 			"{ %p[%s], ident: %" PRIdPTR " / 0x%" PRIxPTR ", handler: %p }",
 			dux_type(du._du)->dst_kind, dou._dq,
 			dou._dq->dq_label ? dou._dq->dq_label : "<unknown>",
-			du._du->du_ident, du._du->du_ident, func);
+			(intptr_t)du._du->du_ident, (intptr_t)du._du->du_ident, func);
```
3405691582, line 1051: If we're going to cast here, we might as well choose `uintmax_t` (in the alternative, I had suggested on the other pr defining some custom format strings because the type of `dispatch_unote_ident_t` is platform defined, but casting is a reasonable alternative here.)

src/internal.h
```diff
@@ -36,7 +36,7 @@
 #define __DISPATCH_BUILDING_DISPATCH__
 #define __DISPATCH_INDIRECT__
 
-#ifdef __APPLE__
+#ifdef __APPLE__DISPATCH_USE_KEVENT_WORKLOOP
```
3405691582, line 39: `s/__APPLE__//`?

src/shims/lock.c
```diff
@@ -56,6 +56,21 @@ _dispatch_thread_switch(dispatch_lock value, dispatch_lock_options_t flags,
 #endif
 #endif
 
+#if defined(__FreeBSD__)
```
3405691582, line 59: This could be merged above and the `#ifdef`put in the function body.

src/swift/Source.swift
```diff
@@ -174,7 +174,7 @@ extension DispatchSource {
 	}
 #endif
 
-#if !os(Linux) && !os(Android) && !os(Windows)
+#if !os(Linux) && !os(Android) && !os(Windows) && !os(FreeBSD)
```
3405691582, line 177: ...and then we're going to have to eventually `&& !os(OpenBSD)` -- should we just bite the bullet and rewrite this `os(macOS)`? (throughout)

lin72h:
[libdispatch port](https://www.freshports.org/devel/libdispatch) is back online Thanks for @fluffkhv

etcwilde:
I believe that this PR has been superseded by https://github.com/swiftlang/swift-corelibs-libdispatch/pull/861.
@drexin, are there still changes needed from this PR?

drexin:
> I believe that this PR has been superseded by #861. @drexin, are there still changes needed from this PR?

I don‘t know, it‘s been too long since I put this up.

PR #564, opened 2021-08-22
Optimize _dispatch_alloc_maybe_madvise_page

AtariDreams:
There is no reason to have last_locked be signed. Also, redundant checks were optimized away near the end of the function.

### DIFF
--- a/src/allocator.c
+++ b/src/allocator.c
@@ -534,23 +534,22 @@ _dispatch_alloc_maybe_madvise_page(dispatch_continuation_t c)
 	volatile bitmap_t *page_bitmaps;
 	get_maps_and_indices_for_continuation((dispatch_continuation_t)page, NULL,
 			NULL, (bitmap_t **)&page_bitmaps, NULL);
-	unsigned int i;
-	for (i = 0; i < BITMAPS_PER_PAGE; i++) {
-		if (page_bitmaps[i] != 0) {
+	unsigned int last_locked;
+	for (last_locked = BITMAPS_PER_PAGE; last_locked; last_locked--) {
+		if (page_bitmaps[last_locked] != 0) {
 			return;
 		}
 	}
 	// They are all unallocated, so we could madvise the page. Try to
 	// take ownership of them all.
-	int last_locked = 0;
-	do {
+	for (; last_locked < BITMAPS_PER_PAGE; last_locked++) {
 		if (!os_atomic_cmpxchg(&page_bitmaps[last_locked], BITMAP_C(0),
 				BITMAP_ALL_ONES, relaxed)) {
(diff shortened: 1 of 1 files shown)


AtariDreams:
@swift-ci please test

AtariDreams:
cc @compnerd

AtariDreams:
cc @drexin

etcwilde (approved):
These changes look fine to me.
I would still wait for another reviewer's response as I'm not super familiar with this codebase.

rokhinip:
The older code reads cleaner to me actually - this change seems to add more branches and code repetition

AtariDreams:
Yeah but the check is redundant knowing that when it is greater than 1, it is not 0. I could refactor it though

AtariDreams:
Refactored. Hopefully it isn't as bad now

etcwilde:
Quickly looking at the optimized output, the instruction sequences look the same on first glance for all three cases. I agree with @rokhinip that the original was the easiest to read so lets stay with that.
https://godbolt.org/z/e81fh5G37

I'm fine with the change to using the unsigned int though.

AtariDreams:
All fixed!

AtariDreams:
> Quickly looking at the optimized output, the instruction sequences look the same on first glance for all three cases. I agree with @rokhinip that the original was the easiest to read so lets stay with that.
> https://godbolt.org/z/e81fh5G37
> 
> I'm fine with the change to using the unsigned int though.

Addressed!

AtariDreams:
> Quickly looking at the optimized output, the instruction sequences look the same on first glance for all three cases. I agree with @rokhinip that the original was the easiest to read so lets stay with that.
> https://godbolt.org/z/e81fh5G37
> 
> I'm fine with the change to using the unsigned int though.

Actually, dothing3 is the odd one out in terms of the generated assembly, being smaller and shorter

etcwilde:
So, number 3 has fewer instructions, yes, but also involves more control flow from a readability perspective.
I haven't run any perf numbers, but statically, it looks like a difference of 4 instructions between the two.
Without a profile demonstrating how much time is saved by these, I would push for making the code more readable over reducing instructions.

AtariDreams:
Alright. I removed the change

AtariDreams:
@swift-ci please test

rokhinip:
What is the problem you're trying to address here @AtariDreams? Refactors are well and good but what purpose does this serve?

AtariDreams:
> What is the problem you're trying to address here @AtariDreams? Refactors are well and good but what purpose does this serve?

better readability and code generation

rokhinip:
> better readability and code generation

I don't see how this change does that. I think this change adds more churn than benefit. I'd like for us to drop this change please. Thanks.

PR #565, opened 2021-08-23
Remove unused switch table in init.c

AtariDreams:
I don't know what was meant to be here, but a switch statement with all outcomes being a simple break isn't needed, so it would be beneficial for readability to just remove it.


AtariDreams:
@etcwilde Hopefully this doesn't cause any issues. I haven't seen it fail any tests, but as with code like this I want to be careful and make sure something else wasn't intended.

AtariDreams:
cc @rokhinip

AtariDreams:
cc @rokhinip @compnerd

rokhinip:
This switch statement is intentional in there and is meant to catch us if we fail to handle all possible cases in the project.

PR #566, opened 2021-08-23
last_locked should be unsigned

AtariDreams:
That is the only change, since every other index is unsigned, but it is signed in the one case where it doesn't need to be.

The rest of the function shall be left alone.


AtariDreams:
cc @etcwilde

AtariDreams:
cc: @compnerd

rokhinip:
While this change is not wrong, it really isn't doing anything useful here either. I'm closing this PR.

Please only make PRs which are meaningfully fixing something that is broken.

PR #567, opened 2021-08-23
Change dqai_concurrent and dqai_inactive to booleans

AtariDreams:
Since this is how they are used, it makes sense to define them as such


AtariDreams:
cc: @rokhinip

rokhinip:
While this change is not wrong, it really isn't doing anything useful here either. I'm closing this PR. 

Please only make PRs which are meaningfully fixing something that is broken.

PR #568, opened 2021-08-26
[android] Update headers that were split in NDK 23

buttaface:
NDK 23 split `stdatomic.h` up and that causes problems when invoked from C++ files, so use the new `bits/stdatomic.h` instead as a workaround, as seen in the test for aosp-mirror/platform_bionic@76e2b15.

[Building libdispatch for Android with the new NDK causes linkage issues in the headers](https://github.com/buttaface/swift-android-sdk/runs/3402803617?check_suite_focus=true#step:7:889), so I used this same workaround that can be seen in the C++ test for that Bionic commit that split the header up and libdispatch builds for Android again.

I don't know if there's a better way to fix these linkage issues, so let me know if there is.

### DIFF
--- a/src/shims/atomic.h
+++ b/src/shims/atomic.h
@@ -35,7 +35,11 @@
 #if defined(__cplusplus)
 #define _Bool bool
 #endif
+#if defined(__ANDROID__) && __NDK_MAJOR__ >= 23
+#include <bits/stdatomic.h>
+#else
 #include <stdatomic.h>
+#endif
 
 #define memory_order_ordered    memory_order_seq_cst
 #define memory_order_dependency memory_order_acquire


drodriguez:
Have you tried...

```
#if defined(__cplusplus)
#include <atomic>
#else
#include <stdatomic.h>
#endif
```

One should not access the `bits/` headers outside the standard headers. They are private details of the implementation.

buttaface:
No, I'm not familiar with these linkage issues when mixing C headers with C++, will try your way instead and let you know.

buttaface:
Just tried it, doesn't work either. If you look at the error trace from my link above, the problem is that `src/internal.h` starts a `__BEGIN_DECLS` block where it includes these shims. The issue is that is an `extern C` block that's invoked inside a C++ file, so `stadomic.h` gives it the C++ `<atomic>` too, which then complains about the enclosing `extern C` block. I don't know why this doesn't hit on other platforms though.

This problem remains with your proposed fix, whereas mine makes sure only the C declarations are called, even if it calls an internal header to do so. If you have another fix in mind, let me know.

buttaface:
@drexin, any input on this? Maybe you've dealt with these mixed linkage issues before, I haven't.

triplef:
I think this also applies to previous NDK versions. We’ve been [using](https://github.com/gnustep/tools-android/blob/master/patches/libdispatch-fix-atomic.patch) the following to work around these for a while:

```c
#if defined(ANDROID) && __has_include(<bits/stdatomic.h>)
#include <bits/stdatomic.h>
#else
#include <stdatomic.h>
#endif
```

See also the following related discussions in the NDK project for more background:

- https://github.com/android/ndk/issues/1177
- https://github.com/android/ndk/issues/1178

buttaface:
Thanks for all the links. [According to Dan Albert last year, what libdispatch does in `src/internal.h` is against the C++ standard](https://github.com/android/ndk/issues/1177#issuecomment-593662065), so this pull is just a workaround for that:

"I did get an authoritative answer on this from multiple members of the C++ standard committee, and using `#include` within `extern "C"` is disallowed by the standard.

http://eel.is/c++draft/using.headers#3

> A translation unit shall include a header only outside of any declaration or definition and, in the case of a module unit, only in its global-module-fragment, and shall include the header or import the corresponding header unit lexically before the first reference in that translation unit to any of the entities declared in that header. No diagnostic is required.

An `extern "C"` block is a declaration, so it's disallowed."

@ktopley-apple, is there a better change we can make to `internal.h` to make it standard-compliant?

shrukul:
@buttaface - I'm facing this issue in NDK r22, and can confirm that including `<bits/stdatomic.h>` fixes the issue!

Hence, wanted to ask: Any particular reason for enabling this fix only for `__NDK_MAJOR__ >= 23` ? Given that this was broken in r21b, can we enable this fix starting from r21b? (Let me know if I'm mistaken/incorrect)

buttaface:
@shrukul, the NDK issues @triplef linked above note that the Bionic header change that triggered this was reverted in later versions of NDK 21, ie the last version 21e doesn't have it. While NDK 22 did, it wasn't an LTS NDK and has now been obsoleted by NDK 23, so I chose the first LTS NDK that made this change, 23.

Anyway, hopefully we can get this fixed properly in `internal.h` without regard to NDK versions.

@etcwilde, any input?

etcwilde:
> @shrukul, the NDK issues @triplef linked above note that the Bionic header change that triggered this was reverted in later versions of NDK 21, ie the last version 21e doesn't have it. While NDK 22 did, it wasn't an LTS NDK and has now been obsoleted by NDK 23, so I chose the first LTS NDK that made this change, 23.
> 
> Anyway, hopefully we can get this fixed properly in `internal.h` without regard to NDK versions.
> 
> @etcwilde, any input?

I'm not familiar with android development, so I don't have an opinion on a version, so long as it work.
@rokhinip is probably the best person to page with options in this repository.

buttaface:
@etcwilde, I was more hoping for an opinion on [this comment, which has nothing to do with Android](https://github.com/apple/swift-corelibs-libdispatch/pull/568#issuecomment-913758466), but is about how one of this project's headers is doing something non-standard, that happens to break for Android. If we fix it there, we might fix this issue without reference to Android or its versions.

I'm unfamiliar with these arcane C/C++ issues, so I was hoping you might have some input on that.

etcwilde:
> @etcwilde, I was more hoping for an opinion on [this comment, which has nothing to do with Android](https://github.com/apple/swift-corelibs-libdispatch/pull/568#issuecomment-913758466), but is about how one of this project's headers is doing something non-standard, that happens to break for Android. If we fix it there, we might fix this issue without reference to Android or its versions.
> 
> I'm unfamiliar with these arcane C/C++ issues, so I was hoping you might have some input on that.


Ah, sorry, I misunderstood. The answer is a little bit round-a-bout. So as per the C++11 spec (7.6.1.2, Table 14), the C++11 standard library must vend an `atomic` library, so you should be guaranteed that `#include <atomic>` will work. If `__STDC_NO_ATOMICS__` is defined, then the header `stdatomic.h` is not available. Apparently this is the case on Android.
If this file were just C++, it would be fine to replace the include of `stdatomic.h` with `atomic`, fix up the differences, and move on.

The part I don't know is how standards compliant the Android development kits are, or which version of developer kit contains which version(s) of the standard.

This [proposal](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p0943r6.html), which targets C++23, actually comments on this issue.  Unfortunately, right now the only portable way is to include `atomic` from C++ and have a bunch of `using` statements to strip off the `std` namespace in the case of C++.

e.g.

```c++
#ifdef __cplusplus
  #include <atomic>
  using std::atomic_int;
  using std::memory_order;
  using std::memory_order_acquire;
  ...
#else /* not __cplusplus */
  #include <stdatomic.h>
#endif /* __cplusplus */
```

buttaface:
@etcwilde, I was more hoping to hear about the issue causing this, ie that [`src/internal.h` includes several headers inside its `__BEGIN_DECLS` block](https://github.com/apple/swift-corelibs-libdispatch/blob/main/src/internal.h#L323), which Dan notes is against the standard. It so happens that Android's `stdatomic.h` is the one erroring because of that, but others could also later. Is there something we could do to fix the root cause in `src/internal.h` instead?

I just noticed [another `#include` changing its linkage inside the block](https://github.com/apple/swift-corelibs-libdispatch/blob/main/src/internal.h#L821) so I tried that with `shims.h` there too, but that caused other errors.

Any ideas for other changes to `src/internal.h` that could work around this issue there?

buttaface:
> There's a horrible thing that comes to mind - and I'd prefer to avoid it if at all possible - we could wrap the header via include_next and convert things around either with macros or other things. If we can work out another workaround, I think that would be preferable. But I think that we have some more potent tools available if absolutely required.

I don't think that would work either. Let me state the problem as briefly as possible:

Prior to NDK 22, `stdatomic.h` in Android's Bionic libc provided both C++ and C declarations, but for whatever reason, the C++ declarations were disabled. That worked fine with `src/internal.h` as it only could use C declarations, even when compiling C++.

However, once NDK 22 enabled those C++ declarations, it broke compiling C++ files in this codebase, because of the incorrect usage of `extern C` in `src/internal.h`. Given that underlying problem, I don't see how `include_next` or anything other than including `bits/stdatomic.h`, which is where the C declarations were moved in NDK 22, would work around the improper usage in `src/internal.h`.

I'd rather fix the underlying problem in `src/internal.h` and not try to work around it, but I don't know a way to do that.

> Could you share what the "other errors" were with switching linkage to c++?

Some other C++ definition issues, I'll rerun that and paste the error later this weekend.

compnerd:
Changing the linkage to C++ before including the header as a workaround for android seems reasonable to accommodate the non-conforming behaviour on android.  I'd  say that once we have the errors we can likely understand the situation better.

buttaface:
> the non-conforming behaviour on android

I don't think there was anything to "conform" to for `stdatomic.h`, but [Dan notes that C++ is moving to precisely this solution](https://github.com/android/ndk/issues/1178#issuecomment-580471798). So whenever other C++ implementations catch up to Bionic, they will break on libdispatch too.

buttaface:
Btw, [the Android CI has been broken for the last month with this issue](https://ci-external.swift.org/job/oss-swift-RA-linux-ubuntu-16.04-android-arm64/lastBuild/console), because the LLVM rebranch last month switched from linking Android against libgcc to compiler-rt, apple/llvm-project@a478b0a, so Daniel switched the Android NDK on the CI to the latest version 23 that uses such a recent clang too, but has a sysroot that also made this `stdatomic.h` change.

Since this pull only affects Android and is a workaround for another header that does not conform to the C++ standard in this codebase, I think we should just get this in as a workaround for Android, since nobody has come up with another solution that works.

drodriguez:
@swift-ci please test

drodriguez:
@shahmishal can we merge this one to unblock the Android CI before the Thanksgiving break? Thanks!

PR #569, opened 2021-08-30
shims: repair the build on Windows i686

compnerd:
The Windows i686 builds expect this function to use the stdcall calling
convention.  The implementation correctly decorates the function to do
so.  However, the declaration did not match, which caused a build
failure on windows i686.


compnerd:
@swift-ci please test

compnerd:
CC: @rokhinip

PR #571, opened 2021-09-11
shims: correct the check for newer SLPI enumerators

compnerd:
The previous check did not work as the enumertors are not defines but
rather a proper enumerator.  This adds a build time check for the
support and conditionalises the compilation.


compnerd:
CC: @rokhinip 
Unfortunately, the CI hosts are updated outside of my control by Microsoft.  They bumped the SDK, so we need this for the release.

compnerd:
@swift-ci please test

rokhinip:
Sounds good.

PR #575, opened 2021-09-20
Determine offset AFTER kern_return_t check, not before

AtariDreams:
Not only for minor performance but also for better readability, and most importantly, so that we don't read the dal_first_allocation_offset field in the event kr is not KERN_RETURN_SUCCESS

### DIFF
--- a/src/allocator.c
+++ b/src/allocator.c
@@ -542,15 +542,15 @@ _dispatch_alloc_maybe_madvise_page(dispatch_continuation_t c)
 	}
 	// They are all unallocated, so we could madvise the page. Try to
 	// take ownership of them all.
-	int last_locked = 0;
-	do {
+	unsigned int last_locked;
+	for (last_locked = 0; lastlocked < BITMAPS_PER_PAGE; i++) {
 		if (!os_atomic_cmpxchg(&page_bitmaps[last_locked], BITMAP_C(0),
 				BITMAP_ALL_ONES, relaxed)) {
 			// We didn't get one; since there is a cont allocated in
 			// the page, we can't madvise. Give up and unlock all.
 			goto unlock;
 		}
-	} while (++last_locked < (signed)BITMAPS_PER_PAGE);
+	}
 #if DISPATCH_DEBUG
 	//fprintf(stderr, "%s: madvised page %p for cont %p (next = %p), "
 	//		"[%u+1]=%u bitmaps at %p\n", __func__, page, c, c->do_next,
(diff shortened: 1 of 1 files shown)


AtariDreams:
@compnerd thoughts?

AtariDreams:
@shahmishal thoughts?

shahmishal:
cc: @rokhinip

src/allocator.c
```diff
@@ -542,15 +542,15 @@ _dispatch_alloc_maybe_madvise_page(dispatch_continuation_t c)
 	}
 	// They are all unallocated, so we could madvise the page. Try to
 	// take ownership of them all.
-	int last_locked = 0;
-	do {
+	unsigned int last_locked;
```
rokhinip, line 545: This change here was already turned down as part of PR https://github.com/apple/swift-corelibs-libdispatch/pull/564 so it really shouldn't be reintroduced here.

src/allocator.c
```diff
@@ -654,27 +654,39 @@ _dispatch_allocator_enumerate(task_t remote_task,
...
-		//        the bitmap.
-		kern_return_t kr = reader(remote_task, zone_address, heap_size, &heap);
-		size_t offs = remote_dal->dal_first_allocation_offset;
-		if (kr) return kr;
-		while (offs < heap_size) {
-			void *isa = *(void **)(heap + offs + dc_flags_offset);
-			if (isa && isa != remote_dal->dal_deferred_free_isa) {
-				recorder(zone_address + offs, heap + offs, dc_size, &stop);
-				if (stop) return KERN_SUCCESS;
+	if (zone_address)
```
rokhinip, line 657: This seems like it just changes a while statement into an if branch then a do-while. What is the benefit of this??

rokhinip (commented):
I fail to see what problem this PR is trying to solve.

src/allocator.c
AtariDreams: To avoid the assignments if not needed at all.

src/allocator.c
AtariDreams: Well for starters, it's the only odd loop out of the others that enumerate the bitmap, and the only one that casts bitmaps to page to signed for no reason. Or at least, whatever reason it was in the past, isn't relevant now. It's making a compromise for nothing. It should be a simple, canonical, similar in style, for loop.

AtariDreams:
> I fail to see what problem this PR is trying to solve.

Put shortly, it's doing some operations out of order

rokhinip:
I'd like to avoid making unnecessary changes unless its really solves a problem - whether it's a logic or a performance problem. I'm not convinced that either of those are at play here. 

A bigger refactor (NFC) to improve readability/simplify a part of the codebase is also valuable. But I don't see a readability problem here either and don't think this PR is improving that.

Trivial changes like this really adds more risk and churn for maintainers without a clear value-add to the project so in the future, it's appreciated if the changes fix a clear problem.

PR #576, opened 2021-09-22
Fix unused variable errors

etcwilde:
There are build errors due to unused variables in two of the tests.
The build system builds these with `-Werror` so instead of a warning, it is a hard failure.

This is affecting the swift rebranch work, which has a newer clang that emits warnings in more cases, which is why we are seeing this now.

Failing build log:
https://ci.swift.org/view/Swift%20rebranch/job/oss-swift-rebranch-incremental-RA-linux-ubuntu-18_04/23/console

### DIFF
--- a/tests/bsdtestharness.c
+++ b/tests/bsdtestharness.c
@@ -72,13 +72,6 @@ main(int argc, char *argv[])
 	assert(res == 0);
 #endif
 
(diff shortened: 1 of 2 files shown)


etcwilde:
@swift-ci please test

PR #578, opened 2021-10-01
[Windows] Prevent socket callback from falsely detecting a connection close

lxbndr:
This fixes a hard-to-reproduce issue we did observe in our live product (powered by Swift on Windows). The application communicates with the backend a lot, and some network requests were randomly timing out. All attempts to reproduce that behavior synthetically failed, but in the end we got some low but stable repro rate and had a chance to debug this.

It appeared that sometimes Dispatch catches `FD_READ` event, but `ioctlsocket` says that we have nothing to read at the moment of event processing. That is perfectly ok, non-zero result is not guaranteed there, as the reading could occur asynchronously.

But 0 is also used as a special value in `_dispatch_event_merge` as a marker of a closed socket. Such sockets are exclude from any further handling, making the connection hang.

The suggested solution avoids read/write processing with 0 available bytes, except for `FD_CLOSE` event, when we really need to poke all handlers to actually shut down the socket. The `FD_READ` event with 0 bytes becomes a no-op and doesn't do anything.

We applied this fix locally more than a year ago and since then the connection behavior seems to be stable.

### DIFF
--- a/src/event/event_windows.c
+++ b/src/event/event_windows.c
@@ -364,7 +364,11 @@ _dispatch_socket_callback(PTP_CALLBACK_INSTANCE inst, void *context,
 		} else if (lNetworkEvents & FD_READ) {
 			ioctlsocket(sock, FIONREAD, &dwBytesAvailable);
 		}
-		if (lNetworkEvents & FD_READ) {
+		// ioctlsocket could legally return 0 and that doesn't mean the socket is closed.
+		// Here and below we have to double-check dwBytesAvailable and make sure we don't 
+		// wake up read/write handlers with 0 value for live (not closed) socket, because
+		// that would force handler to shut down.
+		if (((lNetworkEvents & FD_READ) && dwBytesAvailable > 0) || (lNetworkEvents & FD_CLOSE)) {
 			_dispatch_muxnote_retain(dmn);
 			if (!PostQueuedCompletionStatus(hPort, dwBytesAvailable,
 					(ULONG_PTR)DISPATCH_PORT_SOCKET_READ, (LPOVERLAPPED)dmn)) {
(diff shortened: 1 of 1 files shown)


lxbndr:
CC: @compnerd

src/event/event_windows.c
```diff
@@ -364,15 +364,19 @@ _dispatch_socket_callback(PTP_CALLBACK_INSTANCE inst, void *context,
...
+		if (((lNetworkEvents & FD_READ) && dwBytesAvailable > 0) || (lNetworkEvents & FD_CLOSE)) {
 			_dispatch_muxnote_retain(dmn);
 			if (!PostQueuedCompletionStatus(hPort, dwBytesAvailable,
 					(ULONG_PTR)DISPATCH_PORT_SOCKET_READ, (LPOVERLAPPED)dmn)) {
 				DISPATCH_INTERNAL_CRASH(GetLastError(),
 						"PostQueuedCompletionStatus");
 			}
 		}
-		if (lNetworkEvents & FD_WRITE) {
+		if (((lNetworkEvents & FD_WRITE) && dwBytesAvailable > 0) || (lNetworkEvents & FD_CLOSE)) {
```
compnerd, line 379: This doesn't make sense - `dwBytesAavailable` is guaranteed `1` if *only* a write event occurred.  This condition doesn't seem correct.  It should be either a write or a close event occurred I believe.

src/event/event_windows.c
```diff
@@ -364,15 +364,19 @@ _dispatch_socket_callback(PTP_CALLBACK_INSTANCE inst, void *context,
 		} else if (lNetworkEvents & FD_READ) {
 			ioctlsocket(sock, FIONREAD, &dwBytesAvailable);
 		}
-		if (lNetworkEvents & FD_READ) {
+		// ioctlsocket could legally return 0 and that doesn't mean the socket is closed.
+		// Here and below we have to double-check dwBytesAvailable and make sure we don't 
+		// wake up read/write handlers with 0 value for live (not closed) socket, because
+		// that would force handler to shut down.
+		if (((lNetworkEvents & FD_READ) && dwBytesAvailable > 0) || (lNetworkEvents & FD_CLOSE)) {
```
compnerd, line 371: This seems plausible to occur.  But `FD_READ` is already set when `FD_CLOSE` is set, so this is rather confusing.

lxbndr:
Closing in favor of #581

PR #579, opened 2021-10-03
Replace bzero with memset

AtariDreams:
bzero is deprecated ever since 2008, and having to define it for Windows when we can use memset directly is also unnecessary. Let us replace it with memset instead.


AtariDreams:
cc: @rokhinip

AtariDreams:
Any updates?

rokhinip:
My feedback given above still stands. I don't see much value add from this change.

PR #581, opened 2021-10-11
event: handle socket events more precisely

compnerd:
This adjusts the socket event to correct a potential spurious wakeup
which can incorrectly mark the socket as closed.

When working with a `SOCK_STREAM` socket, `FIONREAD` returns the total
amount of data that can be read in a single receive operation.
Normally, this is the same amount of data queued on the socket, but this
is not guaranteed as a data-stream is byte-oriented.  Additionally, a
read on a separate thread may have drained the data between the event
trigger and the ioctl to read the available bytes.  Only wake up the
handler if there are bytes to read.

In order to differentiate between the trigger for a close event and a
read-event, we explicitly check the close and the read or write event.
This allows waking the handlers one last time after a socket is closed.


compnerd:
CC: @rokhinip

PR #447, opened 2019-02-22
Merge darwin/libdispatch-1121 to master

ktopley-apple:
Merge tag darwin/libdispatch-1121 on darwin/trunk.

This includes the latest Apple source drop libdispatch-1121 (up from libdispatch-913.1.4 in #447), on top of darwin/darwin-012.

PR #583, opened 2021-11-06
Merge darwin/libdispatch-1324.41.2 to main

rokhinip:
Merge tag darwin/libdispatch-1324.41.2 on darwin/trunk.

This includes the latest Apple source drop libdispatch-1324.41.2 (up from libdispatch-1121) on top of darwin/libdispatch-1121 (merged in #447)


rokhinip:
@swift-ci please test

CMakeLists.txt
```diff
@@ -40,6 +40,20 @@ int main(int argc, char *argv[]) {
   if(DISPATCH_HAVE_EXTENDED_SLPI_20348)
     add_compile_definitions(DISPATCH_HAVE_EXTENDED_SLPI_20348)
   endif()
+
```
das, line 43: not sure what's up with the github diff viewer but this diff isn't real, this file is identical between the two branches

os/workgroup_base.h
```diff
@@ -0,0 +1,78 @@
...
+#endif
+
+#include <sys/types.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <string.h>
+#include <stdlib.h>
+
+#include <mach/port.h>
```
3405691582, line 15: Unprotected `mach/port.h` is a portability break. `Availability.h` appears to be missing, and `os/base.h` doesn't exist. `os/object.h` and `os/clock.h` does exist, but `os/clock.h` also refers to `os/base.h`, so something is missing here.

3405691582: Seems like `os/base.h` should switch on `os/generic_*_base.h`. `os/generic_*_basae.h` needs to also #define away SPI_UNAVAILABLE...

3405691582:
There's other Mach-specific stuff that needs to be untangled too that I haven't felt up to untangling right now, but that'll trip up on supported non-Mach platforms too...

compnerd (commented):
I'll try to run through a build of this in a the next day or two.

compnerd:
There is a lot of use of mach_port_t, but I don't yet understand the use of them here.  I wonder if we need to provide equivalences for those APIs on Linux/Android/Windows.

MadCoder:
> There is a lot of use of mach_port_t, but I don't yet understand the use of them here. I wonder if we need to provide equivalences for those APIs on Linux/Android/Windows.

the workgroup stuff is not relevant for non Darwin platforms and should be entirely hidden behind `HAVE_MACH`

compnerd:
Thanks @MadCoder!  That actually was very helpful.  I do have a handful of patches to upload that improves stuff but is insufficient on its own.  What would you prefer @rokhinip @MadCoder @das?  I can push to the same branch, or create a separate branch that adds them.  They aren't particularly freestanding, so I cannot just create PRs for main to address these items.

das:
thanks @compnerd for looking at this!
It would be very useful to have a separate merge commit for the followup changes to this PR that we can cleanly split out, so I think a separate branch initially created off-of this one would be ideal so that we can either merge it to this branch or just merge the two to main one after the other (after a rebase)

compnerd:
@das sounds good, I'm going to be pushing a series of changes to the same branch and have created a PR for ease of review.  Expect a large number of changes, since I want to keep them tiny and targeted so we can deal with the individual issues.  I've CC'ed the people here on #584.

PR #585, opened 2021-11-15
event: support workqueue monitoring on Windows

compnerd:
This enables workqueue monitoring on Windows to allow for overcommitting
threads if necessary.  We use a combination of WCT (wait chain tracing)
- to determine if there is an object being waited upon - and IO checking
- to determine if there is pending IO - to identify threads which are
executing workloads.  If no threads are executing workloads, then we may
be in a starvation state and will initialize new threads.  This should
fix the issue that has been previously reported with SPM hangs.

This is non-portable and ideally we would have an alternative approach
to tracking thread state, but this should at least unblock the current
situation.

### DIFF
--- a/src/event/workqueue.c
+++ b/src/event/workqueue.c
@@ -22,6 +22,10 @@
 
 #if DISPATCH_USE_INTERNAL_WORKQUEUE
 
+#if defined(_WIN32)
+#include <wct.h>
+#endif
+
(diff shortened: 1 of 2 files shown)


compnerd:
CC: @rokhinip

compnerd:
@swift-ci please test

PR #586, opened 2021-11-17
add arm64 support for M1 Mac

DanboDuan:


DanboDuan:
@compnerd here for libdispatch

cmake/modules/SwiftSupport.cmake
```diff
@@ -9,6 +9,8 @@
 function(get_swift_host_arch result_var_name)
   if("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64")
     set("${result_var_name}" "x86_64" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "arm64")
```
compnerd, line 12: Probably best to match the other packages and merge the arm64 and aarch64 cases via a MATCH instead of STREQUALs.

compnerd (changes requested):

cmake/modules/SwiftSupport.cmake
DanboDuan: done.

compnerd (approved):

compnerd:
@swift-ci please test

compnerd:
CC: @rokhinip

PR #589, opened 2021-11-27
dispatch: remove stray semicolon (NFC

compnerd:
This cleans up a stray semicolon to allow building with `-Werror=c++98-compat-extra-semi` which is enabled by default in some versions of clang and produces an annoying number of diagnostics when building Swift.


compnerd:
CC: @rokhinip

compnerd:
@swift-ci please test

PR #590, opened 2021-12-10
Fix possible crash in _dispatch_wait_for_enqueuer on Android armeabi-v7a

triplef:
Caused by `__builtin_arm_wfe()` sometimes causing a `SIGILL`.

For details please see https://bugs.swift.org/browse/SR-15166.

### DIFF
--- a/src/shims/yield.c
+++ b/src/shims/yield.c
@@ -36,7 +36,7 @@ void *
 _dispatch_wait_for_enqueuer(void **ptr)
 {
 #if !DISPATCH_HW_CONFIG_UP
-#if defined(__arm__) || defined(__arm64__)
+#if (defined(__arm__) && defined(__APPLE__)) || defined(__arm64__)
 	int spins = DISPATCH_WAIT_SPINS_WFE;
 	void *value;
 	while (unlikely(spins-- > 0)) {


buttaface:
@etcwilde, would you review?

src/shims/yield.c
```diff
@@ -44,7 +44,7 @@ _dispatch_wait_for_enqueuer(void **ptr)
 			__builtin_arm_clrex();
 			return value;
 		}
-		__builtin_arm_wfe();
+		dispatch_hardware_pause();
```
rokhinip, line 47: Can you please specify here out what the disassembly of this function is with this replacement? This change is not going to work here because on darwin we want this to be wfe. We do not want yield.

yield is not what we want for this - we're trying to monitor a memory region for concurrent rights which is what wfe and ldrex does.

rokhinip: If you want to take the hardware_pause() path, we might want to change the defines on line 39 instead since the loop below implements things correctly for when you want to use pause().

Does armeabi-v7a/NEON not support WFE?

rokhinip (changes requested):

src/shims/yield.c
rokhinip: @das pointed out that in armv7 spec (https://developer.arm.com/documentation/ddi0406/cd) it says that WFE is available at all privilege levels.

See Section B1.8.13 of armv7 spec, page B1-1201:
```
The Wait For Event instruction, WFE, is available at all privilege levels, see WFE on page A8-1105.
```

So the hardware should have implemented wfe and we shouldn't be hitting a SIGILL.

triplef: Yes indeed, WFE should be available on armeabi-v7a/NEON. Furthermore as mentioned in https://bugs.swift.org/browse/SR-15166, if I interpret the assembly correctly only the _second_ WFE instruction of the unrolled loop crashes.

Most affected devices are running Cortex-A53 CPUs, but we‘re seeing a wide range of devices by various manufacturers affected.

I’m happy to provide any other information that could help figure out why this is crashing at all.

MadCoder: If you need to do this patch for broken HW, you need to make it a build configuration and not the default. on aarch64 this wfe is _precisely_ what we want as it waits for a store on the cacheline while holding the monitor. This is not the required behavior on armv7 cores, so we use WFE assuming that there is a decrementer which is not required so it wouldn't be insane to tune between WFE and YIELD.

YIELD however really isn't an equivalent and when YIELD is used you also want to tune the number of SPINLOOPS done to look more like Intel's.

all in all this patch is incorrect under its current form

buttaface: @MadCoder, this is a hack that I suggested in the linked bug without really understanding this problem space, after a couple minutes looking in a search engine, that happens to stop the crashes on Android armv7. Since you know this area better, perhaps you can suggest an alternate fix that would be correct.

rokhinip: > Yes indeed, WFE should be available on armeabi-v7a/NEON. Furthermore as mentioned in https://bugs.swift.org/browse/SR-15166, if I interpret the assembly correctly only the _second_ WFE instruction of the unrolled loop crashes.
> 
> Most affected devices are running Cortex-A53 CPUs, but we‘re seeing a wide range of devices by various manufacturers affected.

I think we should do the following:

Investigate further to figure out why WFE is crashing on the devices sometimes. If we're certain that we're actually trying to execute WFE (work back from the encoding of the instruction that the CPU is trying to execute) and the problem is that this crashes on some CPUs sometimes, then we might want to account for broken hardware by changing the following instead. 

This makes sure that the WFE behaviour happens on Apple platforms where we know this is well behaved and we fallback to YIELD based behaviour for all non-apple platforms:

```
void *
_dispatch_wait_for_enqueuer(void **ptr)
{
- #if !DISPATCH_HW_CONFIG_UP
+ #if !DISPATCH_HW_CONFIG_UP && defined(__APPLE__)
   #if defined(__arm__) || defined(__arm64__)
```

MadCoder: @rokhinip yes that makes sense, we still need to adjust the number of "loops". The ARM setting on APPLE platforms assumes a wfe that takes µs's to break out, where yield will typically be a few cycles and "go 100 to 1000x as fast". So looping 5 times only around it (or whatever it is our setting is for arm/arm64) is insufficient.

That setting should also be predicated on __APPLE__ indeed.

triplef: Thank you both for your feedback and thoughts on this!

I think we could further restrict this to only non-Apple 32-bit ARM like this, as it doesn’t happen on arm64:
```c
#if (defined(__arm__) && defined(__APPLE__)) || defined(__arm64__)
```

Regarding adjusting to the number of loops, is that not already done in the thus activated code path using `DISPATCH_WAIT_SPINS` instead of `DISPATCH_WAIT_SPINS_WFE`? These are defined as this:
```c
#define DISPATCH_WAIT_SPINS_WFE 10
#define DISPATCH_WAIT_SPINS 1024
```

While I couldn’t find anything concrete on this issue, there’s at least one other project that seemed to be running into the same issue: https://github.com/google/filament/commit/0cbf6f8d76c333629c96b6fef8d994f65c88c8cf

rokhinip: Oh you're right, `DISPATCH_WAIT_SPINS_WFE` and `DISPATCH_WAIT_SPINS` handles this correctly already. 

It seems like we have both `DISPATCH_WAIT_SPINS_WFE` and `DISPATCH_WAIT_SPINS` are defined on all multi-processor configs so we're okay with that.

triplef: I’ve updated the patch, but I’ll keep this unresolved until I’m able to test this with a couple armeabi-v7a devices after the holidays.

rokhinip:
@swift-ci please test

src/shims/yield.c
triplef: Sorry for the delay. I’ve now tested the change on a number of Android devices and it works as expected. I’d appreciate if this could be merged. 🙏

triplef: Thanks for the link @buttaface. We don’t use Swift though, just Objective-C – sorry I can’t help there.

rokhinip (approved):

PR #592, opened 2022-02-08
[swift] Stop using _SwiftDispatchOverlayShims

lorentey:
Stop importing _SwiftDispatchOverlayShims from the Swift stdlib; instead, copy its contents to a new private _DispatchOverlayShims module, and import it with the @_implementationOnly attribute so that clients no longer need to have access to it to merely import Dispatch.

The Swift stdlib currently ships the _SwiftDispatchOverlayShims module, but we don't want to keep it there -- it declares APIs that belong to the dispatch project, and it is best maintained by the same people who maintain the library itself.

The Dispatch overlay on Darwin has not been importing this module for a couple years now.

### DIFF
--- a/src/swift/Block.swift
+++ b/src/swift/Block.swift
@@ -11,7 +11,7 @@
 //===----------------------------------------------------------------------===//
 
 import CDispatch
-import _SwiftDispatchOverlayShims
+@_implementationOnly import _DispatchOverlayShims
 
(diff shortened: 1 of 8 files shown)


lorentey:
@swift-ci test

Azoy (approved):
❤️

rokhinip (approved):

PR #595, opened 2022-02-22
[Win32] Initialize Windows Runtime on queue threads

triplef:
On Windows, all threads that want to use Windows Runtime or COM components must be initialized with a call to [CoInitializeEx()](https://docs.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coinitializeex) or the slightly more modern [RoInitialize()](https://docs.microsoft.com/en-us/windows/win32/api/roapi/nf-roapi-roinitialize) (same as `Windows::Foundation::Initialize`).

This change initializes all libdispatch threads so it’s possible to use COM and Windows Runtime APIs.

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -6264,8 +6264,13 @@ _dispatch_worker_thread(void *context)
 static unsigned WINAPI
 _dispatch_worker_thread_thunk(LPVOID lpParameter)
 {
-  _dispatch_worker_thread(lpParameter);
-  return 0;
+	HRESULT hr = CoInitializeEx(NULL, COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE);
+	if (FAILED(hr)) {
+		_dispatch_client_assert_fail("Error %ld initializing Windows Runtime", hr);
+	}
+	_dispatch_worker_thread(lpParameter);
+	CoUninitialize();
+	return 0;
 }
 #endif // defined(_WIN32)
 #endif // DISPATCH_USE_PTHREAD_POOL


src/queue.c
```diff
@@ -6264,8 +6267,15 @@ _dispatch_worker_thread(void *context)
 static unsigned WINAPI
 _dispatch_worker_thread_thunk(LPVOID lpParameter)
 {
-  _dispatch_worker_thread(lpParameter);
-  return 0;
+	HRESULT hr = RoInitialize(RO_INIT_MULTITHREADED);
```
compnerd, line 6270: Why `RoInitialize`?  I think that limiting this to `CoInitialize` is better.  I am still concerned about the threading model though - MTA cannot make input-synchronized calls, and this will convert async to sync due to the MTA, plus this disables the message filtering (or is this equivalent to `COINIT_MULTITHREADED` and this marks everything for free threading?).  I'm not sure if we really want to use the `RoInitialize` here rather than `CoInitialize`.  The COM initializer gives greater flexibility (e.g. disabling DDE, trading off memory for speed, etc).

src/queue.c
```diff
@@ -6264,8 +6267,15 @@ _dispatch_worker_thread(void *context)
...
+
+	_dispatch_worker_thread(lpParameter);
+
+	if (SUCCEEDED(hr)) {
```
compnerd, line 6274: Perhaps this check should be hoisted?  Do we not want to ensure that the operation was successful?

triplef: Good point. Does that look right now, or should we also return `hr`?

src/queue.c
triplef: Thanks for your feedback! We actually tried both `CoInitialize` and `RoInitialize`, and both seemed to work in our tests. I couldn’t find any detailed description on about the differences, and we only chose the latter because of the documentation recommending it for use with Windows Runtime APIs.

I updated it with `COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE`, does that make sense in this context?

src/queue.c
```diff
@@ -6264,8 +6264,12 @@ _dispatch_worker_thread(void *context)
 static unsigned WINAPI
 _dispatch_worker_thread_thunk(LPVOID lpParameter)
 {
-  _dispatch_worker_thread(lpParameter);
-  return 0;
+	HRESULT hr = CoInitializeEx(NULL, COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE);
+	if (SUCCEEDED(hr)) {
+		_dispatch_worker_thread(lpParameter);
+		CoUninitialize();
+	}
```
compnerd, line 6271: This would mean that failure to initialize COM would terminate the thread immediately - even if COM is never used on the thread.  Are we certain that there are no cases in which COM initialization could fail that would not be recoverable?  (e.g. if COM initialization fails, we know that either you have no memory)

triplef: I hoisted the succeeded check based on your previous comment.

I am not sure which way is the right way to do this (ignore COM initialization failure or not). I have a hard time finding good documentation on COM initialization that would answer your questions, and I can see arguments for both options. That being said, I’m leaning to how it is implemented now as I am guessing COM initialization is pretty unlikely to fail (other than out of memory) as it’s such an integral part of Windows programming.

compnerd: I'm worried about the diagnosability of that.  Perhaps it makes sense to make the failure fatal then?  Right now, if this fails, we don't have any notification but are unable to recover from that because any attempt to create a thread will fail.

stevenbrix:
is it possible to have app authors determine which initialization method to use? would this make the `@MainActor` thread an MTA thread? that should likely be STA (which is what conventional Windows GUI apps use today)

src/queue.c
stevenbrix: given the return values specified here: https://learn.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coinitializeex#return-value

i think a `fatalError` is OK since we know we aren't going to pass invalid arguments, and OOM or E_UNEXPECTED sound like fine reasons to `fatalError` to me

compnerd:
> is it possible to have app authors determine which initialization method to use? would this make the `@MainActor` thread an MTA thread? that should likely be STA (which is what conventional Windows GUI apps use today)

No, I don't see how to do that.  These threads are spawned without user intervention, and I don't remember any APIs to control the worker threads it creates.  Is there a global Windows API that lets you set the apartment model?

stevenbrix:
@compnerd given the behavior of .NET it seems correct to initialize background threads to MTA: https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.setapartmentstate?view=net-7.0#remarks

if needed perhaps we could evolve this in the future but not an immediate need

compnerd:
> @compnerd given the behavior of .NET it seems correct to initialize background threads to MTA: https://learn.microsoft.com/en-us/dotnet/api/system.threading.thread.setapartmentstate?view=net-7.0#remarks
> 
> if needed perhaps we could evolve this in the future but not an immediate need

Well, this is only for background threads, so this sounds like we have a path forward.  I think that the only thing missing here is the death for an error scenario.

@triplef any chance you can update the patch?

stevenbrix (commented):
looking at the implementation of .NET, it looks like they call both:  https://github.com/dotnet/runtime/blob/991b15056e276209b64a711face35715ba335c77/src/coreclr/vm/threads.cpp#L4898

https://github.com/dotnet/runtime/blob/991b15056e276209b64a711face35715ba335c77/src/coreclr/vm/threads.cpp#L5054-L5057

also looks like ignoring an unexpected HRESULT might be acceptable?

compnerd:
> looks like ignoring an unexpected HRESULT might be acceptable?

It may be that its easier to debug the .NET runtime, things are a bit more complicated with debugging dispatch, I'd rather be more stringent.  The number of times that the dispatch assertions have saved me hours of debugging is ridiculous.  Blindly following .NET doesn't seem like the right thing to me.

src/queue.c
triplef: Sounds good. What’s the best way to trigger a fatal error here? I think `fatalError()` is only available in Swift. Would `_dispatch_assert_crash()` make sense, or any other function/macro from internal.h?

compnerd: yeah `_dispatch_assert_crash` would be the best way.

triplef: Thanks. I ended up using `_dispatch_client_assert_fail` because it takes a format string, which enables adding the error code to the message.

triplef:
I’ve updated the patch with a fatal error message in case COM initialization fails. Let me know if there’s any other changes needed.

compnerd:
@swift-ci please test

compnerd:

compnerd:

compnerd (approved):

src/queue.c
```diff
@@ -6264,8 +6264,13 @@ _dispatch_worker_thread(void *context)
 static unsigned WINAPI
 _dispatch_worker_thread_thunk(LPVOID lpParameter)
 {
-  _dispatch_worker_thread(lpParameter);
-  return 0;
+	HRESULT hr = CoInitializeEx(NULL, COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE);
+	if (FAILED(hr)) {
+		_dispatch_client_assert_fail("Error %ld initializing Windows Runtime", hr);
```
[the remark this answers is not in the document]
triplef: There was an earlier discussion about these functions [here](https://github.com/apple/swift-corelibs-libdispatch/pull/595#discussion_r812336844), does that maybe answer your question? Unfortunately I’m not familiar with the differences between the two functions.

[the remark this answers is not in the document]
rilysh: `CoInitializeEx` initialize concurrent model as STA, where `RoInitialize` by default uses MTA. Though passing `RO_INIT_SINGLETHREADED` will use STA instead.

PR #596, opened 2022-03-02
Skip unneeded check if last_locked is 0

GameCubeGBA:
It will be be checked anyway, and redundant checks mean more points of failure.


rokhinip:
This seems like a reboot of https://github.com/apple/swift-corelibs-libdispatch/pull/566 and https://github.com/apple/swift-corelibs-libdispatch/pull/564 which were declined.

PR #597, opened 2022-03-10
Fixed bool type for C89

damon-kwok:
Before c99 doesn't have `_Bool‘ and `__STDC_VERSION__'. Compensate. 
``` c
#ifndef __STDC_VERSION__
#define __STDC_VERSION__ 0
#endif
#if defined(_MSC_VER) || (__STDC_VERSION__ < 199901L)
/* `MSVC' and `Before c99' doesn't have <stdbool.h>. Compensate. */
typedef char bool;
#define true (bool) 1
#define false (bool) 0
#else
#include <stdbool.h>
#endif
```

### DIFF
--- a/src/BlocksRuntime/Block_private.h
+++ b/src/BlocksRuntime/Block_private.h
@@ -11,7 +11,18 @@
 #ifndef _BLOCK_PRIVATE_H_
 #define _BLOCK_PRIVATE_H_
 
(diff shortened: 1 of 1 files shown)


compnerd:
I'm going to close this for now. We need a response from the original reporter before we can proceed.

PR #767, opened 2022-09-08
Performance fix: Lower operations and function calls by checking for zone_address before doing expensive memory reads

AtariDreams:
Not only does this defer expensive operations to when we can prove they are needed, but also the reduction of casts also help with performance on libdispatch!


AtariDreams:
@millenomi could you please check this

tomerd:
cc @parkera

PR #769, opened 2022-10-02
Compare with (void *)0 instead of casting to uintptr_t

AtariDreams:
Such casts otherwise pessimize optimization

### DIFF
--- a/src/event/event_internal.h
+++ b/src/event/event_internal.h
@@ -445,7 +445,7 @@ DISPATCH_ALWAYS_INLINE
 static inline bool
 _dispatch_needs_to_return_to_kernel(void)
 {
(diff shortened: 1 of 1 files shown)


rokhinip:
What is this change trying to address? This seems like a no-op.

AtariDreams:
It aligns with the definition of NULL better

rokhinip:
Well firstly, `_dispatch_needs_to_return_to_kernel` is reading an integer value out of a TSD slot, in order to determine if we need to return to the kernel to pump events. So a comparison to NULL is not valid here - nor is `(void *)0` because we're not reading pointers.

This change seems like a change for the sake of it without understanding how this code is used in the larger context. Please refer to the guidelines document on how to identify and make useful changes to the project.

PR #771, opened 2023-01-05
config, os: add support for Musl libc

MaxDesiatov:
According to Musl's FAQ, `__BEGIN_DECLS` and `__END_DECLS` macros come from Glibc private headers, thus we need to define them manually in case those aren't defined.

Additionally, for Musl we need to override `HAVE_DECL_PROGRAM_INVOCATION_SHORT_NAME`, for which a corresponding `#ifndef` was added in `config.h`.


MaxDesiatov:
@swift-ci test

MaxDesiatov:
@swift-ci test

os/generic_win_base.h
```diff
@@ -25,14 +25,6 @@
 typedef int mode_t;
 typedef void pthread_attr_t;
 
-#if defined(__cplusplus)
```
rokhinip, line 28: Should this instead just be added to `os/generic_unix_base.h` instead? This seems a platform specific thing instead of moving it to os/object.h - this declaration has nothing to do with os objects.

config/config.h
```diff
@@ -47,7 +47,9 @@
 
 /* Define to 1 if you have the declaration of `program_invocation_short_name',
    and to 0 if you don't. */
+#ifndef HAVE_DECL_PROGRAM_INVOCATION_SHORT_NAME
```
rokhinip, line 50: I don't see similar ifndef stuff for other knobs that could be configured - I'm not very familiar with this part of the dispatch build system but this feels like it might not be the place to add this.

cc: @compnerd to weigh in.

MaxDesiatov: You're right. In fact, when `_GNU_SOURCE` is defined (on a level above, in the Swift build system) related errors go away and I don't have to add an `#ifndef` here. That's cleaned up now.

os/generic_win_base.h
MaxDesiatov: I moved it to a new `<os/generic_base.h>` header.

MaxDesiatov:
@swift-ci test

os/CMakeLists.txt
```diff
@@ -3,9 +3,10 @@
 # voucher_private.h are included in the source tarball
 
 install(FILES
-          object.h
+          generic_base.h
```
rokhinip, line 6: Should the unix and windows base.h versions just import generic_base.h?

rokhinip (approved):

os/CMakeLists.txt
MaxDesiatov: I assume this new header needs to be installed to be importable at the expected install location.

compnerd:
This regressed the Windows release.  You introduced a new distributed header but did not update the Windows installer manifest.

MaxDesiatov:
Sorry about that, could you point to the installer manifest to be updated? Additionally, is there a Swift CI setup on this repository I could use in subsequent PRs to verify this?

compnerd:
The manifests are now in apple/swift-installer-scripts.  I think that setting up CI here is a great idea, @shahmishal should be able to help with that.  However, that alone would be insufficient, you should also do a cross-repository test with a full toolchain build.  The dispatch builds are tested as part of the regular PR tests on apple/swift and on a toolchain build.

PR #774, opened 2023-01-25
Update runtime.c to match Core Foundation version

AtariDreams:
runtime.c should be consistent between the two.


AtariDreams:
@compnerd

compnerd:
This seems unnecessary and this is entirely unrelated to CF.

PR #777, opened 2023-02-04
Reorder checks in _dispatch_unote_create

AtariDreams:
Because the second conditional implies the first is false, we can move the second conditional upwards, since they both have conditional overlapping.

### DIFF
--- a/src/event/event.c
+++ b/src/event/event.c
@@ -35,11 +35,11 @@ _dispatch_unote_create(dispatch_source_type_t dst,
 	dispatch_unote_linkage_t dul;
 	dispatch_unote_class_t du;
 
-	if (mask & ~dst->dst_mask) {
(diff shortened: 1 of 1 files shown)


AtariDreams:
@compnerd

etcwilde:
Hi, it's unclear what you're trying to do with this patch. It doesn't provide any apparent gain, while being potentially breaking. Re-ordering the conditions is not meaningful. Additionally, the flipping of the short-circuiting `&&`  operands is making it difficult for me to determine whether this results in the same behavior.

If you have measurable performance numbers and there is a statistically significant improvement, or the change meaningfully improves the readability of the code, we could have a discussion about merging this patch, but as far as I can tell, it does neither. My vote is to close this PR. @rokhinip, I'll let you make the final call here.

PR #778, opened 2023-02-04
Ventura Merge

AtariDreams:
This PR aims to merge the changes from macOS Ventura and iOS 16 to this repo.

It is not perfect. Please let me know if there are any concerns.


AtariDreams:
@compnerd Thoughts on this?

compnerd:
I'm not sure that this is a good time to merge this; furthermore, there was a previous merge that was stale and I think that it would be better to figure out why that stalled rather than just start afresh.

PR #781, opened 2023-03-03
Use PIPE_NOWAIT on Windows to avoid blocking on writes

tristanlabelle:
`libdispatch` depends on non-blocking writes to retry later. For pipes on Windows, we achieved this using `NtQueryInformationFile` to get the `WriteQuotaAvailable` for the pipe, but this function returns `0` when there is a blocking read on the receiver side (this appears to be an OS bug), which causes us to never proceed to the write.

This change rather sets `PIPE_NOWAIT` on the pipe's write handle, similar to what is done on other platforms. This call will fail if the pipe handle does not have `FILE_WRITE_ATTRIBUTES` set, which is not a problem with the typical case of using a `Foundation.Pipe`. If we cannot set `PIPE_NOWAIT`, we will proceed with blocking writes, which in some circumstances could lead to hangs.

This resolves an issue with `SourceKit-LSP` on Windows where tests and real-life scenarios would hang because a pipe could never be written to.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -1436,19 +1436,23 @@ _dispatch_fd_entry_create_with_fd(dispatch_fd_t fd, uintptr_t hash)
 				unsigned long value = 1;
 				int result = ioctlsocket((SOCKET)fd, (long)FIONBIO, &value);
 				(void)dispatch_assume_zero(result);
-			}
-			else {
+			} else {
 				// Try to make writing nonblocking, although pipes not coming
 				// from Foundation.Pipe may not have FILE_WRITE_ATTRIBUTES.
-				DWORD pipe_mode = PIPE_NOWAIT;
-				if (!SetNamedPipeHandleState((HANDLE)fd, &pipe_mode, NULL,
-						NULL)) {
-					// We may end up blocking on subsequent writes, but we
-					// don't have a good alternative. The WriteQuotaAvailable
-					// from NtQueryInformationFile erroneously returns 0 when
-					// there is a blocking read on the other end of the pipe.
-					_dispatch_fd_entry_debug("failed to set PIPE_NOWAIT",
-							fd_entry);
+				DWORD dwPipeMode = 0;
+				if (GetNamedPipeHandleState((HANDLE)fd, &dwPipeMode, NULL,
+						NULL, NULL, NULL, 0) && !(dwPipeMode & PIPE_NOWAIT)) {
+					dwPipeMode |= PIPE_NOWAIT;
+					if (!SetNamedPipeHandleState((HANDLE)fd, &dwPipeMode,
+							NULL, NULL)) {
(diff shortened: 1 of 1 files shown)


src/io.c
```diff
@@ -1432,9 +1432,26 @@ _dispatch_fd_entry_create_with_fd(dispatch_fd_t fd, uintptr_t hash)
...
 		if (dwType == FILE_TYPE_PIPE) {
-			unsigned long value = 1;
-			int result = ioctlsocket((SOCKET)fd, (long)FIONBIO, &value);
-			(void)dispatch_assume_zero(result);
+			if (_dispatch_handle_is_socket((HANDLE)fd)) {
+				unsigned long value = 1;
+				int result = ioctlsocket((SOCKET)fd, (long)FIONBIO, &value);
+				(void)dispatch_assume_zero(result);
+			}
+			else {
```
compnerd, line 1440: Please coddle the braces.

```suggestion
			} else {
```

src/io.c
```diff
@@ -1432,9 +1432,26 @@ _dispatch_fd_entry_create_with_fd(dispatch_fd_t fd, uintptr_t hash)
...
+				// Try to make writing nonblocking, although pipes not coming
+				// from Foundation.Pipe may not have FILE_WRITE_ATTRIBUTES.
+				DWORD pipe_mode = PIPE_NOWAIT;
+				if (!SetNamedPipeHandleState((HANDLE)fd, &pipe_mode, NULL,
```
compnerd, line 1444: Would this potentially clear the `PIPE_READMODE_MESSAGE`?  If so, should we do a `GetNamedPipeHandleState` first?

tristanlabelle: Good point, that seems safer

src/io.c
tristanlabelle: Fixed in latest

compnerd:
@swift-ci please test

compnerd:

compnerd:

src/io.c
```diff
@@ -2523,6 +2526,9 @@ _dispatch_operation_perform(dispatch_operation_t op)
 						processed = 0;
 					}
 				}
+				if (bSuccess && processed == 0) {
+					err = EAGAIN;
```
compnerd, line 2530: Does this not need a `goto error`?  This will currently change the path to `return DISPATCH_OP_DELIVER_AND_COMPLETE` rather than `return DISPATCH_OP_COMPLETE_RESUME` or `return DISPATCH_OP_RESUME`.

tristanlabelle: Oops, yes good catch!

compnerd:
@swift-ci please test

compnerd:

compnerd:
@swift-ci please test

tristanlabelle:
I ran tests locally. 22/22 passed.

Issue #669, opened 2017-11-06
[SR-6310] DispatchTimeInterval does not implement Equatable on Linux


swift-ci:
|                  |                 |
   |------------------|-----------------|
   |Previous ID       | SR-6310      |
   |Radar             | rdar://problem/64517050         |
   |Original Reporter | sharom (JIRA User)      |
   |Type              | Bug    |

<details>
  <summary>Environment</summary>

-   Swift version 4.0 (swift-4.0-RELEASE)

-   Ubuntu 16.04 64bit

</details>

   <details>
  <summary>Additional Detail from JIRA</summary>

   |                  |                 |
   |------------------|-----------------|
   |Votes             | 2         |
   |Component/s       | libdispatch    |
   |Labels            | Bug        |
   |Assignee          | None      |
   |Priority          | Medium      |

   

   md5: b7578c3c2431bcac432a15e7a4bcd638

  </details>

**Issue Description:**

`DispatchTimeInterval` satisfies to `Equatable` protocol on Mac but doesn't on Linux.

PR #782, opened 2023-03-06
Mark DispatchTimeInterval Equatable

rnro:
We can mark `DispatchTimeInterval` as equatable because it already implements the equality check operator `==`.

This resolves an API disparity - on Darwin `DispatchTimeInterval` is already `Equatable`.

Doing this will allow `DispatchTimeInterval` to be used with a wide range of library functions which are defined to operate on `Equatable` objects. For example this is useful in tests when you might wish to use `XCTAssertEqual`.

Resolves https://github.com/apple/swift-corelibs-libdispatch/issues/669


MaxDesiatov:
@swift-ci test

rokhinip (approved):

PR #783, opened 2023-03-09
Un-hardcode cmake install directories

tristanlabelle:
CMake install places binary files under the `CMAKE_INSTALL_BINDIR` subdirectory, but some CMake files hardcode `bin` instead (the default values). This change un-hardcodes this. This has no effect by default, but it will allow redirecting the binaries to a different output directory, which will help with packaging installers for multiple architectures in the Windows build.


compnerd:
@swift-ci please test

PR #784, opened 2023-03-31
Fix build break due to calling convention mismatch on Windows

tristanlabelle:
My local Windows libdispatch build recently started failing due to a calling convention mismatch with the function we use as the `PINIT_ONCE_FN` argument of a `InitOnceExecuteOnce` invocation. The fix is to specify it explicitly.


compnerd (approved):

compnerd:
@swift-ci please test

PR #791, opened 2023-05-23
shims: fix a subtle bug in semaphore initialisation on Windows

compnerd:
This function is the initializer for the semaphore.  The seamphore storage itself may be stack allocated (or heap allocated) but without guarantee of 0-initialisation.  As a result, the subsequent CAS for the atomic replacement will fail silently, leaving the previously non-zero value in place, indicating that the value is a valid handle.  This would fail randomly and would ultimately result in a crash in the `CloseHandle` call associated with the clean up.

This issue was identified by SwiftLint on Windows.


compnerd:
@swift-ci please test

PR #793, opened 2023-06-02
dispatch: install the correct modulemap when building static

compnerd:
There is a subtle difference between libdispatch built dynamically and statically: DispatchStubs.  We erroneously emit ObjC runtime calls into the build and the stubs provides a shim to provide a definition on non-ObjC runtime enabled targets.  When built dynamically, this is internalised and distributed as part of dispatch.dll/libdispatch.so, however when built statically, the consumer is responsible for linking against DispatchStubs.  The modulemap reflects this and we need to ensure that we correctly provide that modulemap.

Thanks to @MaxDesiatov for the help with debugging and testing a fix for this!

### DIFF
--- a/dispatch/CMakeLists.txt
+++ b/dispatch/CMakeLists.txt
@@ -1,8 +1,10 @@
 
 if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
   set(DISPATCH_MODULE_MAP ${PROJECT_SOURCE_DIR}/dispatch/darwin/module.modulemap)
-else()
+elseif(BUILD_SHARED_LIBS)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

drexin (approved):

MaxDesiatov:
Both in cross-repo testing and for this PR specifically it fails with
```
<unknown>:0: error: module map file '/home/build-user/swift-corelibs-libdispatch/dispatch/module.modulemap' not found
/home/build-user/swift-corelibs-libdispatch/src/swift/Block.swift:13:8: error: no such module 'CDispatch'
import CDispatch
```

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #796, opened 2023-06-14
Prevent infinite attempts to write to Win32 pipes 

tristanlabelle:
#781 set the `PIPE_NOWAIT` flag on pipes in an attempt to get posix-like semantics. However, Windows will immediately return `0` from `WriteFile` if the *entire* write would not complete. This means that asking to write more bytes than the size of the pipe's buffer will never do anything. The same applies to asking to write more bytes than is left available in the buffer.

This change keeps #781's `PIPE_NOWAIT` and brings back some of its pipe buffer size checking logic, with the caveat to be careful with trusting `FILE_PIPE_LOCAL_INFORMATION.WriteQuotaAvailable` since we've seen it return 0 when a reader is blocked waiting on the other end of the pipe.

Fixes https://github.com/apple/sourcekit-lsp/issues/752

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -2510,6 +2510,23 @@ syscall:
 				}
 				bSuccess = TRUE;
 			} else if (GetFileType(hFile) == FILE_TYPE_PIPE) {
+				// WriteFile with more bytes than are available in the
+				// buffer of a NOWAIT pipe will immediately return 0,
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd (approved):

compnerd:
CC: @rokhinip

PR #801, opened 2023-11-19
Generate libdispatch.pc on non-apple platforms

hmelder:
libdispatch should install a pkg-config file as it is the defacto standard querying mechanism on Linux, and other platforms (excluding Apple).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -285,10 +285,12 @@ else()
   set(INSTALL_OS_HEADERS_DIR "include/os" CACHE PATH "Path where the headers will be installed")
 endif()
 
-configure_file("${PROJECT_SOURCE_DIR}/cmake/libdispatch.pc.in"
-               "${PROJECT_BINARY_DIR}/libdispatch.pc" @ONLY)
+if(NOT APPLE)
+  configure_file("${PROJECT_SOURCE_DIR}/cmake/libdispatch.pc.in"
+                 "${PROJECT_BINARY_DIR}/libdispatch.pc" @ONLY)
 
-install(FILES ${PROJECT_BINARY_DIR}/libdispatch.pc DESTINATION ${CMAKE_INSTALL_LIBDIR}/pkgconfig)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -285,6 +285,12 @@ else()
   set(INSTALL_OS_HEADERS_DIR "include/os" CACHE PATH "Path where the headers will be installed")
 endif()
 
+if(NOT APPLE)
```
compnerd, line 288: This should not happen on Windows or Android, please explicitly check for Linux and BSD.

compnerd (commented):
I'm not sure how I feel about this.  We should instead just provide a DispatchConfig.cmake instead I think.

CMakeLists.txt
hmelder: You are right. We use pkgconfig on windows aswell, but this a specific case. I will change it to UNIX AND NOT APPLE

hmelder:
> I'm not sure how I feel about this. We should instead just provide a DispatchConfig.cmake instead I think.

I will look into that!

hmelder:
> I'm not sure how I feel about this. We should instead just provide a DispatchConfig.cmake instead I think.

Do you mean a Configuration in cmake/config? Should I write a cmake function or just extract the existing generation of the pkg config file?

hmelder:
@compnerd

CMakeLists.txt
compnerd: Who is "we" in this case?  We (the swift project) definitely doesn't use pkg-config on Windows, nor does "we" (The Browser Company).

compnerd:
> > I'm not sure how I feel about this. We should instead just provide a DispatchConfig.cmake instead I think.
> 
> Do you mean a Configuration in cmake/config? Should I write a cmake function or just extract the existing generation of the pkg config file?

No, CMake has its own means of providing the configuration for the project.  Look at the CMake documentation related to `FindPackage`.

Issue #760, opened 2022-05-23
Swift Actor/Tasks concurrency on Linux - Lock contention in rand()


freef4ll:
We have a workload pipeline which is chaining several thousand Actors to each other via AsyncStream processing pipeline.
There is a multiplication affect that a single event at the start of the processing pipeline will be amplified as the event will be delivered to several Tasks processing the events concurrently. The processing time of each wakeup is currently quite small and on several microseconds range currently.

Under Linux, what was observed when stressing this processing pipeline is that ~45% of the stacks show `__DISPATCH_ROOT_QUEUE_CONTENDED_WAIT__()`, which is leading to lock contention in [glibc rand()](https://elixir.bootlin.com/glibc/glibc-2.35/source/stdlib/random.c#L291) - as there are ~60 threads which are created and they all contend here:

```
            7f193794a2db futex_wait+0x2b (inlined)
            7f193794a2db __GI___lll_lock_wait_private+0x2b (inlined)
            7f19378ff29b __random+0x6b (/usr/lib/x86_64-linux-gnu/libc.so.6)
            7f19378ff76c rand+0xc (/usr/lib/x86_64-linux-gnu/libc.so.6)
            7f1937bac612 __DISPATCH_ROOT_QUEUE_CONTENDED_WAIT__+0x12 (/usr/lib/swift/linux/libdispatch.so)
```

This is occurring in every entrance of [__DISPATCH_ROOT_QUEUE_CONTENDED_WAIT__()](https://github.com/apple/swift-corelibs-libdispatch/blob/main/src/queue.c#L5847-L5858),  while using macro [_dispatch_contention_wait_until()](https://github.com/compnerd/swift-corelibs-libdispatch/blob/master/src/shims/yield.h#L118-120) which in turn uses [_dispatch_contention_spins()](https://github.com/compnerd/swift-corelibs-libdispatch/blob/master/src/shims/yield.h#L115), in here the rand() call comes in and the macro produces just these 4 values: 31, 63, 95 and 127 for how many pause/yield instructions to execute.

The following example can reproduce the issue where ~28% of the time when sampling is spent in the code path mentioned.
The example creates 5000 tasks which work between 1μs and 3μs and then sleep for random 6-10 milliseconds. The point of the test is to create the contention and to illustrate the issue with rand():

```swift
// $ swift package init --type executable --name RandomTasks
// $ cat  Sources/RandomTasks/main.swift &&  swift run -c release

import Foundation

[...]

PR #804, opened 2023-12-14
Avoid using rand() that leads to a lock contention

ilya-fedin:
Fixes https://github.com/apple/swift-corelibs-libdispatch/issues/760


rjmccall:
@swift-ci Please test

hassila:
Some results here: https://github.com/apple/swift-corelibs-libdispatch/issues/760#issuecomment-1855746788

Seems this patch eliminates the bottleneck of rand().

ktoso:
I think we can consider this for the linux only release in January. Will update soon

PR #805, opened 2023-12-15
Prefer aligned_alloc over posix_memalign

AtariDreams:
aligned_alloc is more cleaner and portable.

### DIFF
--- a/src/io.c
+++ b/src/io.c
@@ -2374,10 +2374,7 @@ _dispatch_operation_perform(dispatch_operation_t op)
 			}
 			op->buf = _aligned_malloc(op->buf_siz, siInfo.dwPageSize);
 #else
-			err = posix_memalign(&op->buf, (size_t)PAGE_SIZE, op->buf_siz);
-			if (err != 0) {
-				goto error;
-			}
+			op->buf = aligned_alloc((size_t)PAGE_SIZE, op->buf_siz)
 #endif
 			_dispatch_op_debug("buffer allocated", op);
 		} else if (op->direction == DOP_DIR_WRITE) {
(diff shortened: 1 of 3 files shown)


AtariDreams:
@compnerd thoughts?

AtariDreams:
@ktoso can we please test this one too

AtariDreams:
@compnerd Can we please do a test?

compnerd:
Should we not have a fallback path here?

AtariDreams:
> Should we not have a fallback path here?

are we not compiling with c11?

compnerd:
There is no guarantee that the libc is going to be complete (e.g. Windows).  Yes, I know that there is a separate Windows path, but this is already a difficult library to port, lets not complicate it further.

AtariDreams:
> Collaborator

Windows is literally the only libc with this issue. I just checked.

AtariDreams:
> There is no guarantee that the libc is going to be complete (e.g. Windows). Yes, I know that there is a separate Windows path, but this is already a difficult library to port, lets not complicate it further.

That's literally the point of this PR. And let's not assume that the code already assumes you are compiling for C11, as there are C99 and C11 features in this codebase like 0 length arrays.

AtariDreams:
@compnerd

AtariDreams:
Also, the PR I did for swift literally only checks for windows, and then the STDC version, not a complete libc. I don't see why we would need to find the one libc that doesn't have this but somehow has posix_memalign

compnerd:
@swift-ci please test

triplef:
FYI for anyone else running into this: this change makes libdispatch require a minimum Android SDK version of 28 (Android 9 Pie) as noted [here](https://github.com/android/ndk/issues/1339#issuecomment-676811636), because earlier versions don’t have `aligned_alloc`.

Building with an earlier `minSdkVersion` results in this error:
```
Linking CXX shared library ../libdispatch.so
ld: error: undefined symbol: aligned_alloc
>>> referenced by io.c:2377
```

finagolfin:
> Is there any chance to add an Android target to CI to be able to spot things like this sooner?

I will look into it. In the meantime, I have submitted a fallback, #811.

triplef:
Thank you @finagolfin! 🙏

PR #806, opened 2023-12-15
Convert deprecated functions to use modern atomic variable handling

AtariDreams:
Replaced deprecated atomic functions with the recommended C11 equivalents.


AtariDreams:
@compnerd Thoughts on this?

compnerd:
I don't see a good reason for any of the changes in `src/allocator.c`.  Switching the Windows path ... I might want to think a bit more on, but I'm not sure that we should change the non-Windows path.

AtariDreams:
@compnerd I just unified the paths so we don't have to worry anymore

PR #807, opened 2023-12-16
Use proper atomics and ref counting

AtariDreams:
Prefer __atomic_compare_exchange_n over __sync_bool_compare_and_swap.

I chose weak because we are looping and reading the value of old_value constantly anyway, so it would be better to have it weak.

Otherwise, it is equivalent to what it was before.


AtariDreams:
@compnerd Can we please do a test

AtariDreams:
@compnerd Can we please do a test?

ktoso:
@swift-ci please test

AtariDreams:
@ktoso My bad! It compiled and ran fine on my machine but I found the mistake and fixed it.

AtariDreams:
@compnerd Can we please do a test?

compnerd:
I'm not sure that this is valuable enough to do really.

rjmccall:
I'm not sure what the point of the atomics change is.  The control flow change is non-obvious and totally unexplained in either comments or in the PR summary.

src/BlocksRuntime/runtime.c
```diff
@@ -32,9 +33,9 @@
 #define __has_builtin(builtin) 0
 #endif
 
-#if __has_builtin(__sync_bool_compare_and_swap)
+#if __has_builtin(__atomic_compare_exchange_n)
 #define OSAtomicCompareAndSwapInt(_Old, _New, _Ptr)                            \
-  __sync_bool_compare_and_swap(_Ptr, _Old, _New)
+  __atomic_compare_exchange_n(_Ptr, &_Old, _New, true, memory_order_release, memory_order_relaxed)
```
rjmccall, line 38: If this function has the semantics of Darwin's `OSAtomicCompareAndSwapInt`, I think both memory orderings can be relaxed, but it may have to be a strong operation (it's not clear from the docs).

If you're going to use the builtin directly, you should use the builtin `__ATOMIC_RELAXED` macro.  If you're going to rely on `<stdatomic.h>`, you should use the standard operation `atomic_compare_exchange_strong_explicit`.

PR #808, opened 2023-12-18
Sonoma merge

AtariDreams:
This merge brings a lot of changes and refinements specifically for Swift, so having them here too is quite beneficial.


src/event/event_windows.c
lxbndr, line 1: #581 seems being reverted by this. That was a nasty hard to find random issue, I'd really like to keep those changes🙏

AtariDreams: I'll do a remerge cross referencing every change

PR #815, opened 2024-02-21
Changes for Musl compatibility

al45tair:
We should be using `<poll.h>` and not `<sys/poll.h>`, and also we need a way to override the Swift library subdirectory name, rather than assuming we can always derive it by lowercasing `CMAKE_SYSTEM_NAME`.


al45tair:
@swift-ci Please test Windows platform

PR #828, opened 2024-05-09
build: android bundles pthread into libc

compnerd:
Add an additional flag to indicate that `-lc` is sufficient for pthread support and that `-lpthread` is not needed. This allows us to build with newer NDKs.


compnerd:
CC: @drodriguez @finagolfin @hyp

compnerd:
@swift-ci please test

PR #832, opened 2024-06-26
[Linux] Enable build IDs.

al45tair:
We should use build IDs on Linux so that we can identify the built artefacts, and also so that we can match them up with debug information should we choose to separate it.

rdar://130582819


al45tair:
@swift-ci Please test Linux platform

al45tair:
@swift-ci Please test Windows platform

al45tair:
@swift-ci Please test Windows platform

PR #838, opened 2024-08-02
Add more comments for #endif in .h file.

lamtrinhdev:
This is a long `.h` file. I would like to are more comment for `#endif`.


lamtrinhdev:
This PR have 2 comments, let close it to move the new PR with single comment.

PR #840, opened 2024-08-08
Disable new `cast-function-type-mismatch`

bnbarham:
Should fix this properly, but for now just disable the warning itself.


bnbarham:
@swift-ci please test

rokhinip (approved):

PR #852, opened 2025-01-20
Handle multibyte paths on Windows

lxbndr:
`DispatchIO` is used in [FoundationNetworking.URLSession](https://github.com/swiftlang/swift-corelibs-foundation/blob/c466923dcaeeb058fc570f0e0cadbb04afb2b5df/Sources/FoundationNetworking/URLSession/BodySource.swift#L171) for a file reading. The path is taken from `URL.withUnsafeFileSystemRepresentation`, which look as expected behavior to me. It is multibyte string in fact, and on the Dispatch side we have to handle this accordingly.

We noted this as a crashes/hangs in our product, happening when we tried to POST files with non-ascii characters in the path. We are running with this patch without any issues for a few months already, so it is kind of tested. We're running on 5.10.1 release branch, but as long as `withUnsafeFileSystemRepresentation` is not changed, everything should be the same.


lxbndr:
cc @compnerd

src/io.c
```diff
@@ -1333,9 +1347,14 @@ _dispatch_fd_entry_guarded_open(dispatch_fd_entry_t fd_entry, const char *path,
 	} else if (oflag & _O_TRUNC) {
 		dwCreationDisposition = TRUNCATE_EXISTING;
 	}
-	return (dispatch_fd_t)CreateFile(path, dwDesiredAccess,
+	int wide_len = MultiByteToWideChar(CP_UTF8, 0, path, -1, NULL, 0);
+	WCHAR *wide_path = malloc(wide_len * sizeof(WCHAR));
+	MultiByteToWideChar(CP_UTF8, 0, path, -1, wide_path, wide_len);
+	dispatch_fd_t fd = (dispatch_fd_t)CreateFileW(wide_path, dwDesiredAccess,
```
compnerd, line 1353: Might be nice to maintain the style for windows and use `wszPath` for the path.

src/io.c
```diff
@@ -1333,9 +1347,14 @@ _dispatch_fd_entry_guarded_open(dispatch_fd_entry_t fd_entry, const char *path,
 	} else if (oflag & _O_TRUNC) {
 		dwCreationDisposition = TRUNCATE_EXISTING;
 	}
-	return (dispatch_fd_t)CreateFile(path, dwDesiredAccess,
+	int wide_len = MultiByteToWideChar(CP_UTF8, 0, path, -1, NULL, 0);
+	WCHAR *wide_path = malloc(wide_len * sizeof(WCHAR));
+	MultiByteToWideChar(CP_UTF8, 0, path, -1, wide_path, wide_len);
```
compnerd, line 1352: Should we extract this into a helper?

src/io.c
lxbndr: Exactly! Something in `wide_path` bothered me and I even wasn't aware what exactly. Now I see (and can't unsee)🙈

src/io.c
```diff
@@ -447,10 +469,23 @@ dispatch_io_create_with_path(dispatch_io_type_t type, const char *path,
...
+		struct _stat st;
+		WCHAR *wszPath = _dispatch_win32_copy_wide_from_utf8(path);
+		if (!wszPath) {
+			err = (int)GetLastError();
+			free(path_data);
+			_dispatch_io_init(channel, NULL, queue, err, cleanup_handler);
+			_dispatch_release(channel);
+			_dispatch_release(queue);
+			return;
+		}
```
compnerd, line 482: How about using `DISPATCH_ASSERT` here and ensuring that we either succeed or abort?

lxbndr: @compnerd Not sure if we want to abort in case of error. Or, perhaps, not on every error. Invalid UTF-8 sequence, as one of failure reasons, is just a bad input. It should not crash an application. Other errors defined for `MultiByteToWideChar` are more like developer errors, and are better candidates for assertion check to catch some unexpected things. Also there is malloc out-of-memory possibility, which probably also should not crash (as I see similar non-crashing check for in the code above).

We can check for actual error code (i.e. make this check a part of `dispatch_assert` condition), but it doesn't make this code simpler. I am assuming that your main point is "code become too complicated". At least, it mostly concerns me here.

What do you think? Does whitelisting some error codes sound ok?

compnerd: Yes, it is about the complexity. Alternatively, is there a way to have a common exit path on the error?

lxbndr: @compnerd I have had an idea, which probably is naturally the most correct way to handle this. We can change the type of `path` field in the `dispatch_io_path_data_t` structure for the platform. This will allow us to convert path once, and then use stored value as is. Some (and this in particular) code from the patch will become unnecessary, but some other places would need tweaks, because there are assumptions about structure size in bytes.

In general there seem to be not much usage of this field, so I'd like to try and look what how it goes.

lxbndr:
Reworked version
- some type/function aliasing for Win/Non-Win platforms
- `path` in `dispatch_io_path_data_s` is now `WCHAR[]` on Windows
- one single conversion from utf-8 to mbcs
- general flow is same, no new edge cases
- most significant changes are in `dispatch_io_create_with_path` part related to path data initialization

src/io.c
```diff
@@ -407,9 +407,13 @@ dispatch_io_create_f(dispatch_io_type_t type, dispatch_fd_t fd,
 }
 
 #if defined(_WIN32)
-#define _is_separator(ch) ((ch) == '/' || (ch) == '\\')
+#define _is_separator(ch) ((ch) == L'/' || (ch) == L'\\')
+#define _dispatch_stat _wstat
```
compnerd, line 411: The only thing that I wonder about here is if we should be using `_wstat` or `_wstati64` to get 64-bit times and lengths.

compnerd (approved):

compnerd:

src/io.c
lxbndr: Not sure if I'm following🤔 I see no usage of time/size fields. But as a future-proof change it seems reasonable. That would be a simple drop-in replacement, as everything compiles without warnings and works same way. Should I push it?

compnerd: Yes please! The idea is to be prepared in case we start using that.

lxbndr: Done

compnerd (approved):

compnerd:
@swift-ci please test

compnerd:

PR #854, opened 2025-01-24
Switch Win32 pipes to PIPE_WAIT with sentinel bufsize

z2oh:
Fixes https://github.com/swiftlang/swift-corelibs-libdispatch/issues/820 . There is a lot of useful context in this issue, which I will partially reproduce below.

I have an alternate PR that fixes this problem in a less fundamental way, but without a small requirements change to libdispatch which I describe below: https://github.com/swiftlang/swift-corelibs-libdispatch/pull/853.

Problem description, copied from the aforementioned PR:

> * In https://github.com/swiftlang/swift-corelibs-libdispatch/pull/781, the pipe was changed from `PIPE_WAIT` to `PIPE_NOWAIT` to achieve POSIX-like `O_NONBLOCK` semantics, as attempting to write a large buffer would cause the dispatch queue thread to block long enough to hit a time out.
> * The Windows pipe synchronization thread, `_dispatch_pipe_monitor_thread`, uses a blocking 0-byte read as a synchronization mechanism to signal to the actual reader threads that there is data waiting in the pipe. Of course, for this to work, the read must actually be blocking. With a `PIPE_NOWAIT` pipe, the Read will not block, causing the thread to spin endlessly and constantly wake the actual reader thread to perform 0 byte reads. This spinning thread is the root source of the problem in this issue.
> 
> Switching the pipe back to `PIPE_WAIT` potentially resurfaces the blocking writes problem, however I think that https://github.com/swiftlang/swift-corelibs-libdispatch/pull/796 (which was a follow-up PR to fix a problem with the `PIPE_NOWAIT` implementation) also resolves the blocking write problem _in practice_ but not necessarily in all cases.

The real issue here is the inability to distinguish on the write side of a pipe between the following cases:
* The pipe's output buffer is full (we do not want to write as calling WriteFile will block)
* The read side of the pipe has requested a full buffer's worth of data (we _do_ want to write because we need to make progress and there is a reader waiting for it)

This PR implements the same switch back to `PIPE_WAIT` as in https://github.com/swiftlang/swift-corelibs-libdispatch/pull/853, but uses a trick to enable `WriteQuotaAvailable == 1` to be used as a sentinel value to distinguish the "output buffer is full" case. By always leaving a free byte of space in the buffer when writing, the only way that `WriteQuotaAvailable == 0` can happen is if a reader has requested a full buffer's worth of data.

This means that when we see `WriteQuotaAvailable == 1`, we know the output buffer is "full" and not attempt to perform a blocking write.

The downside to this approach is that we can no longer serially write-then-read a full buffer's worth of data, because the write will not finish writing the last byte until reading has commenced. I don't see this scenario ever arising in practice: if you are trying to move data serially from one buffer to another on the same thread, there are better ways to do that than through a libdispatch pipe. But regardless, this requirement is implicit due to the serial nature of the `dispatch_io_pipe` tests. If if is acceptable to relax this requirement from "writes of `<=bufSize` are guaranteed to complete in the absence of a reader" to "writes of `<bufSize` are guaranteed to complete in the absence of a reader" then I think this solution is more robust than just https://github.com/swiftlang/swift-corelibs-libdispatch/pull/853.

It's entirely possible I'm missing some good reason why this requirement is there though, in which case I think https://github.com/swiftlang/swift-corelibs-libdispatch/pull/853 should be acceptable.

cc @compnerd

### DIFF
--- a/src/event/event_windows.c
+++ b/src/event/event_windows.c
@@ -277,6 +277,9 @@ _dispatch_pipe_monitor_thread(void *context)
 		char cBuffer[1];
 		DWORD dwNumberOfBytesTransferred;
 		OVERLAPPED ov = {0};
+		// Block on a 0-byte read; this will only resume when data is
+		// available in the pipe. The pipe must be PIPE_WAIT or this thread
+		// will spin.
 		BOOL bSuccess = ReadFile(hPipe, cBuffer, /* nNumberOfBytesToRead */ 0,
 				&dwNumberOfBytesTransferred, &ov);
 		DWORD dwBytesAvailable;
--- a/src/io.c
+++ b/src/io.c
@@ -1437,20 +1437,20 @@ _dispatch_fd_entry_create_with_fd(dispatch_fd_t fd, uintptr_t hash)
 				int result = ioctlsocket((SOCKET)fd, (long)FIONBIO, &value);
 				(void)dispatch_assume_zero(result);
 			} else {
-				// Try to make writing nonblocking, although pipes not coming
-				// from Foundation.Pipe may not have FILE_WRITE_ATTRIBUTES.
+				// The _dispatch_pipe_monitor_thread expects pipes to be
+				// PIPE_WAIT and exploits this assumption by using a blocking
+				// 0-byte read as a synchronization mechanism.
 				DWORD dwPipeMode = 0;
 				if (GetNamedPipeHandleState((HANDLE)fd, &dwPipeMode, NULL,
-						NULL, NULL, NULL, 0) && !(dwPipeMode & PIPE_NOWAIT)) {
-					dwPipeMode |= PIPE_NOWAIT;
+						NULL, NULL, NULL, 0) && !(dwPipeMode & PIPE_WAIT)) {
+					dwPipeMode |= PIPE_WAIT;
 					if (!SetNamedPipeHandleState((HANDLE)fd, &dwPipeMode,
 							NULL, NULL)) {
-						// We may end up blocking on subsequent writes, but we
-						// don't have a good alternative.
-						// The WriteQuotaAvailable from NtQueryInformationFile
-						// erroneously returns 0 when there is a blocking read
-						// on the other end of the pipe.
-						_dispatch_fd_entry_debug("failed to set PIPE_NOWAIT",
+						// If setting the pipe to PIPE_WAIT fails, the
+						// monitoring thread will spin constantly, saturating
+						// a core, which is undesirable but non-fatal.
+						// The semantics will still be correct in this case.
+						_dispatch_fd_entry_debug("failed to set PIPE_WAIT",
 								fd_entry);
 					}
 				}
--- a/src/io.c
+++ b/src/io.c
@@ -2518,13 +2518,40 @@ syscall:
 				NTSTATUS status = _dispatch_NtQueryInformationFile(hFile,
 						&iosb, &fpli, sizeof(fpli), FilePipeLocalInformation);
 				if (NT_SUCCESS(status)) {
-					// WriteQuotaAvailable is unreliable in the presence
-					// of a blocking reader, when it can return zero, so only
-					// account for it otherwise
-					if (fpli.WriteQuotaAvailable > 0) {
-						len = MIN(len, fpli.WriteQuotaAvailable);
+					// WriteQuotaAvailable is the free space in the output buffer
+					// that has not already been reserved for reading. In other words,
+					// WriteQuotaAvailable =
+					//    OutboundQuota - WriteQuotaUsed - QueuedReadSize.
+					// It is not documented that QueuedReadSize is part of this
+					// calculation, but this behavior has been observed experimentally.
+					// Unfortunately, this means that it is not possible to distinguish
+					// between a full output buffer and a reader blocked waiting for a
+					// full buffer's worth of data. This is a problem because if the
+					// output buffer is full and no reader is waiting for data, then
+					// attempting to write to the buffer of a PIPE_WAIT, non-
+					// overlapped I/O pipe will block the dispatch queue thread.
+					//
+					// In order to work around this idiosyncrasy, we bound the size of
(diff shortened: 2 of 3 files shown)


tests/dispatch_io_pipe.c
```diff
@@ -404,7 +404,14 @@ test_dispatch_write(int kind, int delay)
 	dispatch_group_t g = dispatch_group_create();
 	dispatch_group_enter(g);
 
-	const size_t bufsize = test_get_pipe_buffer_size(kind);
+	// We must subtract 1 from the bufsize because the write operation
+	// uses 1 byte of the buffer as a sentinel value to ensure that a
+	// reader is making progress. Because these tests operate serially,
```
tristanlabelle, line 409: It might be easier to follow if we describe the libdispatch behavior rather than the rationale of its implementation:

> The libdispatch implementation writes at most `bufsize-1` before requiring the reader to start draining the pipe.

src/io.c
```diff
@@ -1437,20 +1437,20 @@ _dispatch_fd_entry_create_with_fd(dispatch_fd_t fd, uintptr_t hash)
...
-						// don't have a good alternative.
-						// The WriteQuotaAvailable from NtQueryInformationFile
-						// erroneously returns 0 when there is a blocking read
-						// on the other end of the pipe.
-						_dispatch_fd_entry_debug("failed to set PIPE_NOWAIT",
+						// If setting the pipe to PIPE_WAIT fails, the
+						// monitoring thread will spin constantly, saturating
+						// a core, which is undesirable but non-fatal.
+						// The semantics will still be correct in this case.
+						_dispatch_fd_entry_debug("failed to set PIPE_WAIT",
```
tristanlabelle, line 1453: See comments from https://github.com/swiftlang/swift-corelibs-libdispatch/pull/853

tristanlabelle (approved):
I'm comfortable with this fix

src/event/event_windows.c
```diff
@@ -277,6 +277,9 @@ _dispatch_pipe_monitor_thread(void *context)
 		char cBuffer[1];
 		DWORD dwNumberOfBytesTransferred;
 		OVERLAPPED ov = {0};
+		// Block on a 0-byte read; this will only resume when data is
+		// available in the pipe. The pipe must be PIPE_WAIT or this thread
+		// will spin.
 		BOOL bSuccess = ReadFile(hPipe, cBuffer, /* nNumberOfBytesToRead */ 0,
```
maruel, line 283: Why not do a WaitForSingleObjectEx()? In particular pay attention to the bAlertable condition.

IMHO, I don't see the need to use a thread, WaitForMultipleObjectsEx() would do just fine and be more performant.

Note that I don't have a good understanding of this code base so my humble opinion may be wrong but I have some experience with win32 pipes.

src/io.c
```diff
@@ -2518,13 +2518,38 @@ _dispatch_operation_perform(dispatch_operation_t op)
...
+					// WriteQuotaAvailable =
+					//    OutboundQuota - WriteQuotaUsed - QueuedReadSize.
+					// Unfortunately, this means that it is not possible to distinguish
+					// between a full output buffer and a reader blocked waiting for a
+					// full buffer's worth of data. This is a problem because if the
+					// output buffer is full and no reader is waiting for data, then
+					// attempting to write to the buffer of a PIPE_WAIT, non-
+					// overlapped I/O pipe will block the dispatch queue thread.
+					//
+					// In order to work around this idiosynchrasy, we bound the size of
```
maruel, line 2532: idiosyncrasy (or idiocracy, like the movie)

maruel (commented):
Thanks for looking at this! Just a comment, no opinion on the change.

src/event/event_windows.c
z2oh: I'm not sure I follow. I looked through the docs of [WaitForSingleObjectEx](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitforsingleobjectex) and I don't see where it applies here. You can't wait on a pipe, no?

maruel: Effectively. The idea is to wait on the events that are connected to the pending overlappedio.

maruel: Expanding; https://learn.microsoft.com/en-us/windows/win32/devio/overlapped-operations
> The OVERLAPPED structure must contain a handle to a manual-reset (not an auto-reset) event object

https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-overlapped
especially the section about hEvent.

z2oh: In general we cannot rely on Overlapped I/O here (which would simplify many things). Pipes are created externally and their handles are passed into libdispatch, and there's no way to make a pipe `FILE_FLAG_OVERLAPPED` after the fact (at least, none that I could find, and I looked for this).

We also need to support anonymous pipes, which [do not support Overlapped I/O](https://learn.microsoft.com/en-us/windows/win32/ipc/anonymous-pipe-operations?redirectedfrom=MSDN). This requirement is embedded in the libdispatch tests, which use [anonymous pipes](https://github.com/swiftlang/swift-corelibs-libdispatch/blob/b2a1fad6b49a9394cf864b87ab0034ea8300e2de/tests/dispatch_io_pipe.c#L153).

It could be I'm misunderstanding something about your proposal though. After discounting Overlapped I/O because the requirements mentioned above, I stopped looking into that solution space, so am less familiar with the APIs you are referencing.

maruel: Oh ok that's a bummer. Sorry I didn't mean to send you off track. From https://learn.microsoft.com/en-us/windows/win32/ipc/anonymous-pipe-operations:

> In addition, the lpOverlapped parameter of ReadFile and WriteFile is *ignored* when these functions are used with anonymous pipes.

Emphasis mine. That's why the code works. I'd recommend to pass NULL instead or &ov to avoid confusion. Other than that, lgtm.

compnerd (approved):

compnerd:
@swift-ci please test

compnerd:

PR #856, opened 2025-01-31
[Don't merge] Lower timeout to force failure to check for log output

z2oh:
This is just to help check if a change to Swift CI will report verbose libdispatch test failures


weliveindetail:
That worked! I think we don't need this anymore.

PR #857, opened 2025-01-31
swift: use `internal` instead of `@_implementationOnly`

compnerd:
Prefer to use the new `internal import` functionality over the reserved `@_implementationOnly` attribute.


compnerd:
@swift-ci please test

compnerd:
Ugh, this will require distributing the module as `internal` still requires the module to be present.

PR #858, opened 2025-02-08
🍒[6.1] Handle multibyte path strings on Windows

lxbndr:
Cherry-pick of #852


lxbndr:
cc @compnerd

compnerd:
@swift-ci please test

rokhinip (approved):

compnerd (approved):
I think it is fine for this particular case. But in general, we should be including the CCC for the backports at this point.

PR #861, opened 2025-03-17
[FreeBSD] support FreeBSD

michael-yuji:
This patch adds FreeBSD support to libdispatch.

This patch is partly based on #561 with bug fixes and uses the pipe based mechanism based on #559 to bridge with CFRunLoop, since part of the CFRunLoop changes already landed in CoreFoundation.

In addition to above changes, also implements workqueue and limited DispatchSource support.


private/private.h
```diff
@@ -189,8 +189,10 @@ void _dispatch_prohibit_transition_to_multithreaded(bool prohibit);
 
 #if TARGET_OS_MAC
 typedef mach_port_t dispatch_runloop_handle_t;
-#elif defined(__linux__) || defined(__FreeBSD__)
+#elif defined(__linux__)
 typedef int dispatch_runloop_handle_t;
+#elif defined(__FreeBSD__)
+typedef uint64_t dispatch_runloop_handle_t;
```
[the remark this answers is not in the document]
michael-yuji: this will be an integer shaped value. This is used to adopt the `pipe` based mechanism to bridge between Foundation https://github.com/swiftlang/swift-corelibs-foundation/pull/3004

cmake/modules/SwiftSupport.cmake
```diff
@@ -27,6 +27,8 @@ function(get_swift_host_arch result_var_name)
     set("${result_var_name}" "armv7" PARENT_SCOPE)
   elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "armv7l")
     set("${result_var_name}" "armv7" PARENT_SCOPE)
+  elseif("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "amd64")
+    set("${result_var_name}" "x86_64" PARENT_SCOPE)
```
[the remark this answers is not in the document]
etcwilde: https://github.com/swiftlang/swift-corelibs-libdispatch/pull/875 should have made this change unnecessary. It'll just grab the right value from the compiler.

private/private.h
```diff
@@ -189,9 +189,11 @@ void _dispatch_prohibit_transition_to_multithreaded(bool prohibit);
 
 #if TARGET_OS_MAC
 typedef mach_port_t dispatch_runloop_handle_t;
-#elif defined(__linux__) || defined(__FreeBSD__)
+#elif defined(__linux__)
 typedef int dispatch_runloop_handle_t;
-#elif defined(__unix__) && !defined(__linux__) && !defined(__FreeBSD__)
+#elif defined(__FreeBSD__)
+typedef uint64_t dispatch_runloop_handle_t;
```
etcwilde, line 195: Do we need this case? The case below looks like it's is typedef'ing the same things and the `defined(__unix__)` should capture on FreeBSD as well.

src/queue.c
```diff
@@ -6490,7 +6490,9 @@ _dispatch_runloop_queue_get_handle(dispatch_lane_t dq)
 #elif defined(__linux__)
 	// decode: 0 is a valid fd, so offset by 1 to distinguish from NULL
 	return ((dispatch_runloop_handle_t)(uintptr_t)dq->do_ctxt) - 1;
-#elif defined(__unix__) && !defined(__linux__)
+#elif defined(__FreeBSD__)
+	return ((dispatch_runloop_handle_t)(uintptr_t)dq->do_ctxt);
```
etcwilde, line 6494: Do we need this? It looks like it's the same as the `defined(__unix__)` case below, which should capture on FreeBSD.

src/queue.c
```diff
@@ -6509,6 +6511,8 @@ _dispatch_runloop_queue_set_handle(dispatch_lane_t dq,
 #elif defined(__linux__)
 	// encode: 0 is a valid fd, so offset by 1 to distinguish from NULL
 	dq->do_ctxt = (void *)(uintptr_t)(handle + 1);
+#elif defined(__FreeBSD__)
```
etcwilde, line 6514: Do we need this? It looks like it's the same as the `defined(__unix__)` case below, which should capture on FreeBSD.

etcwilde (commented):
A few places that look like it's duplicating the `__unix__` case. Otherwise I think this looks pretty good.

etcwilde:
@swift-ci please test

src/shims/lock.c
```diff
@@ -54,9 +54,20 @@ _dispatch_thread_switch(dispatch_lock value, dispatch_lock_options_t flags,
...
+#if !HAVE_UL_UNFAIR_LOCK
+DISPATCH_ALWAYS_INLINE
+static inline void
+_dispatch_thread_switch(dispatch_lock value, dispatch_lock_options_t flags,
+		uint32_t timeout)
+{
+  (void)value;
+  (void)flags;
+  (void)timeout;
+	sched_yield();
```
CodingCarpincho, line 67: This is a nitpick, but the indentation is not consistent here.

src/shims/lock.c
```diff
@@ -539,6 +550,16 @@ _dispatch_wait_on_address(uint32_t volatile *_address, uint32_t value,
 			? INFINITE : ((nsecs + 1000000) / 1000000);
 	if (dwMilliseconds == 0) return ETIMEDOUT;
 	return WaitOnAddress(address, &value, sizeof(value), dwMilliseconds) == TRUE;
+#elif defined(__FreeBSD__)
+  (void)flags;
+  if (nsecs != DISPATCH_TIME_FOREVER) {
+		struct timespec ts = {
```
CodingCarpincho, line 556: There seems to be more indentation issues here.

CodingCarpincho (approved):

michael-yuji:
@swift-ci please test

src/event/workqueue.c
```diff
@@ -293,6 +293,52 @@ _dispatch_workq_count_runnable_workers(dispatch_workq_monitor_t mon)
...
+	int mib[] = {CTL_KERN, KERN_PROC, KERN_PROC_PID | KERN_PROC_INC_THREAD, (int)getpid()};
+
+	// get size we need
+	if (sysctl(mib, 4, NULL, &size, NULL, 0) < 0) {
+		_dispatch_debug("workq: failed to get size for kinfo_proc[] from sysctll");
+		return;
+	}
+
+	// only care about up to WORKQ_MAX_TRACKED_TIDS threads
+	size = size > sizeof(kp) ? sizeof(kp) : size;
```
compnerd, line 316: Why not write this using `MIN`?

```suggestion
	size = MIN(sizeof(kp), size);
```

src/swift/Source.swift
```diff
@@ -124,8 +124,17 @@ extension DispatchSource {
 		public static let exit = ProcessEvent(rawValue: 0x80000000)
 		public static let fork = ProcessEvent(rawValue: 0x40000000)
 		public static let exec = ProcessEvent(rawValue: 0x20000000)
+#if os(FreeBSD)
+		public static let track = ProcessEvent(rawValue: 0x00000001)
```
compnerd, line 128: Why the different name? Should we not be using the name names to maintain source consistency?

michael-yuji: This is because `NOTE_TRACK` is only available on FreeBSD but not Darwin

compnerd: Sure, having an extended FreeBSD only option is fine, but this does remove the `signal` note doesn't it? I take it that `NOTE_TRACK` is meant to replace that? If so, why not keep the name incongruous with the underlying note to allow easier porting?

michael-yuji: Sorry, I should've add more context in my previous reply. `NOTE_SIGNAL` on Darwin and `NOTE_TRACK` on FreeBSD are not interchangeable and each are platform specific.

`NOTE_SIGNAL` on Darwin allow users to check if the process was sent a signal,

on the other hand, `NOTE_TRACK` on FreeBSD cause kqueue to follow the target process across fork and automatically register events with same monitor. For example, `NOTE_TRACK | NOTE_EXIT` on a pid cause exit event of the tracing process, and all its decedents to be reported.

src/event/workqueue.c
```diff
@@ -293,6 +293,52 @@ _dispatch_workq_count_runnable_workers(dispatch_workq_monitor_t mon)
...
+static void
+_dispatch_workq_count_runnable_workers(dispatch_workq_monitor_t mon)
+{
+	struct kinfo_proc kp[WORKQ_MAX_TRACKED_TIDS];
+	size_t size;
+	int count, runners = 0;
+	int mib[] = {CTL_KERN, KERN_PROC, KERN_PROC_PID | KERN_PROC_INC_THREAD, (int)getpid()};
+
+	// get size we need
+	if (sysctl(mib, 4, NULL, &size, NULL, 0) < 0) {
```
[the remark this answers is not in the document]
michael-yuji: I don't really think it's a good idea to introduce a macro, especially just for one place.

Although it makes it more readable, but I think it is reasonable to expect the next person who need to change this line are already familiar with the correct usage of `sysctl`

etcwilde:
It's not in this PR, but I just came across an additional build failure in the tests:
https://github.com/swiftlang/swift-corelibs-libdispatch/blob/beb137b99c92c54ae6903483f40dd1a3d3b61008/tests/dispatch_workqueue.c#L33-L47

I think we need to adjust the macro to `#if defined(__linux__) || defined(__OpenBSD__) || defined(__FreeBSD__)`.
I know that FreeBSD has `sysctlbyname`, but `hw.activecpu` returns the physical core count, while `dwNumberOfProcessors` is the logical core count, so `sysconf(_SC_NPROCESSORS_ONLN)` seems to be more consistent. Up to you if you want to add that in this PR or not.

etcwilde:
@swift-ci please test

etcwilde:
@swift-ci please test Windows

michael-yuji:
> It's not in this PR, but I just came across an additional build failure in the tests:
> 
> https://github.com/swiftlang/swift-corelibs-libdispatch/blob/beb137b99c92c54ae6903483f40dd1a3d3b61008/tests/dispatch_workqueue.c#L33-L47
> 
> I think we need to adjust the macro to `#if defined(__linux__) || defined(__OpenBSD__) || defined(__FreeBSD__)`. I know that FreeBSD has `sysctlbyname`, but `hw.activecpu` returns the physical core count, while `dwNumberOfProcessors` is the logical core count, so `sysconf(_SC_NPROCESSORS_ONLN)` seems to be more consistent. Up to you if you want to add that in this PR or not.
> 
> edit: Realizing I didn't add the actual failure. The error is that we're hitting the `sysctlbyname` call without including `sys/sysctl.h`.

Will be addressing this in the next PR

PR #865, opened 2025-03-25
build: remove cross-compilation support

compnerd:
These can be controlled by the user external to the project. Remove this to simplify the rules.


compnerd:
@swift-ci please test

etcwilde (approved):

Issue #820, opened 2024-03-06
DispatchIO.read spinning while pipe is open on Windows


DimaKozhinov:
The sourcekit-lsp.exe consumes ~24% of CPU power all the time, even when idle. This bug does not appear under Linux, and does not appear in Swift versions earlier than v5.9. It is still present in v5.10. I tested this under VirtualBox, my virtual machine has 4 virtual processors. The host machine has 13th Gen Intel Core i7 processor, so ~24% load seems like a lot of mysterious calculations.

How to reproduce:

1. Install Swift version 5.9 .. 5.10. Install Microsoft VS Code and Swift extension for it (v1.8.1 from Swift Server Work Group).
3. Create a test Swift package:
    mkdir test
    cd test
    swift package init
4. Open the test folder in VS Code. (File | Open Folder...). At this point CPU usage increases, you don't even need to edit any source file. On my (usually silent) PC fan immediately starts rotating much faster, producing loud noise.
5. Check CPU usage by the sourcekit-lsp.exe: Press Ctrl-Shift-Esc, in the process list find the Visual Studio Code, click ">" next to it to see its sub-processes, and see high CPU usage by sourcekit-lsp .
6. Return to step 1 and this time install Swift version 5.8.1 or earlier. Repeat step 2 (create Swift package) because earlier version of Swift won't open a package created with a later version. Repeat all next steps and see ~0% CPU usage by sourcekit-lsp.exe .

This bug has nothing to do with VS Code. I found this bug when using my own app that runs sourcekit-lsp.exe, and not VS Code.

ahoppen:
@tristanlabelle I have reached that stage. If you could investigate it, I would really appreciate it.

tristanlabelle:
Hi @ahoppen , we'll look into it. Currently our priorities in this area are roughly:
1. SourceKit-LSP file locking issue
2. https://github.com/apple/sourcekit-lsp/issues/1139
3. This

ahoppen:
It looks like we are spending the time in `_dispatch_event_loop_leave_immediate`. @tristanlabelle Have you seen this before in your previous fixes of hangs in libdispatch?

<img width="742" alt="Screenshot 2024-07-30 at 19 21 13" src="https://github.com/user-attachments/assets/6e0b8b52-10be-4a6c-9574-4b0c3dbdec67">

tristanlabelle:
[...]

PR #866, opened 2025-04-16
IO: switch Win32 pipes to `PIPE_WAIT` and use a sentinel byte

compnerd:
Cherry-pick changes from main to use a `PIPE_WAIT`-mode pipe on Windows. Additionally, steal a byte from the buffer to synchronise on as a sentinel byte to check if we are waiting on a client to read from the buffer.

### DIFF
--- a/src/event/event_windows.c
+++ b/src/event/event_windows.c
@@ -277,6 +277,9 @@ _dispatch_pipe_monitor_thread(void *context)
 		char cBuffer[1];
 		DWORD dwNumberOfBytesTransferred;
 		OVERLAPPED ov = {0};
+		// Block on a 0-byte read; this will only resume when data is
+		// available in the pipe. The pipe must be PIPE_WAIT or this thread
+		// will spin.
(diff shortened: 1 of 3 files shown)


compnerd:
@swift-ci please test

compnerd:
- **Explanation**:
    Switch pipes on Windows to `PIPE_WAIT` mode to get a blocking pipe semantics. This allows the IOCP loop in the IO monitor to continue to make progress. We take the last byte of the buffer to synchronise the read and write buffer for the pipe.
  - **Scope**:
    This change impacts only pipes on Windows.
  - **Issues**:
    #820 
  - **Original PRs**:
    #854
  - **Risk**:
    In theory, this change can cause dispatch clients to possibly hang in the scenario that a pipe is given to dispatch with a writer that fills the buffer and no reader that will ever drain the buffer.
  - **Testing**:
    LSP was used with this change and was verified to no longer cause a high CPU utilisation when idle.
  - **Reviewers**:
    @compnerd @rokhinip

compnerd:

PR #871, opened 2025-05-11
build: install cmake config

ilya-fedin:
This makes it easier to consume dispatch via package managers such as vcpkg or Linux system ones


cmake/modules/CMakeLists.txt
```diff
@@ -1,7 +1,23 @@
...
+  @ONLY)
+
+if(LibRT_FOUND)
+  file(COPY FindLibRT.cmake DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+endif()
 
 get_property(DISPATCH_EXPORTS GLOBAL PROPERTY DISPATCH_EXPORTS)
 export(TARGETS ${DISPATCH_EXPORTS} FILE ${DISPATCH_EXPORTS_FILE})
+
+if(NOT ENABLE_SWIFT)
```
compnerd, line 14: Why the `NOT ENABLE_SWIFT` condition for this?

ilya-fedin: I've adapted only the native library part. I don't have swift, I use dispatch as a native library in a C++ project and a cmake config for it would make consumption way easier.

PR #872, opened 2025-05-11
dispatch: split `DISPATCH_EXPORT` and prepare for static linking

compnerd:
This change accomplishes two items:
1. splits out a new `DISPATCH_EXTERN` macro to decorate with `extern` or `extern "C"` based on the compilation mode (C vs C++).
2. enables a new configuration macro (`dispatch_STATIC`) that allows building with libdispatch for static linking. The default remains dynamic linking.

The motivation for `DISPATCH_EXTERN` is primarily being able to concisely define `DISPATCH_EXPORT`. This follows the naming scheme that most other Apple frameworks use (e.g. UIKit, Foundation, etc).

The change to introduce support for static linking is motivated by Windows. With the gradual roll out of static linking of the runtime, it would be convenient to be able to statically link dispatch into a fully sealed Swift program. Doing so requires building libdispatch without the `__declspec(dllexport)` and `__declspec(dllimport)` attributes on Windows. Similarly, it would be unfortunate to have dispatch's ABI be subsumed by a client library and have it participate in dynamic linking.

The user is responsible for defining `dispatch_STATIC` when using a static copy of libdispatch. This allows the default to remain dynamic linking (which is the preferred style on Darwin and Windows).

### DIFF
--- a/dispatch/base.h
+++ b/dispatch/base.h
@@ -186,16 +186,24 @@
 # endif
 #endif
 
-#if defined(_WIN32)
-#if defined(__DISPATCH_BUILDING_DISPATCH__)
-#define DISPATCH_EXPORT DISPATCH_EXTERN __declspec(dllexport)
+#if defined(dispatch_STATIC)
+# if __GNUC__
+#   define DISPATCH_EXPORT DISPATCH_EXTERN __attribute__((__visibility__("hidden")))
+# else
+#   define DISPATCH_EXPORT DISPATCH_EXTERN
+# endif
 #else
-#define DISPATCH_EXPORT DISPATCH_EXTERN __declspec(dllimport)
-#endif
-#elif __GNUC__
-#define DISPATCH_EXPORT DISPATCH_EXTERN __attribute__((visibility("default")))
-#else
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

dispatch/base.h
```diff
@@ -178,24 +178,32 @@
...
+#   define DISPATCH_EXTERN extern "C"
+# else
+#   define DISPATCH_EXTERN extern
+# endif
 #endif
-#elif __GNUC__
-#define DISPATCH_EXPORT extern __attribute__((visibility("default")))
+
+#if defined(dispatch_STATIC)
+# if defined(__GNUC__)
```
etcwilde, line 190: We're trying to push away from mixing `#if` and `#ifdef` checks on a macro (unless it's actually desired) since it's confusing to have someone pass `-DFOO=0` and still have the code be treated like it's enabled.
```suggestion
# if __GNUC__
```

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:

PR #877, opened 2025-05-30
[Tests] Fix Dispatch tests to work in container environments.

al45tair:
We need to use `pthread_getaffinity_np()` to get the active core count; using `sysconf()` will tell us about the underlying machine, which may have many more cores than the container in which we are executing.

rdar://152301793


al45tair:
@swift-ci Please test

PR #879, opened 2025-06-06
Set the FreeBSD clock types

etcwilde:
Dispatch passes everything through in nanoseconds, but the default time on FreeBSD is in milliseconds, resulting in delays taking a million times longer than expected. This stands out most with `Task.sleep`, where a 2 second sleep would take hours.

By setting the EVFILT_TIMER filter flag to use nanoseconds, we get the expected delay.

Also sets up the absolute time flag. The spelling on FreeBSD is `NOTE_ABSTIME` instead of `NOTE_ABSOLUTE`.


CodingCarpincho (approved):

akmorrison:
Could you check that this runs in approximately the right amount of time under all three clocks? I'm not 100% sure how monotonic time and wall time are computed on freebsd, but there might be other bugs lurking there.

Something like:
```
dispatch_queue_t queue = dispatch_get_global_queue(0, 0);

dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 2 * NSEC_PER_SEC), queue, ^{
    printf("Ran after 2s of uptime clock\n");
});
dispatch_after(dispatch_time(DISPATCH_WALLTIME_NOW, 3 * NSEC_PER_SEC), queue, ^{
    printf("Ran after 3s of walltime clock\n");
});
dispatch_after(dispatch_time(DISPATCH_MONOTONICTIME_NOW, 4 * NSEC_PER_SEC), queue, ^{
    printf("Ran after 4s of monotonictime clock\n");
});
```

etcwilde:
```c
#include "dispatch/dispatch.h"
#include "dispatch/private.h"

#include <stdio.h>
#include <stdlib.h>

int main(int argc, char** argv) {
        dispatch_queue_t queue = dispatch_get_global_queue(0, 0);

        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, 2 * NSEC_PER_SEC), queue, ^{
                        printf("Ran after 2s of uptime clock\n");
        });
        dispatch_after(dispatch_time(DISPATCH_WALLTIME_NOW, 3 * NSEC_PER_SEC), queue, ^{
                        printf("Ran after 3s of walltime clock\n");
        });
        dispatch_after(dispatch_time(DISPATCH_MONOTONICTIME_NOW, 4 * NSEC_PER_SEC), queue, ^{
                        printf("Ran after 4s of monotonictime clock\n");
        });
        dispatch_after(dispatch_time(DISPATCH_MONOTONICTIME_NOW, 5 * NSEC_PER_SEC), queue, ^{
                printf("Exiting after 5s of monotonic clock\n");
                exit(0);
        });
        dispatch_main();
}
```

```
ewilde@Taquito /tmp %  time ./a.out 
Ran after 2s of uptime clock
Ran after 3s of walltime clock
Ran after 4s of monotonictime clock
Exiting after 5s of monotonic clock
./a.out  0.75s user 4.15s system 98% cpu 5.003 total
```

Seems to be working as I would expect.

akmorrison (approved):

michael-yuji:
cc @3405691582 as OpenBSD might need this too

3405691582:
Note I fudged around this here (l2406), not knowing about these defines. OpenBSD probably needs this too, but would have to undo this fudge factor. I can't test the change for another week or two though.  https://github.com/swiftlang/swift-corelibs-libdispatch/commit/c4814ec404e0bbb36080abf31c0eedd9f6fff238

etcwilde:
Ah yeah, I saw that cutout while tracing through the code and wondered what that was about. I'll merge this once it passes and when you have a chance can put up the PR to change the clocks once you have a chance to try it.

PR #880, opened 2025-06-09
Silence signedness change through implicit conversion error

AnthonyLatsis:
The error pops up when using the rebranch Clang (stable/20250402):

```
/home/build-user/swift-corelibs-libdispatch/src/event/event_epoll.c:92:27: error: implicit conversion changes signedness: 'int' to 'uint32_t' (aka 'unsigned int') [-Werror,-Wsign-conversion]
92 |         return dmn->dmn_events & ~dmn->dmn_disarmed_events;
   |                                ~ ^~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/src/event/event_epoll.c
+++ b/src/event/event_epoll.c
@@ -89,7 +89,9 @@ DISPATCH_ALWAYS_INLINE
 static inline uint32_t
 _dispatch_muxnote_armed_events(dispatch_muxnote_t dmn)
 {
-	return dmn->dmn_events & ~dmn->dmn_disarmed_events;
+  uint32_t events = dmn->dmn_events;
+  uint16_t disarmed_events = dmn->dmn_disarmed_events;
+  return events & ~(uint32_t)disarmed_events;
 }
 
 DISPATCH_ALWAYS_INLINE


src/event/event_epoll.c
```diff
@@ -89,7 +89,7 @@ DISPATCH_ALWAYS_INLINE
 static inline uint32_t
 _dispatch_muxnote_armed_events(dispatch_muxnote_t dmn)
 {
-	return dmn->dmn_events & ~dmn->dmn_disarmed_events;
+	return dmn->dmn_events & ~(uint32_t)(dmn->dmn_disarmed_events);
```
compnerd, line 92: I think that casting to `uint16_t` is better as that is the declaration. That should avoid any unintentional zexting.

AnthonyLatsis: The error says `'int' to 'uint32_t'`. I’m curious where the `int` comes from. If `int` is what the selected negation returns, then `(uint16_t)~dmn->dmn_disarmed_events` seems like it would cause a double conversion, whereas the current change appears to make it select a negation that returns an `uint32_t`.

AnthonyLatsis: Switched to `decltype(dmn->dmn_events)` after discussing offline with Ben.

compnerd (approved):

compnerd:
@swift-ci please test

bnbarham:
Oh, hadn't updated yet so this is just going to fail again 😅

bnbarham (approved):

bnbarham:
@swift-ci please test

PR #884, opened 2025-07-01
build: hoist the BlocksRuntime `add_subdirectory`

compnerd:
Move the subdirectory recursion for the BlocksRuntime to the top level as this is a separate target entity.


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #886, opened 2025-07-11
build: correct linking to static libdispatch on Windows

compnerd:
When building against the static libdispatch, we need to inform the dispatch headers that the library will be linked statically via `-Ddispatch_STATIC`. Wire this up onto the dispatch library to allow `swiftDispatch` to properly propagate that flag.


compnerd:
@swift-ci please test

compnerd:

Issue #887, opened 2025-07-15
Race condition in muxnote lifecycle on Windows


hmelder:
In event_windows.c, reference counting seems to be used to model a happens-before relationship:

Thread A is currently in `_dispatch_event_loop_drain` and waits on a completion event. Completion events are emitted by handlers such as the `_dispatch_pipe_monitor_thread`. Let thread B be such a thread. Before thread B posts the completion event, the reference count of the muxnote is incremented with `_dispatch_muxnote_retain`. This is an atomic increment with relaxed memory order.

Thread A receives the completion event and may call `_dispatch_event_merge_pipe_handle_read` which releases the muxnote. A comment above this release states that the muxnote is "[r]etained when posting the completion packet", but this  happens in Thread B.

Therefore, when `_dispatch_muxnote_release` is called, the retain operation may not be visible to the thread performing the release, resulting in a disposal of the muxnote despite still being used.

If I understand this comment and the logic correctly, there needs to be some form of synchronisation between thread A and thread B and the reference counting does not provide this.

See:
#844 and #833

CC: @compnerd

compnerd:
The relaxed is sufficient for atomic refcount handling, not the thread synchronisation. I think that the `_dispatch_muxnote_release` needs the `acq_rel` ordering to decrement the refcount. This would ensure that the acquire semantics in the `_dispatch_pipe_monitor_thread` are not violated.

Issue #844, opened 2024-09-16
Occasional `swiftc` crash on Windows, "disposed a muxnote with an active thread"


z2oh:
Upon upgrading our Azure CI machines to use the new Azure Cobalt ARM64 processors, we started seeing frequent compiler crashes when building a large Swift project. After some investigation, the culprit appears to be a lifecycle violation in libdispatch in the Windows pipe handling code.

The crashing line: https://github.com/apple/swift-corelibs-libdispatch/blob/e85f6a0d5c9ea1f32f5013c3fa34e4fc146cd0eb/src/event/event_windows.c#L240

And the stack trace:
```
[Inline Frame] dispatch.dll!_dispatch_muxnote_dispose(dispatch_muxnote_s * dmn) Line 240s://github.com/apple/swift-corelibs-libdispatch/blob/e85f6a0d5c9ea1f32f5013c3fa34e4fc146cd0eb/src/event/event_windows.c#L240)	C
[Inline Frame] dispatch.dll!_dispatch_muxnote_release(dispatch_muxnote_s * dmn) Line 265	C
[Inline Frame] dispatch.dll!_dispatch_event_merge_pipe_handle_read(dispatch_muxnote_s * dmn, unsigned long dwBytesAvailable) Line 669	C
dispatch.dll!_dispatch_event_loop_drain(unsigned int flags) Line 915	C
dispatch.dll!_dispatch_mgr_invoke() Line 5419	C
dispatch.dll!_dispatch_mgr_thread(dispatch_lane_s * dq, dispatch_invoke_context_s * dic, <unnamed-tag> flags) Line 5447	C
[Inline Frame] dispatch.dll!_dispatch_continuation_pop_inline(dispatch_object_t dou, dispatch_invoke_context_s * dic, <unnamed-tag> flags, dispatch_queue_class_t dqu) Line 2496	C
dispatch.dll!_dispatch_root_queue_drain(dispatch_queue_global_s * dq, unsigned int pri, <unnamed-tag> flags) Line 6114	C
dispatch.dll!_dispatch_worker_thread(void * context) Line 6250	C
dispatch.dll!_dispatch_worker_thread_thunk(void * lpParameter) Line 6272	C
[External Code]	
```

I suspect this is not an Cobalt/ARM64 specific issue, but is more likely a long-standing bug which has become common on this particular line of CPUs due to some scheduling or timing change.

The interesting section is here:
[...]

PR #888, opened 2025-07-15

compnerd:

Fixes: #887, #844

### DIFF
--- a/src/event/event_windows.c
+++ b/src/event/event_windows.c
@@ -260,8 +260,16 @@ _dispatch_muxnote_retain(dispatch_muxnote_t dmn)
 static void
 _dispatch_muxnote_release(dispatch_muxnote_t dmn)
 {
-	uintptr_t refcount = os_atomic_dec(&dmn->dmn_refcount, relaxed);
+	// We perform a minor optimization here - perform the decrement with
+	// release semantics. In the case that we are going to dispose of the
+	// value, we perform the acquire fence. This reduces the cost on the
+	// normal path by avoiding the acquire fence. This should be more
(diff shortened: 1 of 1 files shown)


hmelder:
There is an example in the boost docs where only release and an additional memory fence is used. Here is the rational behind it:

> It would be possible to use memory_order_acq_rel for the fetch_sub operation, but this results in unneeded "acquire" operations when the reference counter does not yet reach zero and may impose a performance penalty.

Would it make sense to adopt a similar approach and use `os_atomic_thread_fence`?

```c++
#include <boost/intrusive_ptr.hpp>
#include <boost/atomic.hpp>

class X {
public:
  typedef boost::intrusive_ptr<X> pointer;
  X() : refcount_(0) {}

private:
  mutable boost::atomic<int> refcount_;
  friend void intrusive_ptr_add_ref(const X * x)
  {
    x->refcount_.fetch_add(1, boost::memory_order_relaxed);
  }
  friend void intrusive_ptr_release(const X * x)
  {
    if (x->refcount_.fetch_sub(1, boost::memory_order_release) == 1) {
      boost::atomic_thread_fence(boost::memory_order_acquire);
      delete x;
    }
  }
};
```
https://www.boost.org/doc/libs/1_57_0/doc/html/atomic/usage_examples.html#boost_atomic.usage_examples.example_reference_counters

compnerd:
Yeah, that would be a bit cheaper on x64 at least, and without too much more complexity; let me see if I can integrate that approach.

compnerd:
@swift-ci please test

compnerd:
Going to merge this for now - this bug has haunted us for a while, and it would be good to get this tested more broadly.

PR #898, opened 2025-08-14

compnerd:

Fixes: #887, #844


compnerd:
@swift-ci please test

compnerd:
- **Explanation**: Fixes a subtle race condition in dispatch which is more common ARM64
- **Scope**: The change is limited to Windows
- **Issues**: #844, #887
- **Original PRs**: #888 
- **Risk**: This is medium risk on Windows as it _could_ uncover additional race conditions. There is no risk to any other platform.
- **Testing**: 24 hour stress test on ARM64 cobalt CPU
- **Reviewers**: @compnerd

PR #900, opened 2025-08-14

compnerd:

Fixes: #887, #844


compnerd:
@swift-ci please test

PR #906, opened 2025-09-02
Test: Check result of `sysctlbyname`

etcwilde:
Adding a check to verify that `sysctlbyname` didn't fail in `dispatch_workqeueu`. If it does, `activecpu` will contain random garbage. The test will fail, but it isn't clear why.


etcwilde:
@swift-ci please test

PR #908, opened 2025-09-16
Add thread names on linux

da-viper:
Truncate the name of threads from the front instead to avoid allocating a new string and the end of the name tends to be more unique when creating multiple threads.


rokhinip:
@swift-ci test

ktoso:
@swift-ci test

PR #917, opened 2025-10-30
Enable Windows ARM64/ARM64EC

anthony-linaro:
This PR allows building for ARM64 and ARM64EC.

ARM64 worked out of the box, but the `dispatch_select` test failed, due to a use-after-free. For some reason this didn't affect other platforms/architectures, but writing to the `buffer` variable caused a memory fault on native Windows ARM64.

ARM64EC has a few define checks added, to ensure x64 ASM calls aren't run.

Tested using VS2022 and LLVM 21.1.2.


compnerd (approved):

compnerd:
@swift-ci please test

PR #931, opened 2026-02-17
Handle kevent timer delays on OpenBSD properly.

3405691582:
Dispatch likes durations specified in ns. Earlier, for whatever reason, we didn't use the flags for EVFILT_TIMER to expect ns to and instead scaled manually to match the default, which is ms. This scaling was conditioned on NOTE_ABSOLUTE not being defined. In #879, support was added to ensure uses of NOTE_ABSOLUTE were matched to the BSD-like spelling, NOTE_ABSTIME. But this then meant that we were no longer scaling the durations on OpenBSD.

This was all a terrible hack to begin with. Remove the manual scaling condition and use NOTE_NSECONDS like we should have done back in the beginning. (What was I thinking?)


grynspan (approved):

src/event/event_kevent.c
jakepetroules: Is this _ever_ valid for these to be zero or should it be something like "#error Define this for your platform..."

Seems like it has a good chance of tripping someone up if Swift is ported to NetBSD in future, for example.

PR #936, opened 2026-03-17
[6.3] Fix worker thread name handling on Linux

KushalP:
**Explanation:** The existing functionality is brittle on Linux, and the UX can be improved.
**Scope:** Narrow. Hard codes the thread name for queue workers.
**Original PRs:**
- https://github.com/swiftlang/swift-corelibs-libdispatch/pull/934
- https://github.com/swiftlang/swift-corelibs-libdispatch/pull/937
- https://github.com/swiftlang/swift-corelibs-libdispatch/pull/938

**Risk:** Very low.
**Testing:** CI

PR #939, opened 2026-04-16
Guard pthread_getname_np usage in dispatch_thread_name test

triplef:
With #937, building for Android < r26 fails with the following:

```
  -- Looking for pthread_setname_np - found
  -- Looking for pthread_getname_np - not found

...

ld: error: undefined symbol: pthread_getname_np
>>> referenced by dispatch_thread_name.c:74
>>>               CMakeFiles/dispatch_thread_name.dir/dispatch_thread_name.c.o:(__main_block_invoke)
```

This is because `pthread_getname_np` is only available starting at API level 26 and the `dispatch_thread_name` test calls `pthread_getname_np` unguarded by the define.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -163,7 +163,7 @@ if(EXTENDED_TEST_SUITE)
 endif()
 
 # add C tests for platform-specific functionality when applicable
-if(UNIX AND (HAVE_PTHREAD_SETNAME_NP OR HAVE_PTHREAD_SET_NAME_NP))
+if(UNIX AND (HAVE_PTHREAD_SETNAME_NP OR HAVE_PTHREAD_SET_NAME_NP) AND (HAVE_PTHREAD_GETNAME_NP OR HAVE_PTHREAD_GET_NAME_NP))
   list(APPEND DISPATCH_C_TESTS
        thread_name)
 endif()


triplef:
@3405691582 and @etcwilde I’d appreciate your reviews.

3405691582 (approved):
Seems reasonable. It might be worth declaring some intermediary variables to make the conditional clause simpler, but I'm not a cmake expert.

KushalP:
As requested by @akmorrison, I've cherry-picked this commit onto #936.

akmorrison:
@swift-ci please test

akmorrison:
Surprising to me that these went in at different times: https://cs.android.com/android/platform/superproject/main/+/main:bionic/libc/include/pthread.h

```
#if defined(__USE_GNU) && __BIONIC_AVAILABILITY_GUARD(26)
int pthread_getname_np(pthread_t __pthread, char* _Nonnull __buf, size_t __n) __INTRODUCED_IN(26);
#endif

int pthread_setname_np(pthread_t __pthread, const char* _Nonnull __name);
```

Change looks good, approving and will merge once the tests come back

akmorrison (approved):

etcwilde (approved):

PR #943, opened 2026-05-22
Use relative kqueue deadlines for BSD uptime and monotonic timers

max-potapov:
## Summary

On FreeBSD and OpenBSD, `NOTE_NSECONDS | NOTE_ABSTIME` uses a realtime absolute deadline. libdispatch supplied boot-relative values for uptime and monotonic timers, so kqueue treated them as expired and repeatedly woke and re-armed the timer.

This change:
- keeps wall-clock timers on the absolute path;
- programs uptime and monotonic timers as relative delays on FreeBSD and OpenBSD;
- adds a public-API regression test that detects CPU use while a far-future timer is idle.

## Validation

Tested on FreeBSD 15.1 with Swift 6.3-dev:

- current `main`: the new idle timer test reports a relative CPU ratio of approximately `0.74` and fails;
- patched: the test reports approximately `0.0006` and passes;
- full default C test suite: `25/25` passed;
- the original Hummingbird workload no longer produces the kqueue re-arm loop.

Runtime validation was performed on FreeBSD. The OpenBSD change follows the same existing `NOTE_ABSTIME` clock semantics.

## Shared clock helper experiment

I also tested adding `kern.boottime` to `_dispatch_uptime` and `_dispatch_monotonic_time` instead of changing the kqueue backend. It stopped the timer loop, but introduced two regressions in local testing:
- `dispatch_time(DISPATCH_TIME_NOW, ...)` increased from approximately `43 ns` to `565 ns` per call because the shared path performs a `sysctl`;
- after a test shim advanced only the returned `kern.boottime` value by one hour, a pre-existing 200 ms monotonic deadline expired in `0.032 ms`. The backend-local version expired in `200.142 ms`.

FreeBSD adjusts `kern.boottime` when system time is stepped and after resume, so this revision leaves the shared uptime and monotonic clock helpers unchanged.

## Update history

- 2026-05-22: added the initial backend-local fix and FreeBSD workload reproduction.
- 2026-07-15: rebased onto current `main`, added the idle CPU regression test, reduced the backend comments, and evaluated the `kern.boottime` alternative.

### DIFF
--- a/src/event/event_kevent.c
+++ b/src/event/event_kevent.c
@@ -68,10 +68,22 @@ DISPATCH_STATIC_GLOBAL(struct dispatch_muxnote_bucket_s _dispatch_sources[DSL_HA
 #define DISPATCH_NOTE_CLOCK_UPTIME    0
 #endif
 
+#if defined(__FreeBSD__) || defined(__OpenBSD__)
+// BSD NOTE_ABSTIME deadlines use CLOCK_REALTIME.
+#define DISPATCH_NOTE_ABSOLUTE_WALL      NOTE_ABSOLUTE
+#define DISPATCH_NOTE_ABSOLUTE_MONOTONIC 0
+#define DISPATCH_NOTE_ABSOLUTE_UPTIME    0
+#else
+#define DISPATCH_NOTE_ABSOLUTE_WALL      NOTE_ABSOLUTE
+#define DISPATCH_NOTE_ABSOLUTE_MONOTONIC NOTE_ABSOLUTE
+#define DISPATCH_NOTE_ABSOLUTE_UPTIME    NOTE_ABSOLUTE
+#endif
+
 static const uint32_t _dispatch_timer_index_to_fflags[] = {
 #define DISPATCH_TIMER_FFLAGS_INIT(kind, qos, note) \
 	[DISPATCH_TIMER_INDEX(DISPATCH_CLOCK_##kind, DISPATCH_TIMER_QOS_##qos)] = \
-			DISPATCH_NOTE_CLOCK_##kind | NOTE_ABSOLUTE | NOTE_LEEWAY | (note)
+			DISPATCH_NOTE_CLOCK_##kind | DISPATCH_NOTE_ABSOLUTE_##kind | \
+			NOTE_LEEWAY | (note)
 	DISPATCH_TIMER_FFLAGS_INIT(WALL, NORMAL, 0),
 	DISPATCH_TIMER_FFLAGS_INIT(UPTIME, NORMAL, 0),
 	DISPATCH_TIMER_FFLAGS_INIT(MONOTONIC, NORMAL, 0),
@@ -2414,6 +2426,10 @@ _dispatch_event_loop_timer_arm(dispatch_timer_heap_t dth, uint32_t tidx,
 	}
 #if !NOTE_ABSOLUTE
 	target = range.delay;
+#elif defined(__FreeBSD__) || defined(__OpenBSD__)
+	if (clock != DISPATCH_CLOCK_WALL) {
+		target -= _dispatch_time_now_cached(clock, nows);
+	}
 #endif
 
 	_dispatch_event_loop_timer_program(dth, tidx, target, range.leeway,
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -140,6 +140,10 @@ set(DISPATCH_C_TESTS
     io_pipe_close
     select)
 
+if(UNIX)
+  list(APPEND DISPATCH_C_TESTS timer_idle)
+endif()
+
 # Tests that usually pass, but occasionally fail.
 # Excluded by default for purposes of Swift CI
 if(EXTENDED_TEST_SUITE)
--- a/tests/dispatch_timer_idle.c
+++ b/tests/dispatch_timer_idle.c
@@ -1,0 +1,150 @@
+/*
+ * This source file is part of the Swift.org open source project
+ *
+ * Copyright (c) 2026 Apple Inc. and the Swift project authors
+ *
+ * Licensed under Apache License v2.0 with Runtime Library Exception
+ *
+ * See https://swift.org/LICENSE.txt for license information
+ * See https://swift.org/CONTRIBUTORS.txt for the list of Swift project authors
+ *
+ */
+
+#include <dispatch/dispatch.h>
+
+#include <bsdtests.h>
+#include "dispatch_test.h"
+
+#include <errno.h>
+#include <stdatomic.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/resource.h>
+#include <time.h>
+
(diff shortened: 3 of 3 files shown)


src/event/event_kevent.c
```diff
@@ -50,20 +50,41 @@ DISPATCH_STATIC_GLOBAL(struct dispatch_muxnote_bucket_s _dispatch_sources[DSL_HA
...
+ * libdispatch computes for DISPATCH_CLOCK_WALL, so wall timers stay on
+ * the absolute-time path and continue to track wall-clock adjustments.
+ *
+ * For UPTIME / MONOTONIC libdispatch hands kqueue a CLOCK_MONOTONIC
+ * nanosecond value which BSD kqueue would interpret as realtime
+ * (decades in the past after boot), reporting the timer as
+ * already-expired on every register and producing a kevent() tight
+ * loop.  Drop NOTE_ABSOLUTE for these clocks and let
+ * _dispatch_event_loop_timer_arm fall through to the relative-time
+ * path below. */
```
kevans91: IMO this comment reads a little more complicated than it probably needs to, and maybe the one below as well.  The root of the problem seems to be that `_dispatch_time_now_cached` is assumed below to always returrn an absolute time, but our UPTIME/MONOTONIC (which are synonyms, at least on FreeBSD) are relative to boot @ time=0.

Writing that out loud, I'm wondering if this breaks other assumptions of users that consume these clocks through `_dispatch_time_now*` -- it might be better to have `_dispatch_uptime` and `_dispatch_monotonic_time` fetch kern.boottime and add that in to avoid bugs like this in the future.

kevans91: (That is: I see the note in the PR description, but I'm not sure that's actually a much larger change and it avoids other inconsistencies at the same time)

max-potapov: Thanks @kevans91 – agreed, I can trim the comments.

I also agree that normalizing the BSD uptime/monotonic path around boottime is probably the cleaner architectural fix if those values are meant to feed the absolute-deadline timer path.

My hesitation is mostly validation, not preference for the narrow patch. The existing dispatch_timer tests pass on FreeBSD with both upstream and this patch, and the simple dispatch_after wall/uptime smoke tests also pass. So they don't currently exercise the long-running re-arm/tight-loop behavior that exposed this in the integration workload.

Before I rework this toward the boottime-based fix, what validation would you consider sufficient here? I can either:
- keep this PR as the narrow fix and a record of the bug,
- rework it into the boottime-based approach and validate with the current suite plus the original repro,
- or first try to add a BSD-specific regression test that actually fails upstream.

I'm fine with whichever path is more useful for the project.

kevans91: > Thanks @kevans91 – agreed, I can trim the comments.
> 
> I also agree that normalizing the BSD uptime/monotonic path around boottime is probably the cleaner architectural fix if those values are meant to feed the absolute-deadline timer path.
> 
> My hesitation is mostly validation, not preference for the narrow patch. The existing dispatch_timer tests pass on FreeBSD with both upstream and this patch, and the simple dispatch_after wall/uptime smoke tests also pass. So they don't currently exercise the long-running re-arm/tight-loop behavior that exposed this in the integration workload.
> 
> Before I rework this toward the boottime-based fix, what validation would you consider sufficient here? I can either:
> 
> * keep this PR as the narrow fix and a record of the bug,
> * rework it into the boottime-based approach and validate with the current suite plus the original repro,
> * or first try to add a BSD-specific regression test that actually fails upstream.
> 

IMO option 3 is the better option here- presumably you wouldn't really need to call it a BSD-specific test: it may not provide the exact same value on Linux, but it exercises a path through libdispatch that is seemingly not well-exercised today (whether that be kqueue or epoll/etc).

> I'm fine with whichever path is more useful for the project.

With a disclaimer that I'm only tangentially involved in libdispatch things, though heavily involved in FreeBSD itself- I'd probably want someone else to stamp the path before you expand too much energy on it since their standards and values will certainly differ.

max-potapov: @kevans91 I added the regression test first, as suggested. It creates a far-future dispatch source timer and compares idle process CPU use against a calibrated busy loop.
On current `main` on FreeBSD 15.1 it reports a relative CPU ratio of approximately `0.74` and fails. With the backend-local fix it reports approximately `0.0006` and passes.

I also tested adding `kern.boottime` to `_dispatch_uptime` and `_dispatch_monotonic_time`. It stops the timer loop, but local testing found two regressions:
- `dispatch_time(DISPATCH_TIME_NOW, ...)` increased from approximately `43 ns` to `565 ns` per call due to the `sysctl` on the shared path.
- A test shim advanced only the returned `kern.boottime` value by one hour after creating a 200 ms monotonic deadline. The deadline then expired in `0.032 ms`; the backend-local version expired in `200.142 ms`.

FreeBSD adjusts `kern.boottime` when system time is stepped and after resume.

For this revision I kept the shared clock helpers unchanged and retained the backend-local relative conversion. The branch is rebased onto current `main`; the full default C test suite passes (`25/25`).

kevans91: > @kevans91 I added the regression test first, as suggested. It creates a far-future dispatch source timer and compares idle process CPU use against a calibrated busy loop. On current `main` on FreeBSD 15.1 it reports a relative CPU ratio of approximately `0.74` and fails. With the backend-local fix it reports approximately `0.0006` and passes.
> 

Awesome!

> I also tested adding `kern.boottime` to `_dispatch_uptime` and `_dispatch_monotonic_time`. It stops the timer loop, but local testing found two regressions:
> 
> * `dispatch_time(DISPATCH_TIME_NOW, ...)` increased from approximately `43 ns` to `565 ns` per call due to the `sysctl` on the shared path.
> * A test shim advanced only the returned `kern.boottime` value by one hour after creating a 200 ms monotonic deadline. The deadline then expired in `0.032 ms`; the backend-local version expired in `200.142 ms`.
> 
> FreeBSD adjusts `kern.boottime` when system time is stepped and after resume.
> 

*sigh* yes, sorry: https://github.com/freebsd/freebsd-src/blob/main/sys/kern/kern_tc.c#L1446

brad0:
This looks like it is related to some unusual behaviour noticed on OpenBSD when we updated from 6.0 to 6.4.

max-potapov:
@swift-ci please test

max-potapov:
Rebased onto current main and requested Swift CI. The regression test now demonstrates the FreeBSD timer wakeup/CPU loop and the backend-local relative-deadline fix resolves it without changing shared clock semantics. Could the libdispatch maintainers give a go/no-go on this approach, especially given the related OpenBSD behavior? @compnerd @akmorrison @kevans91 @brad0

brad0:
@max-potapov This has been committed to our ports tree. Thanks.

akmorrison (approved):
Looks good to me, if the tests are passing for you and it's working in the upstream ports tree, I have no issue with this. I don't have a local box to test the BSD changes on, so relying on the data from @max-potapov on that.

akmorrison:
@swift-ci please test

Issue #779, opened 2023-02-09
Add `Sendable` conformances to thread-safe types


fumoboy007:
All or almost all of `Dispatch`’s types are thread-safe, so they should be marked as [`Sendable`](https://developer.apple.com/documentation/swift/sendable).

mman:
There is this weird state in Swift 5.10 where on Linux usage of Dispatch with StrictConcurrencyChecking enabled generates these nonsensical errors that do not exist on macOS.

For example:

```
main.swift:131:31: warning: passing argument of non-sendable type 'DispatchQueue' outside of main actor-isolated context may introduce data races
        try await host.start(queue: queue)
                              ^
Dispatch.DispatchQueue:1:14: note: class 'DispatchQueue' does not conform to the 'Sendable' protocol
public class DispatchQueue : DispatchObject {
             ^
main.swift:2:1: remark: add '@preconcurrency' to suppress 'Sendable'-related warnings from module 'Dispatch'
import Dispatch
^
```

This needs fixing...

PR #946, opened 2026-07-01
Add missing Swift wrapper conformances

dblsaiko:
DispatchSourceProtocol conforms to NSObjectProtocol on macOS, using AnyObject here. The other DispatchSource protocols all conform to Sendable.

Partial fix for #779.

### DIFF
--- a/src/swift/Wrapper.swift
+++ b/src/swift/Wrapper.swift
@@ -213,7 +213,7 @@ internal class __DispatchData : DispatchObject {
 
 public typealias DispatchSourceHandler = @convention(block) () -> Void
 
-public protocol DispatchSourceProtocol {
+public protocol DispatchSourceProtocol : AnyObject {
 	func setEventHandler(qos: DispatchQoS, flags: DispatchWorkItemFlags, handler: DispatchSourceHandler?)
 
 	func setEventHandler(handler: DispatchWorkItem)
@@ -241,20 +241,20 @@ public protocol DispatchSourceProtocol {
 	var isCancelled: Bool { get }
 }
 
-public protocol DispatchSourceUserDataAdd : DispatchSourceProtocol {
+public protocol DispatchSourceUserDataAdd : DispatchSourceProtocol, Sendable {
 	func add(data: UInt)
 }
 
-public protocol DispatchSourceUserDataOr : DispatchSourceProtocol {
+public protocol DispatchSourceUserDataOr : DispatchSourceProtocol, Sendable {
 	func or(data: UInt)
 }
 
-public protocol DispatchSourceUserDataReplace : DispatchSourceProtocol {
+public protocol DispatchSourceUserDataReplace : DispatchSourceProtocol, Sendable {
 	func replace(data: UInt)
 }
 
 #if HAVE_MACH
-public protocol DispatchSourceMachSend : DispatchSourceProtocol {
+public protocol DispatchSourceMachSend : DispatchSourceProtocol, Sendable {
 	public var handle: mach_port_t { get }
 
 	public var data: DispatchSource.MachSendEvent { get }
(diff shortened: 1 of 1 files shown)


dblsaiko:
Hm, this makes Swift not build correctly due to early swift-driver being built against the origin libdispatch without AnyObject but then run with LD_LIBRARY_PATH when building swiftpm including the new library with the ABI with AnyObject which then makes it crash at runtime when calling DispatchSource.makeSignalSource. Will have to investigate more closely.

dblsaiko:
The patch is fine, Swift's build tools just separates the two toolchains uncleanly. I got it to build correctly with this additional patch to swiftlang/swift so that it won't mix libraries while bootstrapping.

<details>

```patch
Commit ID: f6958042f4e619a0aff993d6c69abb6b41454ced
Change ID: yyswruqowsoylrykqlnuslmomqzuuquv
Author   : Katalin Rebhan <me@dblsaiko.net> (02 Jul 2026 02:23:57)
Committer: Katalin Rebhan <me@dblsaiko.net> (02 Jul 2026 13:55:28)

    Unset (DY)LD_LIBRARY_PATH when calling bootstrap Swift compiler

    While bootstrapping SwiftPM, in the second stage where we build SwiftPM
    with itself, the (DY)LD_LIBRARY_PATH variables are set in order for
    SwiftPM to run correctly. However, the swiftc here comes from the
    bootstrap toolchain (i.e. earlyswiftdriver), which is still partially
    linked with libraries from the previous Swift compiler's toolchain, such
    as libdispatch. Inheriting (DY)LD_LIBRARY_PATH here causes libraries
    built from the new toolchain to be loaded for the binary built against
    the old toolchain, even if they have incompatible ABI. This can cause
    issues where they don't match, for example while writing [1].

    [1] https://github.com/swiftlang/swift-corelibs-libdispatch/pull/946

diff --git a/utils/swift_build_support/swift_build_support/products/swiftpm.py b/utils/swift_build_support/swift_build_support/products/swiftpm.py
index 5a3e938637..de715ea4a7 100644
--- a/utils/swift_build_support/swift_build_support/products/swiftpm.py
+++ b/utils/swift_build_support/swift_build_support/products/swiftpm.py
@@ -11,6 +11,7 @@
 # ----------------------------------------------------------------------------
 
 import os
+import textwrap
 
 from . import cmark
 from . import foundation
@@ -58,6 +59,32 @@
         swiftc = os.path.join(toolchain_path, "bin", "swiftc")
         clang = os.path.join(clang_tools_path, "bin", "clang")
 
+        # The SwiftPM build script later sets LD_LIBRARY_PATH to some of its
+        # built libraries when building SwiftPM with itself. However, the
+        # native toolchain must not be run with these libraries since they
+        # might have different ABI. Make sure LD_LIBRARY_PATH is not
+        # propagated to it.
+        os.makedirs(self.build_dir, exist_ok=True)
+        wrapped_swiftc = os.path.join(self.build_dir, 'swiftc.py')
+        with open(wrapped_swiftc, 'w') as f:
+            print(
+                textwrap.dedent(f'''
+                #!/usr/bin/env python3
+                import subprocess
+                import sys
+                import os
+
+                for key in ['LD_LIBRARY_PATH', 'DYLD_LIBRARY_PATH']:
+                    try:
+                        del os.environ[key]
+                    except KeyError:
+                        pass
+
+                completed_process = subprocess.run([{swiftc!r}, *sys.argv[1:]])
+                sys.exit(completed_process.returncode)
+                ''').strip(), file=f)
+            os.fchmod(f.fileno(), 0o755)
+
         # FIXME: We require llbuild build directory in order to build. Is
         # there a better way to get this?
         build_root = os.path.dirname(self.build_dir)
@@ -75,7 +102,7 @@
             helper_cmd.append("--release")
 
         helper_cmd += [
-            "--swiftc-path", swiftc,
+            "--swiftc-path", wrapped_swiftc,
             "--clang-path", clang,
             "--cmake-path", self.toolchain.cmake,
             "--ninja-path", self.toolchain.ninja,
```

</details>

PR #948, opened 2026-07-07
Add DragonFlyBSD compile-time check for missing NOTE_NSECONDS

### LINKED ISSUE #933: Handle the scenario when NOTE_NSECONDS is not available for the kevent backend.
DragonFlyBSD seems to be the only kevent implementer that doesn't support flags like NOTE_NSECONDS, and would require manual adjustment, since by default EVFILT_TIMER defaults to ms.

We should add an `#error` or `#warning` when `NOTE_NSECONDS` is not defined to make the adjustment in the correct place, or just do it preemptively.

somiljain2006:
Fixes #933 

Add a compile-time check for DragonFlyBSD timer support by detecting the absence of NOTE_NSECONDS. This prevents silently using nanosecond timer values with EVFILT_TIMER implementations that default to milliseconds and makes the required platform-specific adjustment explicit while remaining compatible with future DragonFlyBSD versions that may add NOTE_NSECONDS support.


somiljain2006:
@akmorrison @compnerd Can you review this pr?

compnerd (approved):

compnerd:
@swift-ci please test

akmorrison (approved):

PR #951, opened 2026-07-29
Fix importing dispatch's static build with cmake

keith:
Without this you end up with errors like:

```
The link interface of target "dispatch" contains:

  RT::rt

but the target was not found.
```

### DIFF
--- a/cmake/modules/dispatchConfig.cmake.in
+++ b/cmake/modules/dispatchConfig.cmake.in
@@ -1,24 +1,15 @@
+include(CMakeFindDependencyMacro)
+
 set(DISPATCH_HAS_SWIFT_SDK_OVERLAY @ENABLE_SWIFT@)
 
 if(NOT @BUILD_SHARED_LIBS@)
-  if(NOT TARGET Threads::Threads)
-    add_library(Threads::Threads INTERFACE IMPORTED)
-    if("@THREADS_HAVE_PTHREAD_ARG@")
-      set_property(TARGET Threads::Threads PROPERTY INTERFACE_COMPILE_OPTIONS
-        "$<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:SHELL:-Xcompiler -pthread>"
-        "$<$<AND:$<NOT:$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>>,$<NOT:$<COMPILE_LANGUAGE:Swift>>>:-pthread>")
-    endif()
-    if(NOT "@CMAKE_THREAD_LIBS_INIT@" STREQUAL "")
-      set_property(TARGET Threads::Threads PROPERTY INTERFACE_LINK_LIBRARIES
-        "@CMAKE_THREAD_LIBS_INIT@")
-    endif()
-  endif()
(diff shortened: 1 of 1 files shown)


keith:
@swift-ci please test

keith:
used in https://github.com/swiftlang/swift/pull/91049

compnerd (changes requested):
This is the standard "exported target has external dependencies" case: because `Threads::Threads` and `RT::rt` are `PRIVATE` on `dispatch`, the static archive's `INTERFACE_LINK_LIBRARIES` carries them as `$<LINK_ONLY:...>`, so the imported target references names the consumer has never heard of. Resolving those is the package config's responsibility — the client should only ever have to write `find_package(dispatch)`.

That's precisely what `find_dependency` is for, so let's delegate to `FindThreads` rather than reimplement it:

```suggestion
include(CMakeFindDependencyMacro)

set(DISPATCH_HAS_SWIFT_SDK_OVERLAY @ENABLE_SWIFT@)

if(NOT @BUILD_SHARED_LIBS@)
  set(CMAKE_THREAD_PREFER_PTHREAD @CMAKE_THREAD_PREFER_PTHREAD@)
  set(THREADS_PREFER_PTHREAD_FLAG @THREADS_PREFER_PTHREAD_FLAG@)
  find_dependency(Threads)

  if(@LibRT_FOUND@)
    list(APPEND CMAKE_MODULE_PATH "@CMAKE_CURRENT_SOURCE_DIR@")
    find_dependency(LibRT)
  endif()
endif()
```

Forwarding the two preference variables is what keeps the client's `Threads::Threads` identical to the one we linked against, so there's no need to expand `@THREADS_HAVE_PTHREAD_ARG@` or `@CMAKE_THREAD_LIBS_INIT@` by hand. `find_dependency` over a bare `find_package` because it propagates `QUIET`/`REQUIRED` from the client's `find_package(dispatch)` and fails as `dispatch_NOT_FOUND_MESSAGE` instead of erroring out mid-configure.

Two notes on what the suggestion drops:

- The `$<NOT:$<COMPILE_LANGUAGE:Swift>>` guard on `-pthread` isn't needed. `$<LINK_ONLY:>` suppresses everything but the link, so `INTERFACE_COMPILE_OPTIONS` never reaches the consumer's compiles and `-pthread` never reaches `swiftc`.
- `RT::rt` comes from our own `FindLibRT.cmake` rather than a system module, hence the `CMAKE_MODULE_PATH` line. It resolves for the build tree since this file is configured out of `cmake/modules`; if we add an installed config later, that module has to ship next to it and the path becomes `${CMAKE_CURRENT_LIST_DIR}`.

compnerd (approved):
I think that this looks right; including Evan incase he has thoughts.

CC: @etcwilde

compnerd:
@swift-ci please test

etcwilde (approved):
This looks fine. Might need to make some changes if we ever want to install this file since `CMAKE_CURRENT_SOURCE_DIR` won't exist. Something like `CMAKE_CURRENT_LIST_DIR` might be more appropriate? Otherwise, LGTM

compnerd:
@swift-ci please test

PR #953, opened 2026-08-22
feat: prepare shared code for cooperative single-threaded event backends

scottmarchant:
## Summary

This PR moves one function, `_dispatch_main_queue_drain`, out of `DISPATCH_COCOA_COMPAT` and under a separate guard. It makes no functional change on any current platform.

## Context

WebAssembly (`wasm32-wasip1`) runs libdispatch on a single thread. The port registers with the host event loop as an event source, like libdispatch registers with CFRunLoop on Darwin. A poke asks the host for one later callback. The host then invokes libdispatch, and `_dispatch_main_queue_drain` drains the main queue.

An earlier revision of this PR added hook macros around eight critical sections. Those hooks supported a drain-on-poke model that ran work on the caller stack. The port dropped that model, and this PR dropped the hooks. The complete port lives on our fork as one reviewable branch: [PassiveLogic/swift-corelibs-libdispatch#3](https://github.com/PassiveLogic/swift-corelibs-libdispatch/pull/3).

## What changes

`_dispatch_main_queue_drain` moves under a guard of its own. The two runloop-only steps stay under `DISPATCH_COCOA_COMPAT`: the runloop-handle initialization, and the thread-QoS override propagation. The guard also names `defined(__wasi__)`. The new `__wasi__` arm adds no compiled code in this PR.

## Why this is safe

- On Darwin, `DISPATCH_COCOA_COMPAT` retains both runloop-only steps. Darwin builds the same code as before.
- On Linux, neither version compiles this code, as before.

## What we ask from reviewers

The public macOS CMake build of this repository does not configure, so we cannot compile the `DISPATCH_COCOA_COMPAT` path ourselves. We ask a committer to trigger `@swift-ci please test`. We also ask which CI configuration compiles the `DISPATCH_COCOA_COMPAT` path.

### DIFF
--- a/src/queue.c
+++ b/src/queue.c
@@ -6877,6 +6877,13 @@ _dispatch_main_queue_update_priority_from_thread(void)
 	}
 }
 
+#endif // DISPATCH_COCOA_COMPAT
+#if DISPATCH_COCOA_COMPAT || defined(__wasi__)
+// Shared between the CFRunLoop callback path (DISPATCH_COCOA_COMPAT) and a
+// cooperative single-threaded drain, which owns the thread-bound main
+// queue's drain lock for the lifetime of the program. The runloop-handle
+// initialization and the thread-QoS override propagation are runloop/Darwin
+// machinery and compile only for COCOA_COMPAT.
 static void
 _dispatch_main_queue_drain(dispatch_queue_main_t dq)
 {
@@ -6898,8 +6905,10 @@ _dispatch_main_queue_drain(dispatch_queue_main_t dq)
 				" from the wrong thread");
 	}
 
+#if DISPATCH_COCOA_COMPAT
 	dispatch_once_f(&_dispatch_main_q_handle_pred, dq,
 			_dispatch_runloop_queue_handle_init);
+#endif
 
 	// <rdar://problem/23256682> hide the frame chaining when CFRunLoop
 	// drains the main runloop, as this should not be observable that way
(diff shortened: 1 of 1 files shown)


scottmarchant:
@akmorrison Thanks, great feedback! You are right, and we changed the port to match.

The port now registers with the host event loop as an event source, like libdispatch does with CFRunLoop. A poke records pending work and asks the host for one later callback. The host later invokes libdispatch. `_dispatch_main_queue_drain` then drains the main queue, the same function the `_4CF` callback uses. Drain on poke is gone.

We removed the hook macros with the drain-on-poke model. This PR now changes 14 added lines and one removed line in `src/queue.c`. We updated the PR description.

Consumer code does not change. The host runtime registers the callback, the way Foundation wires the eventfd on Linux. Command modules use `dispatch_main()` and need no wiring.

Just for awareness: wasip1 has no eventfd, pipe, or mach port, so it has no usable `dispatch_runloop_handle_t`. The WASI backend instead registers one host callback and exposes a next-deadline query for timers. That SPI arrives with the event backend slice. The design and its tests are in [PassiveLogic/swift-corelibs-libdispatch#3](https://github.com/PassiveLogic/swift-corelibs-libdispatch/pull/3).

scottmarchant:
Hi @akmorrison, any more thoughts or feedback we can help address for this PR?

akmorrison:
@swift-ci please test

akmorrison:
> Hi @akmorrison, any more thoughts or feedback we can help address for this PR?

I like this a lot more! I'm reading through the scheduler API in 304adaff [here](urhttps://github.com/PassiveLogic/swift-corelibs-libdispatch/pull/3/changes#diff-e588aefc74f0d3d1a2d675faf40e34b3cc47753e082e8e459b560d23e4adfab1), I think `_dispatch_wasi_event_loop_set_scheduler` is a much more reasonable API shape for non COCOA_COMPAT/CFRunloop handle shaped targets, and it keeps the common code much cleaner. I've kicked this into CI, this change alone seems fine to me to merge. Thank you for the change!

akmorrison (approved):

scottmarchant:
Thanks @akmorrison! It looks like Linux CI is green here.

Anything else we can help with on this PR?

PR #954, opened 2026-08-25
dispatch: declare Windows static link dependencies

mozharovsky:
Static Windows dispatch objects now declare their own link dependencies through /DEFAULTLIB pragmas in internal.h, so every extracted archive member carries the same closure.

The seven-library set is this repository's CMake link interface (BlocksRuntime, AdvAPI32, ShLwApi, WS2_32, WinMM, synchronization) plus ole32 from the measured static-consumer closure. DLL builds resolved these imports at their own link, static consumers have to supply them at the final link.

Companion to swiftlang/swift#91668 per the discussion there, with swiftlang/swift#91648 as context.

## Validation

End-to-end rides on the next candidate toolchain build. This is the library set that made the swiftlang/swift#91648 reproducer link and run in swiftlang/swift-driver#2174.

### DIFF
--- a/src/block.cpp
+++ b/src/block.cpp
@@ -30,6 +30,11 @@
 
 #include "internal.h"
 
+// WORKAROUND swiftlang/swift#91648
+#if defined(_WIN32) && defined(dispatch_STATIC)
+#pragma comment(lib, "BlocksRuntime.lib")
+#endif
+
 // NOTE: this file must not contain any atomic operations
 
 #if DISPATCH_DEBUG && DISPATCH_BLOCK_PRIVATE_DATA_DEBUG
--- a/src/data.c
+++ b/src/data.c
@@ -20,6 +20,11 @@
 
 #include "internal.h"
 
+// WORKAROUND swiftlang/swift#91648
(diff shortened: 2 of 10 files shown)


src/internal.h
```diff
@@ -36,6 +36,17 @@
...
+// WORKAROUND swiftlang/swift#91648
+#if defined(_WIN32) && defined(dispatch_STATIC)
+#pragma comment(lib, "BlocksRuntime.lib")
+#pragma comment(lib, "AdvAPI32.lib")
+#pragma comment(lib, "ShLwApi.lib")
+#pragma comment(lib, "WS2_32.lib")
+#pragma comment(lib, "WinMM.lib")
+#pragma comment(lib, "synchronization.lib")
+#pragma comment(lib, "ole32.lib")
+#endif
```
compnerd: Please split these up to what pulls in the dependency. It really shouldn't go into `internal.h` but rather the `.cpp` IMO.

mozharovsky: Yeah, that's cleaner, thanks! I split them into the files that actually reference each library.

compnerd (approved):

PR #956, opened 2026-09-02
FreeBSD: typo in workqueue.c

michael-yuji:
Fix a terrible typo I made

### DIFF
--- a/src/event/workqueue.c
+++ b/src/event/workqueue.c
@@ -331,7 +331,7 @@ _dispatch_workq_count_runnable_workers(dispatch_workq_monitor_t mon)
 
 	for (int i = 0; i < mon->num_registered_tids; ++i) {
 		dispatch_tid tid = mon->registered_tids[i];
-		for (int j = 0; i < count; ++i) {
+		for (int j = 0; j < count; ++j) {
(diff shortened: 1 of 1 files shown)


etcwilde (approved):

src/event/workqueue.c
```diff
@@ -331,8 +331,10 @@ _dispatch_workq_count_runnable_workers(dispatch_workq_monitor_t mon)
 
 	for (int i = 0; i < mon->num_registered_tids; ++i) {
 		dispatch_tid tid = mon->registered_tids[i];
-		for (int j = 0; i < count; ++i) {
-			if ((dispatch_tid)kp[j].ki_tid != tid) { continue; }
+		for (int j = 0; j < count; ++j) {
+			if (((dispatch_tid)kp[j].ki_tid << 2) != tid) {
```
akmorrison, line 335: Ah, the `<< 2` is to match the ones in `_dispatch_tid_self()` for the futex bits. The linux implementation uses the top two bits for waiter signaling, we could have done the same here, but this does seem largely equivalent

akmorrison (approved):

PR #957, opened 2026-09-04
FreeBSD: use eventfd abi instead of pipe2

michael-yuji:
Use `eventfd` based CFRunLoop triggering mechanism instead of `pipe2` on FreeBSD.

Companion dispatch PR: https://github.com/swiftlang/swift-corelibs-foundation/pull/5539


akmorrison (approved):

michael-yuji:
@swift-ci please test

